PR #57, opened 2016-02-16
VA-1322 New Status: transcode_starting

kvenn:
#### Ticket
[VA-1322](https://vimean.atlassian.net/browse/VA-1322)

#### Ticket Summary
There is a status that was added to the `clip` object back in November which we weren't accounting for called `transcode_starting`. It would have just been treated as `NONE` which was probably getting treated similarly to AVAILABLE (or whatever the implementation was in our `default` case).
[The original API Git issue](https://github.vimeows.com/Vimeo/vimeo/issues/17447)

#### Implementation Summary
Instead of added a fall through to all of our switch statements (and various `if` statements), we're just adding in a check in the getter method we were already using.
After speaking with @alfiehanssen it seems to make the most sense to just put it right in the getter since `transcode_starting` will always have the same UI implication as `transcoding` (which is mentioned in the JavaDoc so consumers of the library aren't totally in the dark).

#### How to Test
This one is pretty hard to test exactly - Brian Cooper recommended just hardcoding the value to test which isn't totally guaranteed to be accurate. The hardcode did prove to work on all of our conditional statements though.
For QA - best way to test would just be to hammer on the upload feature.

My beautiful hardcode:
screen shot 2016-02-16 at 11 26 58 am

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -116,15 +116,27 @@ public class Video implements Serializable {
     public List<Category> categories;
 
     /**
+     * This will return the value as it's given to us from the API (or {@link Status#NONE if null}). Unlike
+     * {@link Video#getStatus()}, this will return all known statuses for a video (including {@link Status#TRANSCODE_STARTING}.
+     * <p/>
+     * For a more simplified representation of the video status, use {@link Video#getStatus()}.
+     */
+    public Status getRawStatus() {
+        return status == null ? Status.NONE : status;
+    }
+
+    /**
      * This getter is always guaranteed to return a {@link Status}, {@link Status#NONE if null}. If the Status
      * is equal to {@link Status#TRANSCODE_STARTING} then we'll just return the Status {@link Status#TRANSCODING}
      * since they're functionally equivalent from a client-side perspective.
+     * <p/>
+     * For an all-inclusive getter of the video status, use {@link Video#getRawStatus()}.
      */
     public Status getStatus() {
         if (status == Status.TRANSCODE_STARTING) {
             return Status.TRANSCODING;
         }
-        return status == null ? Status.NONE : status;
+        return getRawStatus();
     }
 
     public void setStatus(Status status) {
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -66,8 +66,11 @@ public class Video implements Serializable {
         AVAILABLE("available"),
         @SerializedName("uploading")
         UPLOADING("uploading"),
+        @SerializedName("transcode_starting")
+        TRANSCODE_STARTING("transcode_starting"),
         @SerializedName("transcoding")
         TRANSCODING("transcoding"),
+        // Errors
         @SerializedName("uploading_error")
         UPLOADING_ERROR("uploading_error"),
         @SerializedName("transcoding_error")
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -66,8 +66,11 @@
         AVAILABLE("available"),
         @SerializedName("uploading")
         UPLOADING("uploading"),
+        @SerializedName("transcode_starting")
+        TRANSCODE_STARTING("transcode_starting"),
         @SerializedName("transcoding")
```
kevinzetterstrom: This is a little nitpicky, but since we are in this space, I suggest using constants in the SerializedName and constructor params, and for all Status items.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -102,17 +105,25 @@ public String toString() {
...
 
+    /**
+     * This getter is always guaranteed to return a {@link Status}, {@link Status#NONE if null}. If the Status
+     * is equal to {@link Status#TRANSCODE_STARTING} then we'll just return the Status {@link Status#TRANSCODING}
+     * since they're functionally equivalent from a client-side perspective.
+     */
     public Status getStatus() {
+        if (status == Status.TRANSCODE_STARTING) {
+            return Status.TRANSCODING;
+        }
```
kevinzetterstrom: I know you and @alfiehanssen discussed this, but I am not too sure this is the correct way to go. Perhaps other consumers would like to know when we are in the transcode starting state, if we go with this then they wouldn't be able to account for it. Likewise, we may want to be privy to that status in the future as well. From our UI side it is currently equivalent, but this doesn't reflect a decision from our API team to separate the two.

kvenn: The way I look at it, a vast majority of the use cases we're not going to want to worry about this state that only lasts for a "very temporary" amount of time. And it would only complicate the consumption of our client side API to cause other consumers of the library to have to worry about this additional status when they most likely also won't care.

I don't think it's necessarily dishonest because `TRANSCODE_STARTING` is technically just a phase of `TRANSCODING`.

That being said, it's included in the API request for what I can only image is for technical transparency. So possibly a solution that can cater to both parties would be an additional getter which returns the raw value. And this one can be renamed to `getSimpleStatus()`. Or the raw one can be named `getAdvancedStatus()` or `getRawStatus()` since it caters to the (presumed) minority. We could also use `getRawStatus()` for any logging which will give us that technical transparency.

kevinzetterstrom: I like the addition getter methods, and I think that would solve for this nicely, since status is private.
Side note: we need to cleanup all our fields soon so that they are private with getters. setters shouldn't be needed because these should always be set just from the API.

kvenn: Coolio I'll make the addition. And yeah as we make new classes we can rely on getters. As for the old ones, I'll ticket it and we can make the change all at once since it'll guaranteedly cause some merge conflicts.

kvenn: b08e72a

kevinzetterstrom:
:+1:

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
kvenn: Kind of can go either way since these strings shouldn't be used anywhere else and also don't necessarily need to be tied to one another.
But it looks prettier: b26ec0e

kevinzetterstrom:
merge when ready

PR #59, opened 2016-02-18
Issue template

kevinzetterstrom:
#### Ticket Summary
Adding an issue template and moving the contributing file. @alfiehanssen @kvenn what do y'all think of the issue template?


alfiehanssen:
Love it :tangerine:

kvenn:
Bomb diggity

We should probably hook iOS up with this, or maybe make a ticket for them since they're probably swamped.

kevinzetterstrom:
Glad you guys like it. I will confirm it works as expected after I merge after which I will take care of adding this to all our public mobile repos, iOS included.

Dashron:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

PR #60, opened 2016-02-18
VA-1324 Upload Invalid Parameters

kvenn:
#### Ticket
[VA-1324](https://vimean.atlassian.net/browse/VA-1324)

#### Ticket Summary
Refer to [App PR](https://github.vimeows.com/MobileApps/Vimeo-Android/pull/503)

#### Implementation Summary
Added all the new error codes and reorganized both this file and the originating error code sheet a little bit. Added convenience methods to invalid parameters for access error code.

#### How to Test
Refer to [App PR](https://github.vimeows.com/MobileApps/Vimeo-Android/pull/503)


vimeo-networking/src/main/java/com/vimeo/networking/model/error/ErrorCode.java
```diff
@@ -196,13 +205,99 @@
...
+    @SerializedName("2412")
+    USER_NOT_ALLOWED_TO_SET_CONTACTS_CLIP_PRIVACY,
+    // Upload
     @SerializedName("8002")
     UNABLE_TO_UPLOAD_VIDEO_MISSING_USER_ID_FOR_AUTHENTICATION_TOKEN,
     @SerializedName("3400")
     USER_NOT_ALLOWED_TO_UPLOAD_VIDEO_UNVERIFIED_EMAIL,
-    @SerializedName("3401")
-    USER_NOT_ALLOWED_TO_UPGRADE_TO_1080_VIDEO,
+    @SerializedName("3401") @Deprecated USER_NOT_ALLOWED_TO_UPGRADE_TO_1080_VIDEO,
```
kevinzetterstrom: Since I believe we are the only ones consuming this library at the moment, I am in favor of just deleting this error code, rather than deprecating it. I recently had a similar issue, and opted for deletion.

kevinzetterstrom:
:100: for organizing the error codes! love it.

vimeo-networking/src/main/java/com/vimeo/networking/model/error/ErrorCode.java
kvenn: Yeah I'm down with that, the only reason I kept it in was because the spreadsheet that this file references still has it in there, but it's marked as "deprecated, not used, or removed" so I figured the @deprecated annotation would make it clear that we know about it.

But also that doesn't provide much use so I just removed it entirely.

6f05023

PR #62, opened 2016-02-24
VA-1364 sync fetch method

kevinzetterstrom:
#### Ticket
[VA-1364](https://vimean.atlassian.net/browse/VA-1364)

#### Ticket Summary
Implementing HTC Blinkfeed requires a synchronous call to fetch content

#### Implementation Summary
Added a synchronous version of fetchContent

#### How to Test
Verify existing fetching works as expected, and synchronous fetching also works as expected.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -933,16 +933,7 @@ public class VimeoClient {
         }
         String cacheHeaderValue = createCacheControlString(cacheControl);
 
-        Map<String, String> queryMap = new HashMap<>();
-        if (refinementMap != null && !refinementMap.isEmpty()) {
-            queryMap = refinementMap;
-        }
-        if (query != null && !query.isEmpty()) {
-            queryMap.put(Vimeo.PARAMETER_GET_QUERY, query);
-        }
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -910,6 +911,47 @@ public void fetchCurrentUser(ModelCallback<User> callback) {
...
+            queryMap.put(Vimeo.PARAMETER_GET_FIELD_FILTER, fieldFilter);
+        }
+
+        Call<Object> call = this.vimeoService.GET(getAuthHeader(), uri, queryMap, cacheHeaderValue);
+        try {
+            return call.execute();
+        } catch (IOException e) {
+            return null;
+        }
+    }
```
kvenn: I think we should pull the rest of this method out into a helper (in addition to the cache control jazz) just because these two methods will always be identical except for 1. having a callback and 2. how it gets executed (so the top and the bottom). Maybe with one for each object we're modifying or maybe just one that modifies both objects.

kevinzetterstrom: @kvenn  https://github.com/vimeo/vimeo-networking-java/commit/0f723ca01a3c5d42a556ddafb6f5610bf2dfd106

kvenn:
Looks gooda :wine_glass:

PR #63, opened 2016-02-26
VA-1309 - unlisted privacy

kevinzetterstrom:
#### Ticket
[VA-1309](https://vimean.atlassian.net/browse/VA-1309)

#### Ticket Summary
https://vimeo.com/blog/post/share-unlisted-videos-with-private-links Vimeo recently introduced Unlisted privacy for Pro/Plus users. We need to reflect that in the app and this library. 

#### Implementation Summary
Added the unlisted privacy option. Also added in the password field for password protected videos.

#### How to Test
Set a video to unlisted privacy and verify it we can see that new privacy in the app.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -255,10 +255,12 @@ public class Video implements Serializable {
     }
 
     /**
-     * The password for the video, only sent when the following conditions are true
-     * 1. The privacy is set to password
-     * 2. The user making the request owns the video
-     * 3. The application making the request is granted access to view this field
+     * The password for the video, only sent when the following conditions are true:
+     * <ul>
+     * <ol>The privacy is set to password</ol>
+     * <ol>The user making the request owns the video</ol>
+     * <ol>The application making the request is granted access to view this field</ol>
+     * </ul>
      *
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -252,6 +254,19 @@ public String recommendationsUri() {
         return recommendationsUri;
     }
 
+    /**
+     * The password for the video, only sent when the following conditions are true
+     * 1. The privacy is set to password
+     * 2. The user making the request owns the video
+     * 3. The application making the request is granted access to view this field
```
kvenn: I think this _might_ show up as one line. If you have it open in your IDE, I would mouse over getPassword and just check if the JavaDoc looks good. 

For numbered lists it might be preferable to use <ul> <ol></ol> </ul>. But if this looks good in the JavaDoc, then it's totally fine.

kvenn:
Just made one small comment. Otherwise good to merge whenever :rocket:

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
kevinzetterstrom: To be honest, I think it looked good the first way, but I made this change.

PR #64, opened 2016-03-04
VA-1418 changes for empty post responses

kevinzetterstrom:
#### Ticket
[VA-1418](https://vimean.atlassian.net/browse/VA-1418)

#### Ticket Summary
POST responses are throwing errors

#### Implementation Summary
Retrofit can not handle empty responses without the Void type in the Call. Updated our current empty post method to take parameters to be more widely used. Also beefed up the logging - instead of logging an error when there is no body to parse, just don't even attempt to parse.

#### How to Test
Run the app and check the logs.


alfiehanssen:
:+1:

kvenn:
Looks good!

PR #66, opened 2016-03-15
VA-1304 Access Token Auth

kvenn:
#### Ticket
[VA-1304](https://vimean.atlassian.net/browse/VA-1304)

#### Ticket Summary
We wanted a simple way for consumers of the library to provide an auth token (provided from the developer console) which would remove the need to make client credentials requests.

#### Implementation Summary
* Added an access token to the configuration builder.
* Moved around account setting so the vimeoAccount on the vimeo client has a tighter coupling to the configurationbuilder.accountStore callbacks.
* Added section to the README

#### How to Test
* Test the VimeoApp to ensure nothing broke (log in, log out, change users, get client credentials)
* Use the example app and go through some weird auth flows

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ compile project(':vimeo-networking-java:vimeo-networking')
 ```
 
 ### Initialization
+
+The `VimeoClient` instance is highly customizable through the use of the `Configuration.Builder` class. There are a few different default `Builder` constructors that cover core functionality.
+In the below sections, we cover examples of ways to customize your `VimeoClient` instance. Full implementations are in the example app.
+
+#### Configuration Builder for Apps with Account Management
 On app launch, configure `VimeoClient` with your client key, secret, and scope strings. And once initialization is complete, authenticate if necessary. This authentication can include [client credentials](#client-credentials-grant) or authentication via [code grant](#oauth-authorization-code-grant). 
 
 If you choose to pass in an `AccountStore`, the authenticated account will automatically be persisted. Once authentication is complete, you can access the persisted `VimeoAccount` object through `VimeoClient#getVimeoAccount()`. For a basic implementation of an `AccountStore`, please refer to [this file in the example app](example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java)
@@ -70,6 +75,20 @@ Configuration.Builder configBuilder =
 Setting the cache directory and deserializer are optional but highly recommended.
 * The deserializer allows for deserialization on a background thread. Without it, object deserialization will occur on the main (UI) thread. This can be bad for performance if the API response bodies are large.
 * The cache directory is to allow caching of requests. Without it, no requests will be cached and all other cache settings will be ignored.
+
+#### Configuration Builder for Apps with Only One Developer Account
+You can skip the need for [client credentials](#client-credentials-grant) or [code grant](#oauth-authorization-code-grant) requests if you provide an access token up front. With this access token you'll be able to make any requests that the access token's scope allows.
+
+You can generate an access token in the Authentication tab once you select your app from the [list here](https://developer.vimeo.com/apps).
+
+Once you have the access token, you can easily initialize your `VimeoClient` instance with the below builder.
+ 
+```java
+String accessToken = getString(R.string.access_token);
+return new Configuration.Builder(Vimeo.VIMEO_BASE_URL_STRING, accessToken);
+```
+
+*Note: You will not be able to log out of the account associated with the access token provided to the `Configuration.Builder`
 
 ### Authentication 
 All calls to the Vimeo API must be [authenticated](https://developer.vimeo.com/api/authentication). This means that before making requests to the API you must authenticate and obtain an access token. Two authentication methods are provided: 
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ On app launch, configure `VimeoClient` with your client key, secret, and scope s
 If you choose to pass in an `AccountStore`, the authenticated account will automatically be persisted. Once authentication is complete, you can access the persisted `VimeoAccount` object through `VimeoClient#getVimeoAccount()`. For a basic implementation of an `AccountStore`, please refer to [this file in the example app](example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java)
 ```java
   /**
-    * @param baseURLString The base url pointing to the Vimeo api. Something like: {@link Vimeo#VIMEO_BASE_URL_STRING}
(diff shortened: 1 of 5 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -195,12 +196,19 @@ public VimeoAccount getVimeoAccount() {
 
     public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
         if (vimeoAccount == null) {
-            vimeoAccount = new VimeoAccount();
+            vimeoAccount = new VimeoAccount(this.configuration.accessToken);
+            this.configuration.saveAccount(vimeoAccount, null, null);
```
kevinzetterstrom: Let's check that accessToken is not null - if it is, we should fallback to the previous way the vimeoAccount was created (default constructor)

kvenn: So before it was just an empty constructor that did nothing (except create an object). So instead I modified the constructor so it can take a nullable access token, so that if it's null, it'll be effectively the same as an empty constructor.

We only needed the empty constructor for this case, so I figured I could remove the need for one entirely with this functionality.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -55,6 +55,8 @@
     public String clientSecret;
     public String scope;
 
+    public String accessToken;
```
kevinzetterstrom: I think in order to realize the full pattern of the Builder, we need to make these fields private and have getters for them - this way the only true way to change them is through the Builder. I suggest doing that here for accessToken and creating a ticket for the other fields.

kevinzetterstrom:
:+1:

kvenn:
@alfiehanssen @howardrigberg Since this is a change to our open source lib, it'd be nice to have your feedback as well before pushing a new release

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kvenn: When I was implementing the access token I had the same thought (and moreover, that the builder could probably be slimmed down a bit) so I added a ticket to the tech debt sheet around 1) fixing up the builder and 2) making our vimeoaccount/auth header generation a little more transparent and mutable.

README.md
```diff
@@ -71,6 +76,22 @@ Setting the cache directory and deserializer are optional but highly recommended
...
+Once you have the access token, you can easily initialize your `VimeoClient` instance with the below builder.
+ 
+```java
+String accessToken = getString(R.string.access_token);
+return new Configuration.Builder(Vimeo.VIMEO_BASE_URL_STRING, accessToken);
+```
+
+After providing the access token, if you'd like to have access to the associated `User` object you'll need to make a call to `VimeoClient#fetchCurrentUser`. If you're using an account store, you can update the `VimeoAccount` with the new `User` object.
+
+*Note: You will not be able to log out of the account associated with the access token provided to the `Configuration.Builder`
```
alfiehanssen: I think it's worth describing why we chose do prohibit this

example/src/main/java/com/vimeo/android/networking/example/TestApp.java
```diff
@@ -28,10 +32,32 @@ public void onCreate() {
...
+        // </editor-fold>
+    }
+
+    public Configuration.Builder getAccessTokenBuilder() {
+        // The values file is left out of git, so you'll have to provide your own access token
+        String accessToken = getString(R.string.access_token);
+        return new Configuration.Builder(Vimeo.VIMEO_BASE_URL_STRING, accessToken);
+    }
+
+    public Configuration.Builder getIdAndSecretBuilder() {
```
alfiehanssen: This has probably been in here a while but consider name change to

`getClientIdAndClientSecretBuilder`

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
alfiehanssen: :+1:

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -174,6 +185,21 @@ private boolean isValid() {
...
+         * The most basic builder constructor. If you've only provided an access token, you'll only be able to
+         * make requests for the given access token's account.
+         * <p/>
+         * If you'd like the ability to switch accounts, you'll have to set a client id and client secret.
+         * If you'd like the ability to persist accounts, you'll have to set an account store
+         *
+         * @param baseURLString The base url pointing to the Vimeo api. Something like: {@link Vimeo#VIMEO_BASE_URL_STRING}
+         * @param accessToken   Token provided by the Vimeo developer console
+         */
+        public Builder(String baseURLString, String accessToken) {
```
alfiehanssen: What's the use case for passing in the `baseURLString`? For CI builds?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -491,7 +497,16 @@ public VimeoAccount logIn(String email, String password) {
...
      */
     public Call<Object> logOut(@Nullable final VimeoCallback<Object> callback) {
-
+        // If you've provided an access token to the configuration builder, we're assuming that you wouldn't
+        // want to be able to log out of it, because this would invalidate the constant you've provided us.
+        if (configuration.accessToken != null &&
+            configuration.accessToken.equals(vimeoAccount.getAccessToken())) {
+            if (callback != null) {
+                callback.failure(new VimeoError(
+                        "You can't log out of the account provided through the configuration builder"));
```
alfiehanssen: Augment to indicate why

README.md
kvenn: "This is because we wouldn't want anyone to accidentally invalidate/delete the token which is being used to authenticate users in a production application. You will still be able to delete the token via the web [developer console](https://developer.vimeo.com/apps/).
If this seems like restricting functionality, please log an issue to the [issue tracker](https://github.com/vimeo/vimeo-networking-java/issues)."

example/src/main/java/com/vimeo/android/networking/example/TestApp.java
kvenn: It's just for the example, but I think definitely makes it more clear - just made the change :+1:

alfiehanssen:
boo yah :+1:

PR #67, opened 2016-03-18
VA-1413 video resource key

kevinzetterstrom:
#### Ticket
[VA-1413](https://vimean.atlassian.net/browse/VA-1413)

#### Ticket Summary
Adding resource_key field for clip objects

#### Implementation Summary
Added the resourceKey field, and use it for comparison methods

#### How to Test
Will need VA-1413 branch of the app for testing.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -125,6 +125,10 @@ public class Video implements Serializable {
     public List<Category> categories;
     @Nullable
     private String password;
+
+    /**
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -125,6 +125,7 @@ public String toString() {
     public List<Category> categories;
     @Nullable
     private String password;
+    private String resourceKey;
```
kvenn: It might be nice to add a JavaDoc to this field just so consumers of the library will get why we have it (since it might be useful to them).

kvenn:
Just the one comment - merge when ready

PR #69, opened 2016-03-31
VA-1432 Cleanup Part 1

kvenn:
#### Ticket
[VA-1432](https://vimean.atlassian.net/browse/VA-1432)

#### Ticket Summary
Part of download cleanup

#### Implementation Summary
* Adding in md5 field to video file for comparison
* Added convenience method in video for access a file by md5
* Added in synchronous getter for video object
* isExpired convenience method on VideoFile (originally in a util)

#### How to Test
* Fetch videos sychronously

@kevinzetterstrom - it'd be nice to have you peep this as well

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -882,9 +882,13 @@ public class VimeoClient {
     }
 
     @Nullable
-    public retrofit2.Response<Video> fetchVideoSync(String uri) {
+    public retrofit2.Response<Video> fetchVideoSync(String uri, @Nullable String fieldFilter) {
         try {
-            return vimeoService.getVideo(getAuthHeader(), uri).execute();
+            HashMap<String, String> paramMap = new HashMap<>();
+            if (fieldFilter != null) {
+                paramMap.put(Vimeo.PARAMETER_GET_FIELD_FILTER, fieldFilter);
+            }
+            return vimeoService.getVideo(getAuthHeader(), uri, paramMap).execute();
         } catch (IOException e) {
             return null;
         }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -31,6 +31,7 @@ import com.vimeo.networking.model.Comment;
 import com.vimeo.networking.model.PictureResource;
 import com.vimeo.networking.model.Privacy;
 import com.vimeo.networking.model.User;
+import com.vimeo.networking.model.Video;
 import com.vimeo.networking.model.VimeoAccount;
 import com.vimeo.networking.model.error.ErrorCode;
 import com.vimeo.networking.model.error.VimeoError;
@@ -880,6 +881,14 @@ public class VimeoClient {
         return call;
     }
 
+    @Nullable
+    public retrofit2.Response<Video> fetchVideoSync(String uri) {
+        try {
+            return vimeoService.getVideo(getAuthHeader(), uri).execute();
+        } catch (IOException e) {
+            return null;
+        }
+    }
     // </editor-fold>
 
     /**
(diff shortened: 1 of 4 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -880,6 +881,14 @@ public void deletePictureResource(String uri, ModelCallback callback) {
         return call;
     }
 
+    @Nullable
+    public retrofit2.Response<Video> fetchVideoSync(String uri) {
```
kevinzetterstrom: How about just returning a Video rather then the entire Response object?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -129,10 +130,18 @@
                                  @Body HashMap<String, String> parameters);
     // </editor-fold>
 
+    // -----------------------------------------------------------------------------------------------------
+    // Concrete Region
+    // -----------------------------------------------------------------------------------------------------
+    // <editor-fold desc="Concrete Region">
+    @GET
+    Call<Video> getVideo(@Header("Authorization") String authHeader, @Url String uri);
```
kevinzetterstrom: Field filters?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kvenn: I thought about it, but then returning "null" in the case of a failure isn't really descriptive enough - the caller may want to know if it was for network or what the the status code was.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -244,9 +266,31 @@ public Integer playCount() {
...
+    // Files
+    // -----------------------------------------------------------------------------------------------------
+    // <editor-fold desc="Files">
+    @Nullable
+    public VideoFile getVideoFileForMd5(String md5) {
+        if (files == null) {
+            return null;
+        }
+        for (VideoFile file : files) {
+            if (file != null && file.md5.equals(md5)) {
```
kevinzetterstrom: Will this throw a NPE is ```file.md5``` is null?

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
```diff
@@ -88,6 +88,9 @@ public String toString() {
     private VideoQuality quality;
     private MimeType type;
     public VideoLog log;
+    /** The md5 provides us with a way to uniquely identify video files at {@link #link} */
+    @SerializedName("md5")
+    public String md5;
```
kevinzetterstrom: Why do you need a ```SerializedNamed``` annotation if it's the same as the variable name? Doesn't hurt to have it, but we don't have it elsewhere.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: Fair enough. Also stays in line with our other sync methods.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
kvenn: Good call, added in nullable param

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
kvenn: Yeah definitely couldn't hurt!

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
kvenn: We should probably be using `@SerializedName` on every field going forward because of how GSON's versioning works. If we persist any object without it, and we rename the field then it will be broken forever. But as long as the annotation never changes, we can rename until the cows come home without any worry a versioning incompatibility.
So it's just a good way to future proof ourselves and we can only really benefit from it.
It's unlikely that our api will change the field names - but it's also totally possible.
Kind of as you said, it couldn't hurt.

kvenn:
@kevinzetterstrom peeeeep 4db7c16

kvenn:
@howardrigberg @kevinzetterstrom pr changes: 542bfa0

PR #70, opened 2016-04-04
VA-1059 deprecate password saving

kevinzetterstrom:
#### Ticket
[VA-1059](https://vimean.atlassian.net/browse/VA-1059)

#### Ticket Summary
Deprecate saving of the password into the account store

#### Implementation Summary
Instead of flat out removal, I created secondary methods without password parameters, and deprecated all methods using the password parameter. 

#### How to Test
Test in the app PR

### DIFF
--- a/example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
+++ b/example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
@@ -38,8 +38,14 @@ public class TestAccountStore implements AccountStore {
         return AccountPreferenceManager.getClientAccount();
     }
 
+    @Deprecated
     @Override
     public void saveAccount(VimeoAccount vimeoAccount, String email, String password) {
+        saveAccount(vimeoAccount, email);
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


kvenn:
Nice! Cleans up the signatures a bit too. We can probably remove the deprecated methods once people catch on to the fact that they're not doing anything.
We can still mark our next release as a minor release since it has backwards compatibility. Then for our next major release we can remove all of our deprecated methods.

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
```diff
@@ -38,8 +38,14 @@ public VimeoAccount loadAccount() {
...
 
+    @Deprecated
     @Override
     public void saveAccount(VimeoAccount vimeoAccount, String email, String password) {
+        saveAccount(vimeoAccount, email);
+    }
+
+    @Override
+    public void saveAccount(VimeoAccount vimeoAccount, String email) {
         AccountPreferenceManager.setClientAccount(vimeoAccount);
```
kvenn: Awesome - keeping the sample up to date

PR #73, opened 2016-04-21
VA-1372: Adding global password to VideosPreference object

howardrigberg:
#### Ticket
**Required for Vimeans only**
[VA-1372](https://vimean.atlassian.net/browse/VA-1372)

#### Ticket Summary
Default password not displaying during upload.

#### Implementation Summary
Adding global password to VideosPreference object.

#### How to Test
Nothing specific to test with this networking ticket alone. Just make sure that video password privacy still works as it used to.


kvenn:
Merge when ready

PR #75, opened 2016-04-26
VA-1588 Play object

kevinzetterstrom:
#### Ticket
[VA-1588](https://vimean.atlassian.net/browse/VA-1588)

#### Ticket Summary
Add the new 'play' object as part of a clip

#### Implementation Summary
Added new objects, modified ```VideoLog```, and updated ```Video``` to make use of the ```Play``` object. Also made a decent amount of changes to ```VideoFile``` for deprecating unused members and force getters.

#### How to Test
Should be examined in Android Studio and in conjunction with the updates to the Android code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ vimeo-networking is a Java networking library used for interacting with the Vime
      * [Callbacks](#callbacks)
      * [Cache Strategy](#cache-strategy)
      * [Sample Requests](#sample-requests)
+* [FAQs](#faqs)
+     * [How do I play a video?](#how-do-i-play-a-video)
 * [Found an Issue?](#found-an-issue)
 * [Questions](#questions)
 * [License](#license)
@@ -239,6 +241,36 @@ VimeoClient.getInstance().fetchCurrentUser(new ModelCallback<User>(User.class) {
           // Log the error and update the UI to tell the user there was an error
      }
 });
+```
+
+## FAQs
+### How do I play a video?
+Once you have a ```Video``` object you have two choices - embed or play natively. All consumers of this library will be able to access the embed data, however, in order to play videos natively (for example, in a VideoView or using ExoPlayer on Android) you must have access to the ```VideoFile``` links - this is restricted to users with Vimeo PRO accounts accessing their own videos.
+#### Embed
+All consumers of this library will have access to a video's embed html. Once you request a video you can get access to this as such:
+```java
+WebView webview = ...; // obtain a handle to your webview
+Video video = ...; // obtain a video in a manner described in the Requests section
+String html = video.embed != null ? video.embed.html : null;
+if(html != null) {
+     // html is in the form "<iframe .... ></iframe>"
+     webview.loadData(html, "text/html", "utf-8");
+}
+```
+#### Native playback
+If you are a Vimeo PRO member, you will get access to your own videos' links; during [initialization](#initialization) of this library, you must provide the ```video_files``` scope. With these, you can choose to stream the videos in any manner you wish. You can get access to HLS, Dash, or progressive streaming video files through a video's ```Play``` object. 
+```java
+VideoFile hlsFile = video.getPlay() != null ? video.getPlay().getHlsVideoFile() : null;
+if(hlsFile != null) {
+     String hlsLink = hlsFile.getLink();
+     // load HLS link
+}
+ArrayList<VideoFile> progressiveFiles = video.getPlay() != null ? video.getPlay().getProgressiveVideoFiles() : null;
+if(progressiveFiles != null && !progressiveFiles.isEmpty()) {
+     VideoFile progressiveStream = progressiveFiles.get(0); // you could sort these files by size, fps, width/height
+     String link = progressiveStream.getLink();
+     // load link
+}
 ```
 
 ## Found an Issue?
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -241,6 +243,31 @@ VimeoClient.getInstance().fetchCurrentUser(new ModelCallback<User>(User.class) {
 });
 ```
 
+## FAQs
+### How do I play a video?
+Once you have a ```Video``` object you have two choices - embed or play natively. All consumers of this library will be able to access the embed data, however, in order to play videos natively (for example, in a VideoView or using ExoPlayer on Android) you must have access to the ```VideoFile``` links - this is restricted to users with Vimeo PRO accounts accessing their own videos.
```
kvenn: I would include a bit here specifying that a pro user can get a list of their videos through `fetchContent` to `User#getVideosConnection().uri` (on the result of fetchCurrentUser) or through hitting the `/me/videos` endpoint when the current PRO user is authenticated.
It might just clear up some ambiguity since that part's pretty crucial for pro users getting their vids

README.md
```diff
@@ -241,6 +243,31 @@ VimeoClient.getInstance().fetchCurrentUser(new ModelCallback<User>(User.class) {
...
+#### Embed
+All consumers of this library will have access to a video's embed html. Once you request a video you can get access to this as such:
+```java
+WebView webview = ...; // obtain a handle to your webview
+Video video = ...; // obtain a video in a manner described in the Requests section
+String html = video.embed != null ? video.embed.html : null;
+if(html != null) {
+     // html is in the form "<iframe .... ></iframe>"
+     webview.loadData(html, "text/html", "utf-8");
```
howardrigberg: It might be more direct to not make assumptions about the use of a WebView.

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
```diff
@@ -104,9 +173,54 @@ public boolean isVP6() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || !(o instanceof VideoFile)) {
+            return false;
+        }
+
+        VideoFile that = (VideoFile) o;
+
+        return (this.getLink() != null && that.getLink() != null) && this.getLink().equals(that.getLink());
```
howardrigberg: I think that the equals/hashCode implementation here and elsewhere in our models are technically incorrect. We should probably button all of them down since this is a public library. The issues discussed below are not just in this class, but all model classes in networking that I have seen.

1) equals id should never be defined based on a mutable field.
2) we might want to add a canEquals method to handle proper inheritance behavior wrt equality

Also, if both objects have a null link, they are not equal? Just want to double check that this is the behavior that we want.

This article goes over some of the fine points:
http://www.artima.com/lejava/articles/equality.html

vimeo-networking/src/main/java/com/vimeo/networking/model/playback/Play.java
```diff
@@ -0,0 +1,206 @@
...
+ * <p/>
+ * Created by zetterstromk on 4/25/16.
+ */
+public class Play implements Serializable {
+
+    private static final long serialVersionUID = -7429617944240759711L;
+
+    public enum Status {
+        // if not done transcoding
+        @SerializedName("unavailable")
```
howardrigberg: Perhaps make all of these strings private static final strings above.

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
```diff
@@ -104,9 +173,54 @@ public boolean isVP6() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        return this.getLink() != null ? this.getLink().hashCode() : 0;
```
howardrigberg: "this.getLink()" accessor is unnecessary. could just use "link" to keep things terse

vimeo-networking/src/main/java/com/vimeo/networking/model/playback/Play.java
```diff
@@ -0,0 +1,206 @@
...
+        if (getProgressiveVideoFiles() != null) {
+            count += getProgressiveVideoFiles().size();
+        }
+        return count;
+    }
+
+
+    @Nullable
+    public String getLoadLoggingUrl() {
+        if (getHlsVideoFile() != null && getHlsVideoFile().getLog() != null &&
```
howardrigberg: This section of code could be terser if you just accessed your members directly. (e.g. mHls instead of getHlsVideoFile)

vimeo-networking/src/main/java/com/vimeo/networking/model/playback/PlayProgress.java
```diff
@@ -0,0 +1,49 @@
...
+public class PlayProgress implements Serializable {
+
+    private static final long serialVersionUID = -3745271302058282379L;
+
+    @SerializedName("seconds")
+    private int mSeconds;
+
+    /**
+     * @return where the user has progressed to in the video, in seconds
+     */
```
howardrigberg: Is this value ever updated in the app as a user watches something?

vimeo-networking/src/main/java/com/vimeo/networking/model/playback/Play.java
kevinzetterstrom: We have done it that way and this way if you look throughout the model package. This is probably easier to read, while the other way (suggested) tends to fit better if we need the string outside of this usage.

PR #76, opened 2016-04-27
VA-1618 Added VOD container model

howardrigberg:
#### Ticket
**Required for Vimeans only**
[VA-1618](https://vimean.atlassian.net/browse/VA-1618)

#### Ticket Summary
Create a mode to support a VOD container object.

#### Implementation Summary
Added a VodItem object and a VodList object. Only supporting the fields that we need at the moment to support VOD.

#### How to Test
Ensure that the project builds as expected and nothing is broken related to existing networking (regression tests pass).

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/VodItem.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/VodItem.java
@@ -1,0 +1,37 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016 Vimeo
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/VodItem.java
```diff
@@ -0,0 +1,220 @@
...
+
+    @SuppressWarnings("unused")
+    @Nullable
+    @SerializedName("description")
+    private String mDescription;
+
+    @SuppressWarnings("unused")
+    @Nullable
+    @SerializedName("type")
+    private String mType;
```
kevinzetterstrom: Instead of String here, use ```VodType```, then have constructor params for the VodType enum. Then "type" will automatically be deserialized into a VodType.

howardrigberg: Did not know that. Very cool.

howardrigberg:
Responded to comments:
https://github.com/vimeo/vimeo-networking-java/commit/9732343edd6eda7613b2e20028926eb9189c35c8

PR #79, opened 2016-05-04
VA-1618: Added model support for final (hopefully) VOD requirements

howardrigberg:
#### Ticket
[VA-1618](https://vimean.atlassian.net/browse/VA-1618)

#### Ticket Summary
Adding model support for networking requirements of VOD.

#### Implementation Summary
Added totals to the Connection object for tracking viewable series counts.
Added purchase_time sort for sorting VODs by date purchased.
Added ondemand and trailer connections
Added stream and expiration to Interaction
Added buy, rent and subscribe to InteractionsCollection
Modified Video object to perform the actual calculations and method implemenations for VOD based on the supporting models
Added some annotations to existing objects
Added some convenience methods to VodItem

#### How to Test
Make sure that no existing VOD and networking features are broken.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
@@ -45,9 +45,9 @@ public class Connection implements Serializable {
     @SerializedName("total")
     public int total;
     @SerializedName("main_total")
-    public int main_total;
+    public int mainTotal;
     @SerializedName("extra_total")
-    public int extra_total;
+    public int extraTotal;
     @SerializedName("viewable_total")
-    public int viewable_total;
+    public int viewableTotal;
 }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
@@ -22,16 +22,32 @@
 
 package com.vimeo.networking.model;
 
+import com.google.gson.annotations.SerializedName;
+
+import org.jetbrains.annotations.Nullable;
+
 import java.io.Serializable;
 import java.util.ArrayList;
 
 /**
+ * This model object represents a Connection.
  * Created by hanssena on 4/23/15.
  */
 public class Connection implements Serializable {
 
     private static final long serialVersionUID = -840088720891343176L;
+    @Nullable
+    @SerializedName("uri")
     public String uri;
+    @Nullable
+    @SerializedName("options")
     public ArrayList<String> options;
+    @SerializedName("total")
     public int total;
+    @SerializedName("main_total")
+    public int main_total;
+    @SerializedName("extra_total")
+    public int extra_total;
+    @SerializedName("viewable_total")
+    public int viewable_total;
 }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
@@ -43,13 +43,13 @@ public class Interaction implements Serializable {
 
     public enum Stream {
         @SerializedName(STREAM_PURCHASED)
-        PURCHASED(STREAM_PURCHASED),
+        PURCHASED(STREAM_PURCHASED), // you have purchased it (without using a promo code)
         @SerializedName(STREAM_RESTRICTED)
-        RESTRICTED(STREAM_RESTRICTED),
+        RESTRICTED(STREAM_RESTRICTED), // you don't have it, can't purchase it in this country
         @SerializedName(STREAM_AVAILABLE)
-        AVAILABLE(STREAM_AVAILABLE),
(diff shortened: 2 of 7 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
```diff
@@ -22,16 +22,32 @@
...
+    @SerializedName("options")
     public ArrayList<String> options;
+    @SerializedName("total")
     public int total;
+    @SerializedName("main_total")
+    public int main_total;
+    @SerializedName("extra_total")
+    public int extra_total;
+    @SerializedName("viewable_total")
+    public int viewable_total;
```
kevinzetterstrom: Nice addition with the serialized name. I would stay consistent with member naming though with other classes, and use camelCase here, such as ```viewableTotal```. Plus, our GsonBuilder is set up so that if the SerializedName annotation was left off, it would correctly parse ```viewableTotal```.

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
```diff
@@ -22,16 +22,59 @@
...
 
+import com.google.gson.annotations.SerializedName;
+
+import org.jetbrains.annotations.Nullable;
+
 import java.io.Serializable;
 import java.util.Date;
 
 /**
+ * This model object represents an Interaction.
```
kevinzetterstrom: 🔪  take that lint!

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
```diff
@@ -22,16 +22,59 @@
...
+        @Override
+        public String toString() {
+            return mName;
+        }
+    }
+
+    @SerializedName("added")
     public boolean added;
+    @Nullable
+    @SerializedName("addedDate")
```
kevinzetterstrom: this annotation should be added_date no?

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
```diff
@@ -22,16 +22,59 @@
...
+
+    public enum Stream {
+        @SerializedName(STREAM_PURCHASED)
+        PURCHASED(STREAM_PURCHASED),
+        @SerializedName(STREAM_RESTRICTED)
+        RESTRICTED(STREAM_RESTRICTED),
+        @SerializedName(STREAM_AVAILABLE)
+        AVAILABLE(STREAM_AVAILABLE),
+        @SerializedName(STREAM_UNAVAILABLE)
+        UNAVAILABLE(STREAM_UNAVAILABLE);
```
kevinzetterstrom: It may be nice to add an inline comment for each type to let consumers known what each is representative of.

howardrigberg: Yeah, in general it would be great to do that, but the networking library is pretty absent of any kind of Java docs. Maybe we should take care of that in a general sweep when we update all of the classes with private getters and setters? What do you think?

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
kevinzetterstrom: Also, when I just did a video request, I saw a field ```added_time``` but not ```added_date```. Perhaps the existing field was named incorrectly...

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
kevinzetterstrom: I am fine if we leave it, but it would be nice to have. Not necessarily a javadoc, but an inline comment maybe.

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
howardrigberg: Great catch! That was due to the fact that I was assuming a 1-to-1 relationship with the names. I didn't know about the GsonBuilder setup.

vimeo-networking/src/main/java/com/vimeo/networking/model/VodItem.java
```diff
@@ -143,6 +143,22 @@ public Metadata getMetadata() {
...
+        }
+        return null;
+    }
+
+    @Nullable
+    public ConnectionCollection getConnections() {
+        if (mMetadata != null) {
+            return mMetadata.connections;
+        }
+        return null;
```
kevinzetterstrom: This method could be collapsed into one line, same as the method about it: ```return mMetadata != null ? mMetadata.connections : null;```

kevinzetterstrom:
Done with review. Looks great, just a few small comments.

howardrigberg:
Here's the update: 
https://github.com/vimeo/vimeo-networking-java/commit/5405c0e399b3254f3f8c19d65a582e3eabbaf601

kevinzetterstrom:
Looks good. Merge when ready!

PR #86, opened 2016-05-13
VA-1654 interceptor support

anthonycr:
#### Ticket
[VA-1654](https://vimean.atlassian.net/browse/VA-1654)

#### Ticket Summary
We want to be able to inspect our network requests with Stetho. This means we have to allow our `OkHttpClient` to have a special `StethoInterceptor` to it. Since we construct the client in the networking library, we need to add a hook to set a custom interceptor.

#### Implementation Summary
- Added a hook to add an Interceptor that is used across all `OkHttpClient` instances built by our builder.
- Upgraded gradle dist version, added some flags to speed up builds

#### How to Test
- add any interceptors you want to the `Configuration` object when initializing `VimeoClient`.
- example:

```
configBuilder.addNetworkInterceptor(new MyInterceptor());
VimeoClient.initialize(configBuilder.build());
```

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -30,8 +30,11 @@ import com.vimeo.networking.model.VimeoAccount;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
 
 import okhttp3.Cache;
+import okhttp3.Interceptor;
 
 /**
  * The configuration object for making API calls with Retrofit.
@@ -60,6 +63,11 @@ public class Configuration {
     @Nullable
     private AccountStore accountStore;
     public GsonDeserializer deserializer;
+
+    @Nullable
+    public List<Interceptor> networkInterceptors;
+    @Nullable
+    public List<Interceptor> interceptors;
 
     public String apiVersionString;
     public String codeGrantRedirectURI;
@@ -156,6 +164,8 @@ public class Configuration {
         this.certPinningEnabled = builder.certPinningEnabled;
         this.debugLogger = builder.debugLogger;
         this.logLevel = builder.logLevel;
+        this.networkInterceptors = builder.networkInterceptors;
+        this.interceptors = builder.interceptors;
     }
 
     private boolean isValid() {
@@ -188,6 +198,9 @@ public class Configuration {
         private int cacheMaxAge = DEFAULT_CACHE_MAX_AGE;
         private String userAgentString = DEFAULT_USER_AGENT;
         public int timeout = DEFAULT_TIMEOUT;
+
+        private List<Interceptor> networkInterceptors;
+        private List<Interceptor> interceptors;
 
         public String codeGrantRedirectUri = "vimeo" + clientID + "://auth";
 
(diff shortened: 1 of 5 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/RetrofitClientBuilder.java
```diff
@@ -66,6 +70,19 @@
...
+     * This will be used as a network interceptor to
+     * intercept all network traffic.
+     *
+     * @param debugInterceptor the interceptor to use as a
+     *                         debug interceptor. Can be null.
+     */
+    public static void setDebugInterceptor(@Nullable Interceptor debugInterceptor) {
+        sDebugInterceptor = debugInterceptor;
+    }
+
```
kevinzetterstrom: I think there is merit in having consumers of this library be able to add more than one other network interceptor (plus non-network interceptors). What I think is the ```Configuration``` object passed in to VimeoClient ```initialize``` can know about these, and when using the RetrofitClientBuilder, it can make use of the existing add*Interceptors.

anthonycr:
Added more generic support for Interceptors, both network and regular https://github.com/vimeo/vimeo-networking-java/pull/86/commits/6d43a3e8ecfddb7ac2c903995598c78f25b3fd63

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -326,6 +339,36 @@ public Builder setLogLevel(LogLevel level) {
             return this;
         }
 
+        public Builder addNetworkInterceptor(Interceptor interceptor) {
+            if (interceptor != null) {
+                this.networkInterceptors = new ArrayList<>();
+                this.networkInterceptors.add(interceptor);
```
kevinzetterstrom: I would say this behavior needs to be modified. As a user, I may want to call ```addNetworkInterceptor``` then ```addNetworkInterceptor``` again. With this logic, the first interceptor I add would be gone.

anthonycr: whoooooooops

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -326,6 +339,36 @@ public Builder setLogLevel(LogLevel level) {
...
+            }
+            return this;
+        }
+
+        public Builder addNetworkInterceptors(List<Interceptor> interceptors) {
+            if (interceptors != null) {
+                this.networkInterceptors = new ArrayList<>(interceptors);
```
kevinzetterstrom: Same with this. This may be good behavior is the method was named "setNetworkInterceptor", but as it stands, I would think the logic is that networkingInterceptors is already created, and this does a ```this.networkInterceptors.addAll(interceptors);```

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -159,6 +159,13 @@ public Response intercept(Chain chain) throws IOException {
                     }
                 });
 
+        if (this.configuration.networkInterceptors != null) {
+            retrofitClientBuilder.addNetworkInterceptors(this.configuration.networkInterceptors);
+        }
+        if (this.configuration.interceptors != null) {
+            retrofitClientBuilder.addInterceptors(this.configuration.interceptors);
+        }
+
```
kevinzetterstrom: 👍

anthonycr:
addressed comments here https://github.com/vimeo/vimeo-networking-java/pull/86/commits/8d3c26bf6054224e4e34766c4ded711317f35273

kevinzetterstrom:
Looks great, merge when ready

PR #90, opened 2016-05-26
Adding more robust logic around purchase expiration

howardrigberg:
#### Ticket
**Required for Vimeans only**
[VA-1622](https://vimean.atlassian.net/browse/VA-1622)

#### Ticket Summary
It is possible for VOD objects to have multiple purchase states set simultaneously. We need to account for this and arbitrate properly. A VOD video's purchase type is chosen based on the following priority:
     1) trailer
     2) purchased
     3) both rental and subscription? choose the later expiration, expirations equal? choose subscription
     4) subscription
     5) rental
The ticket itself doesn't address this, but it is a bucket that I am attaching remaining VOD work to. 

#### Implementation Summary
Modified the video.java model object's convenience methods to arbitrate the purchase type and date according to more complex rules. Also, removed some lint errors.

#### How to Test
Nothing to test independently here. The changes will have to be tested in conjunction with changes to the app that will be layered in later.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -38,6 +38,7 @@ import java.util.List;
  * A model class representing a Video.
  * Created by alfredhanssen on 4/12/15.
  */
+@SuppressWarnings("unused")
 public class Video implements Serializable {
 
     private static final long serialVersionUID = -1282907783845240057L;
@@ -99,16 +100,16 @@ public class Video implements Serializable {
         @SerializedName(STATUS_QUOTA_EXCEEDED)
         QUOTA_EXCEEDED(STATUS_QUOTA_EXCEEDED);
 
-        private String string;
+        private final String mString;
 
         Status(String string) {
-            this.string = string;
+            mString = string;
         }
 
         @Override
         // Overridden for analytics.
         public String toString() {
-            return this.string;
+            return mString;
         }
     }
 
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -425,6 +458,26 @@ private boolean isPossiblePurchase() {
...
+            return metadata.interactions.rent.expiration;
+        }
+        return null;
+    }
+
+    @Nullable
+    public Date getSubscriptionExpiration() {
+        if (isSubscription()) {
+            // isSubscription will validate and prevent possible npes
+            assert metadata.interactions.subscribe != null;
```
kevinzetterstrom: 👍

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -99,16 +100,16 @@
         @SerializedName(STATUS_QUOTA_EXCEEDED)
         QUOTA_EXCEEDED(STATUS_QUOTA_EXCEEDED);
 
-        private String string;
+        private final String mString;
```
anthonycr: Minor: this variable could be a little more descriptive - like "statusTitle" or something

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -381,19 +382,40 @@ public Play getPlay() {
...
             }
-            if (isRental()) {
-                return VodVideoType.RENTAL;
+            if (isPurchase()) {
+                return VodVideoType.PURCHASE;
+            }
+            // rentals or subscriptions that have been purchased will always have an expiration date
+            Date rentalExpires = getRentalExpiration();
+            Date subscriptionExpires = getSubscriptionExpiration();
+            if (rentalExpires != null && subscriptionExpires != null) {
```
anthonycr: Could there be a case where only one of these dates is null? or would that indicate a problem with the API?

anthonycr:
Looks good, just a couple minor comments. If you don't think they are worth looking into, feel free to merge 👍

PR #94, opened 2016-06-16
VA-1367 Gson Type Adapters for concrete requests 

kevinzetterstrom:
#### Ticket
[VA-1367](https://vimean.atlassian.net/browse/VA-1367)

#### Ticket Summary
Adding in Gson TypeAdapters for concrete requests (PictureResource, Comment, VimeoAccount, and Video) and all objects those depend on.... which is most objects.

#### Implementation Summary
I added type adapters, and used the TypeAdapterFactory to provide TypeAdapters with other TypeAdapters to handle nested objects.
I introduced a simple logging mechanism that will use the VimeoClient logger and fallback to system.out if VimeoClient hasn't been initialized. Since operating in the context of models, this is a possibility we need to account for.

#### How to Test
Run through the app (see VimeoAndroid PR) and verify that functionality still works. You can put breakpoints in the TypeAdapter reads (and writes) to see that they are getting used. To test easily, you can add some hacky code to use gson fromJson and toJson methods. 

@howardrigberg @anthonycr Please feel free to review as well.


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -210,6 +211,10 @@ public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
         this.vimeoAccount = vimeoAccount;
     }
 
+    public DebugLoggerInterface getDebugLogger() {
+        return this.configuration.debugLogger;
+    }
```
kvenn: noice!

vimeo-networking/src/main/java/com/vimeo/networking/logging/ClientLogger.java
```diff
@@ -0,0 +1,76 @@
...
+package com.vimeo.networking.logging;
+
+import com.vimeo.networking.VimeoClient;
+
+/**
+ * Simple utility class to delegate logging when it is not known whether VimeoClient has been initialized yet
+ * <p/>
+ * Created by zetterstromk on 6/15/16.
+ */
+public final class ClientLogger {
```
kvenn: When wouldn't VimeoClient be initialized yet? It now seems like we have two different routes we could go through to log something in VimeoClient which feels like it might be too much (like too many classes doing the same thing). Because now we have ClientLogger and DebugLogger which both handle logs - and then this class and VimeoClient which both hold and instance of a logger. Additionally, since this is a wrapper anytime we add something to the DebugLoggerInterface, we'll also have to add it to our DebugLogger and then we'll also have to add it to this class. What if we had a static method on VimeoClient that returned a DebugLoggerInterface - that removes convenience but minimizes the needs to make modifications. Or maybe making DebugLogger.java static kind of combining this class and DebugLogger.  Or maybe ensuring that VimeoClient is actually initialized before hand.
I'm just spitballing, it just seems like we have a bunch of classes and instances around logging so if we can minimize the logging footprint, it would probably be beneficial.

vimeo-networking/src/main/java/com/vimeo/networking/model/BaseResponseList.java
```diff
@@ -22,10 +22,22 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
 
 /**
+ * Response for list requests
+ * <p/>
```
kvenn: 🔥

vimeo-networking/src/main/java/com/vimeo/networking/logging/ClientLogger.java
kvenn: Two more suggestions from our conversation:
1) VimeoClient's access to the debug logger on the configuration gets more closed off and everything routes through client logger (which also implements the interface to give it a tighter contract). This way this class is the defacto source for logging in vimeo client.

2) DebugLoggingInterface becomes an abstract class with method bodies that just log to system.out (giving it default functionality). And then another class could extend it - this class could have static accessors and maybe be initialized with VimeoClient's debug logging functionality and if that doesnt' exist, the class will fallback to it's superclasses default functionality. Pros - it removes the need for a class. Cons - the contract is a little looser.

vimeo-networking/src/main/java/com/vimeo/networking/model/BaseResponseList.java
```diff
@@ -37,6 +49,92 @@
...
+            out.beginObject();
+            if (value == null) {
+                ClientLogger.d("BaseResponseList null in write()");
+                out.endObject();
+                return;
+            }
+
+            out.name("total").value(value.total);
+            out.name("page").value(value.page);
+            out.name("per_page").value(value.perPage);
```
kvenn: These stings should probably be pulled out into constants since they're referenced from multiple locations (with the same value). But also it seems like this has to be something we could find an annotation for. Now if we want to add a field, instead of just adding it with one line, it takes knowing to add it to the read and write methods and also pairing it with what type it is. If it's an array it looks like it's even more work.

Additionally, if we're going this route then we'd have to do this for all of models every time we create a new object which isn't very frequent, but it's also just added time.

vimeo-networking/src/main/java/com/vimeo/networking/model/Comment.java
```diff
@@ -75,12 +90,126 @@ public boolean equals(Object o) {
...
+                if (userTypeAdapter instanceof UserTypeAdapter &&
+                    metadataTypeAdapter instanceof MetadataTypeAdapter) {
+                    return (TypeAdapter<T>) new CommentTypeAdapter((UserTypeAdapter) userTypeAdapter,
+                                                                   (MetadataTypeAdapter) metadataTypeAdapter);
+                }
+            }
+
+            // by returning null, Gson will never check this factory if it can handle this TypeToken again [KZ] 6/15/16
+            return null;
+        }
```
kvenn: Is there anyway to abstract any of this out? There seems to be a ton of boilerplate that comes with these methods

vimeo-networking/src/main/java/com/vimeo/networking/model/Comment.java
```diff
@@ -75,12 +90,126 @@ public boolean equals(Object o) {
...
+        public Comment read(JsonReader in) throws IOException {
+            final Comment comment = new Comment();
+            in.beginObject();
+            while (in.hasNext()) {
+                String nextName = in.nextName();
+                JsonToken jsonToken = in.peek();
+                if (jsonToken == JsonToken.NULL) {
+                    in.skipValue();
+                    continue;
+                }
```
kvenn: Like-wise could we abstract any of this functionality?

kvenn:
I know this took a bunch of time and provided us with some great research - but I think we should make sure it's super clean and easy to interact with if we're going to merge it into our working branch of vimeo-networking.
The amount of boilerplate needed for every single class is hefty (as you've highlighted before) and it seems like it introduces a fair amount of complexity to adding in fields or additional models. Now it seems like it's all just monotonous work - but I think we should discuss if it's worth it. Or if there are other alternatives to provide us with all the benefit but without all the extra code.
It looked like from your analysis you were able to shave off some time - but we should see if that time gain is enough to warrant the introduced complexity.
I'm hopeful for annotations but I know you looked into that. It just seems all of this functionality should be possible with annotations. Also since this is definitely just an optimization, it might be worth looking into other avenues that buy us the same performance gain. Now that we have raw numbers from the gain from this approach, we can compare it to other approaches after running some tests (maybe jackson, maybe something else entirely).

vimeo-networking/src/main/java/com/vimeo/networking/model/Comment.java
kevinzetterstrom: We could maybe have a static method that we call from within the while loop, but I think that is all we can abstract. Every object is unique, and we can't assume each has the same design. Some could be arrays, some objects, etc.

vimeo-networking/src/main/java/com/vimeo/networking/model/BaseResponseList.java
kevinzetterstrom: Without writing some kind of craziness that writes these TypeAdapters for us, then I am not sure we would be able to get around the monotony and tediousness of writing these methods by hand. I know it sucks. Every article I have come across confirms that it sucks. But if you want type adapters, then we have to go this route.

I could go through and rip out the strings into constants, if you think it warrants that.

vimeo-networking/src/main/java/com/vimeo/networking/model/Comment.java
kevinzetterstrom: I wasn't able to think of a good way since each class that needs a TypeAdapterFactory needs it because it's TypeAdapter needs other typeadapters passed in.

vimeo-networking/src/main/java/com/vimeo/networking/logging/ClientLogger.java
kevinzetterstrom: @kvenn this should address these concerns https://github.com/vimeo/vimeo-networking-java/pull/94/commits/d2b9ef2f10bf281a4c6199dfc6be744fda40d4de

PR #96, opened 2016-06-20
ANDTV-27 Pincode Auth

howardrigberg:
#### Ticket
**Required for Vimeans only**
[ANDTV-27](https://vimean.atlassian.net/browse/ANDTV-27)

#### Ticket Summary
Create the business logic related to pin code authentication in the networking layer. It needs to support the use case where a user is presented with a pin code and goes to a link to enter her/his user credentials while the app polls and endpoint until either the user has logged in or the pin code has expired. Details of the API are here:

https://drive.google.com/a/vimeo.com/file/d/0Bwy64gBzhJ8AUHVQVWFQQ0lqNlE/view

#### Implementation Summary
I added a method to VimeoClient called "logInWithPinCode" to be used in association with pin code based authentication for connected but keyboardless devices such as Android TV. The loginWithPinCode method takes two arguments: "pinCodeCallback" and "authCallback." The loginWithPinCode method will return immediately but will run for an extended period of time asynchronously until a timeout expires, authorization has been successful, or a failure occurs. The expected application flow is that the user is initially presented a pin code (returned in pinCodeCallback), which they then enter into a special page (from a link returned in pinCodeCallback) on Vimeo.com on a different device.  Back on the original device, the app is polling the api to check whether the pin code has been authenticated.  

The "pinCodeCallback" will be called after an initial request to retrieve the pin code and activate link. This method will handle polling the api to check if the code has been activated, and it will ultimately call the completion handler (authCallback) when that happens.  If the pin code expires while we're waiting, completion handler (authCallback) will be called with an error.

I also added a method called "cancelPinCodeLogin" to allow a caller to actively cancel any active pin code polling that might be in progress. 

I cleaned up VimeoClient to meet our coding style (prefixing member variables with "m"), adding annotations for non null parameters, and clearing up some linter errors.

I added a related supporting model class called PinCodeInfo.java and added underlying API calls to VimeoService.java.

#### How to Test
Testing is only possible once this PR is merged and related changes are added to the Vimeo application.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -111,20 +111,20 @@ public final class VimeoClient {
     }
 
     private VimeoClient(@NotNull Configuration configuration) {
-        this.mConfiguration = configuration;
-        this.mGson = VimeoNetworkUtil.getGson();
-        this.mCache = this.mConfiguration.getCache();
-        this.mRetrofit = createRetrofit();
-        this.mVimeoService = mRetrofit.create(VimeoService.class);
+        mConfiguration = configuration;
+        mGson = VimeoNetworkUtil.getGson();
+        mCache = mConfiguration.getCache();
+        mRetrofit = createRetrofit();
+        mVimeoService = mRetrofit.create(VimeoService.class);
 
-        VimeoAccount vimeoAccount = this.mConfiguration.loadAccount();
-        this.setVimeoAccount(vimeoAccount);
+        VimeoAccount vimeoAccount = mConfiguration.loadAccount();
+        setVimeoAccount(vimeoAccount);
     }
 
     private Retrofit createRetrofit() {
         return new Retrofit.Builder().baseUrl(mConfiguration.baseURLString)
                 .client(createOkHttpClient())
-                .addConverterFactory(GsonConverterFactory.create(this.mGson))
+                .addConverterFactory(GsonConverterFactory.create(mGson))
                 .build();
     }
 
@@ -142,10 +142,10 @@ public final class VimeoClient {
                                 .build();
                     }
                 })
-                .setReadTimeout(this.mConfiguration.timeout, TimeUnit.SECONDS)
-                .setConnectionTimeout(this.mConfiguration.timeout, TimeUnit.SECONDS)
+                .setReadTimeout(mConfiguration.timeout, TimeUnit.SECONDS)
+                .setConnectionTimeout(mConfiguration.timeout, TimeUnit.SECONDS)
                 .addInterceptor(
-                        new LoggingInterceptor(this.mConfiguration.debugLogger, this.mConfiguration.logLevel))
+                        new LoggingInterceptor(mConfiguration.debugLogger, mConfiguration.logLevel))
                 .addInterceptor(new Interceptor() {
                     @Override
                     public Response intercept(Chain chain) throws IOException {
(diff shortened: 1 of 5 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -88,45 +96,41 @@
...
-            throw new AssertionError("Configuration cannot be null");
-        }
-
-        this.configuration = configuration;
-        this.gson = VimeoNetworkUtil.getGson();
-        this.cache = this.configuration.getCache();
-        this.retrofit = createRetrofit();
-        this.vimeoService = retrofit.create(VimeoService.class);
+    private VimeoClient(@NotNull Configuration configuration) {
+        this.mConfiguration = configuration;
```
anthonycr: Minor: no longer necessary to use the this keyword here (also could be removed everywhere)

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -611,6 +617,165 @@ public void failure(VimeoError error) {
...
+     * will be called after an initial request to retrieve the pin code and activate link.
+     * This method will handle polling the api to check if the code has been activated, and it will ultimately
+     * call the completion handler (authCallback) when that happens.  If the pin code expires while we're
+     * waiting, completion handler (authCallback) will be called with an error
+     * @param pinCodeCallback {@link ModelCallback} that will receive {@link PinCodeInfo} to display
+     * @param authCallback {@link AuthCallback} that will be notified when Authorization is complete
+     * @return a call object for the pin code request
+     */
+    public Call<PinCodeInfo> logInWithPinCode(@NotNull final ModelCallback<PinCodeInfo> pinCodeCallback,
+                                              @NotNull final AuthCallback authCallback) {
```
anthonycr: Be careful with AuthCallback - problems could arise if it is passed in as an anonymous inner class. If it is passed in as such it will not be held in memory since there will be no hard references to it anywhere once this method finishes returning and it will be garbage collected on the next GC cycle (which could easily happen before the network call finishes)

vimeo-networking/src/main/java/com/vimeo/networking/model/PinCodeInfo.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * A model representing pin code information to be used in association with authorization.
+ * Created by rigbergh on 6/16/16.
+ */
+@SuppressWarnings("unused")
+public class PinCodeInfo {
+
+    @Nullable
+    @SerializedName("device_code")
```
anthonycr: I much prefer this than using `implements Serializable` on the class

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -611,6 +617,165 @@ public void failure(VimeoError error) {
...
+        private final WeakReference<AuthCallback> mAuthCallbackWeakReference;
+        private final WeakReference<VimeoClient> mVimeoClient;
+        private final String mScope;
+
+        PinCodePollingTimerTask(@NotNull PinCodeInfo pinCodeInfo, @NotNull Timer timer, int expiresInSeconds,
+                                @NotNull AuthCallback authCallback, @NotNull VimeoClient client,
+                                @NotNull String scope) {
+            mTimer = timer;
+            mPinCodeInfo = pinCodeInfo;
+            final long SECONDS_TO_NANOSECONDS = 1000000000L;
```
anthonycr: Plz use `TimeUnit.SECONDS.toNanos()` instead

anthonycr:
Merge when ready 👍

anthonycr:
👍

PR #101, opened 2016-06-24
VA-1367 Stag integration for custom type adapters

kevinzetterstrom:
#### Ticket
[VA-1367](https://vimean.atlassian.net/browse/VA-1367)

#### Ticket Summary
Add custom type adapters

#### Implementation Summary
Using the new stag library, adding support for all model objects, except lists. Also, this doesn't support enums yet. For both BaseResponseList types and enums, the default Gson Type Adapter is used (EnumTypeAdapter and Reflective). I created a [ticket](https://vimean.atlassian.net/browse/VA-1793) to update those as well.
Currently, performance on the Nexus 9 is the same as pre-stag, which is just slightly higher than without custom type adapters (by the same percentage higher). I have not tested on the S2 yet, as I am WFH while submitting this PR, and only took the Nexus 9, but I will do profiling on the S2 as well.

#### How to Test
Verify the library compiles for you (will have to init/update the submodule) and the app runs as normal.


settings-parent.gradle
```diff
@@ -0,0 +1,6 @@
+apply from: 'settings.gradle'
+
+rootProject.children.each { project ->
+    String relativeProjectPath = project.projectDir.path.replace(settingsDir.path, "")
+    project.projectDir = new File("vimeo-networking-java/$relativeProjectPath")
+}
```
kvenn: Could you maybe throw in a comment with what's going on here?

settings.gradle
```diff
@@ -1 +1,5 @@
+apply from: 'stag-java/settings-parent.gradle'
+
 include ':vimeo-networking', ':example'
+
+rootProject.children.removeAll {proj -> (proj.name == 'sample')}
```
kvenn: And what's this guy doing?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -41,12 +42,20 @@
...
+    @GsonAdapterKey("token_type")
+    public String tokenType;
+    @GsonAdapterKey("scope")
+    public String scope;
+    @GsonAdapterKey("user")
+    public User user;
     private String userJSON;
 
+    public VimeoAccount(){
+        //constructor for stag TypeAdapter generation
```
kvenn: noice

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601.java
```diff
@@ -98,9 +102,6 @@ public static Date toDate(final String iso8601string) throws ParseException {
         // Account for ISO 8601 standard of using the character 'Z' to represent UTC
         String s = iso8601string.replace("Z", "+00:00");
         try {
-            if (s.charAt(22) != ':') {
-                throw new ParseException("Invalid ISO 8601 format", 0);
-            }
```
kvenn: This won't crash it, right? If we remove the wrong portion of the string below then dateFormat.parse() will probably throw a parse exception which this method already throws.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
```diff
@@ -71,7 +72,8 @@ public static Gson getGson() {
      */
     public static GsonBuilder getGsonBuilder() {
         // Example date: "2015-05-21T14:24:03+00:00"
-        return new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+        return new GsonBuilder().registerTypeAdapterFactory(new Stag.Factory())
+                .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
```
kvenn: If we're annotating every single model then we shouldn't need this naming policy anymore

kevinzetterstrom: No, we shouldn't if we are annotating every model, but right now certain things aren't yet annotated and use reflection, namely, every single list (BaseResponseList) since we haven't solved for abstract types... yet. Leaving this in once those are working certainly does not hurt us.

settings-parent.gradle
anthonycr: This basically fixes the relative paths of nested submodules, so that we can have nested submodules without needing to modify references in the build.gradle files from `compile project(':subproject')` to `compile project('parent:subproject')`

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601.java
kevinzetterstrom: Well we throw a ParseException still, and SimpleDateFormat would throw a ParseException. In either case, nothing different than before - the app will crash unless the caller catches a ParseException.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
kvenn: coolio 👍

PR #102, opened 2016-06-28
VA-1777 new search models, search class

kevinzetterstrom:
#### Ticket
[VA-1777](https://vimean.atlassian.net/browse/VA-1777)

#### Ticket Summary
We are introducing a new search endpoint ```/search``` for whitelisted apps. This ticket is about integrating that into the mobile app.

#### Implementation Summary
I created new models for the search representations, and placed them in a ```model.search``` package. I then created a new Retrofit service, which is accessed through a package-private VimeoClient ```search``` method. The way to access this is through the ```Search``` class.
In order to search you call upon the static ````Search.search()``` method, passing it certain parameters. Similar to the existing ```SearchRefinementBuilder```, there is a ```QueryParameterProvider``` to set parameters and generate a Map.

I left the existing search functionality in place and am not deprecating anything, since this new search is only for whitelisted apps - other apps can continue to use the existing search mechanisms.

I have TODOs sprinkled throughout some of the classes, as I mainly focused on upgrading the existing app functionality to this new search; I left out adding new features - such as Category filtering in app, which is being ticketed for future development.

#### How to Test
Test using the Android PR.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Search.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Search.java
@@ -1,0 +1,286 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016 Vimeo
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package com.vimeo.networking;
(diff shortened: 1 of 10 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/Search.java
```diff
@@ -0,0 +1,286 @@
...
+        refinementMap.put(FILTER_TYPE, type.getText());
+        Map<String, String> queryMap =
+                VimeoClient.getInstance().createQueryMap(query, refinementMap, fieldFilter);
+        // VimeoClient is the end-all interactor with the retrofit service
+        return VimeoClient.getInstance().search(queryMap, callback);
+    }
+
+    public static class QueryParameterProvider {
+
+        private final Map<String, String> mQueryParameters;
```
anthonycr: Could be annotated NotNull, also could be initialized here

vimeo-networking/src/main/java/com/vimeo/networking/Search.java
```diff
@@ -0,0 +1,286 @@
...
+            return this;
+        }
+
+        public QueryParameterProvider setFeaturedVideoCount(int count) {
+            mQueryParameters.put(FILTER_FEATURED_COUNT, String.valueOf(count));
+            return this;
+        }
+
+        public Map<String, String> getQueryParameters() {
+            return mQueryParameters;
```
anthonycr: Could be annotated not null

vimeo-networking/src/main/java/com/vimeo/networking/model/search/FacetOptions.java
```diff
@@ -0,0 +1,46 @@
...
+ * Created by zetterstromk on 6/27/16.
+ */
+public class FacetOptions implements Serializable {
+
+    private static final long serialVersionUID = 6525562797608669182L;
+    @GsonAdapterKey("total")
+    public int mTotal;
+    @Nullable
+    @GsonAdapterKey("uri")
+    public String uri;
```
anthonycr: minor: Lack of spacing here makes these fields hard to read

vimeo-networking/src/main/java/com/vimeo/networking/model/search/SearchResponse.java
```diff
@@ -0,0 +1,46 @@
...
+public class SearchResponse extends BaseResponseList<SearchResult> {
+
+    private static final long serialVersionUID = -7915082057592438294L;
+
+    @SerializedName("facets")
+    public SearchFacetCollection mFacetCollection;
+
+    @Override
+    public Class getModelClass() {
+        return SearchResponse.class;
```
anthonycr: This looks incorrect, in all other places we use BaseResponseList, getModelClass returns the same class as the parameterized type. So in this case it would return SearchResult.class. Could you also modify BaseResponseList so that getModelClass returns Class<T> so that we can catch this sort of error in the future via compiler error?

anthonycr:
Looks good, merge when ready

PR #116, opened 2016-07-26
VA-1794 Support for dynamic category filtering

kevinzetterstrom:
#### Ticket
[VA-1794](https://vimean.atlassian.net/browse/VA-1794)

#### Ticket Summary
Support category filtering

#### Implementation Summary
In order to support category filtering, some api changes were necessary. I accommodate for this in the models and how a search is performed (the parameters).

#### How to Test
Use the accompanying Android PR to test search.


vimeo-networking/src/main/java/com/vimeo/networking/Search.java
```diff
@@ -205,16 +202,27 @@ public static FilterCategory fromString(String text) {
...
         }
         refinementMap.put(FILTER_TYPE, type.getText());
+        if (facets != null) {
+            String facetsToUse = "";
+            for (int i = 0; i < facets.size(); i++) {
+                facetsToUse += facets.get(i).getText();
+                if (i + 1 != facets.size()) {
+                    facetsToUse += ",";
+                }
+            }
```
kvenn: Here's a pretty clever way I just found to comma delimit a list (since it never feels like a very pretty thing): 
```java
String delim = "";
for (Item i : list) {
    sb.append(delim).append(i);
    delim = ",";
}
```

http://stackoverflow.com/questions/668952/clearest-way-to-comma-delimit-a-list-java

vimeo-networking/src/main/java/com/vimeo/networking/model/search/SearchFacetCollection.java
```diff
@@ -35,27 +35,62 @@
...
  * Created by zetterstromk on 6/27/16.
  */
+@SuppressWarnings("unused")
 public class SearchFacetCollection implements Serializable {
 
     private static final long serialVersionUID = 3340976215489066653L;
 
     @Nullable
     @GsonAdapterKey("type")
     public SearchFacet mTypeFacet;
```
kvenn: Can we make these fields private?

kvenn:
Before merging I'd just like a quick overview of how the search facets work 📈

vimeo-networking/src/main/java/com/vimeo/networking/model/search/SearchFacetCollection.java
kevinzetterstrom: Unfortunately no, Stag needs them to be public. There is hope to make Stag work with package private members in the future, but for now, they must be public.

vimeo-networking/src/main/java/com/vimeo/networking/Search.java
kevinzetterstrom: 👍 Nice, I'll go with this, it's clean. I also considered using StringUtils (its a sun package) that can do the same, but I opted to not include any extra imports and write this sucker by hand.

PR #117, opened 2016-07-27
DTA-31 add lazy library loading

kevinzetterstrom:
#### Ticket
[DTA-31](https://vimean.atlassian.net/browse/DTA-31)

#### Ticket Summary
For Day Tripper, we want to combine the splash and launch screen, effectively removing the splash screen. A part of this is then to defer library initialization from application onCreate until the launch screen, effectively letting the "splash" show quicker, and then hiding once initialization completes.

#### Implementation Summary
I created a lazy load initialization, which effectively waits until the first call to ```getInstance``` to initialize. Also, I fixed an input stream that needed to be closed.

#### How to Test
Using method tracing, see how app startup time is reduced

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -115,20 +115,20 @@ public final class VimeoClient {
     }
 
     private static LazyInitializationProvider sLazyInitializationProvider;
-    private static VimeoClient mSharedInstance;
+    private static VimeoClient sSharedInstance;
 
     public static VimeoClient getInstance() {
-        if (mSharedInstance == null) {
+        if (sSharedInstance == null) {
             if (sLazyInitializationProvider != null) {
                 initialize(sLazyInitializationProvider.getConfiguration());
-                if (mSharedInstance != null) {
-                    return mSharedInstance;
+                if (sSharedInstance != null) {
+                    return sSharedInstance;
                 }
             }
             throw new AssertionError("Instance must be configured before use");
         }
 
-        return mSharedInstance;
+        return sSharedInstance;
     }
 
     public static void initializeLazily(@NotNull Configuration configuration) {
@@ -136,7 +136,7 @@ public final class VimeoClient {
     }
 
     public static void initialize(@NotNull Configuration configuration) {
-        mSharedInstance = new VimeoClient(configuration);
+        sSharedInstance = new VimeoClient(configuration);
     }
 
     private VimeoClient(@NotNull Configuration configuration) {
@@ -801,7 +801,7 @@ public final class VimeoClient {
                 VimeoClient.mContinuePinCodeAuthorizationRefreshCycle = true;
                 mPinCodeAuthorizationTimer.scheduleAtFixedRate(
                         new PinCodePollingTimerTask(pinCodeInfo, mPinCodeAuthorizationTimer,
-                                                    pinCodeInfo.getExpiresIn(), authCallback, mSharedInstance,
+                                                    pinCodeInfo.getExpiresIn(), authCallback, sSharedInstance,
                                                     SCOPE), 0,
                         SECONDS_TO_MILLISECONDS * pinCodeInfo.getInterval());
             }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -99,17 +99,43 @@ public final class VimeoClient {
      */
     // <editor-fold desc="Configuration">
 
+    private static class LazyInitializationProvider {
+
+        @NotNull
+        private final Configuration mConfiguration;
+
+        public LazyInitializationProvider(@NotNull Configuration configuration) {
+            mConfiguration = configuration;
+        }
+
+        @NotNull
+        public Configuration getConfiguration() {
+            return mConfiguration;
+        }
+    }
+
+    private static LazyInitializationProvider sLazyInitializationProvider;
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -99,17 +99,43 @@
...
+                }
+            }
             throw new AssertionError("Instance must be configured before use");
         }
 
         return mSharedInstance;
     }
 
-    public static void initialize(Configuration configuration) {
+    public static void initializeLazily(@NotNull Configuration configuration) {
```
anthonycr: Do we need to expose this to the consumer of this library? If we just replace the `initialize` method with this implementation, then lazy loading will work with existing code (the get instance code will need to create the new instance instead of calling initialize)

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -99,17 +99,43 @@
...
+        }
+
+        @NotNull
+        public Configuration getConfiguration() {
+            return mConfiguration;
+        }
+    }
+
+    private static LazyInitializationProvider sLazyInitializationProvider;
     private static VimeoClient mSharedInstance;
```
anthonycr: Good opportunity to rename this to `sSharedInstance`

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -99,17 +99,43 @@
...
 
     public static VimeoClient getInstance() {
```
anthonycr: We should probably synchronize this `getInstance` method so that it can safely be called from multiple threads

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: This is exactly why I added all 3 of you to the ticket - should we just go with lazy loading over the current ```initialize``` approach? This does force all consumers to effectively lazy load until their first ```getInstance``` call, but is that a bad thing?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: Nice callout

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: Another great call!

kvenn:
While I think we definitely need to support some form of lazy initialization, I think there might be a better way to do it that all consumers of the library can benefit from.

Without the lazy, we were initializing everything in onCreate of the app which had the splash screen displaying - this is a situation where dropped frames don't matter (as much). But now if we're doing it lazily like this, it's unpredictable when we'll drop frames when an expensive operation is done on the UI thread.

With this lazy init, it's possible that you're making your first request with the library at a time when an animation is happening (a progress bar, and app bar layout, or even a user tries to scroll) and then the UI will be halted for 20ms, 40ms, 100ms, 200ms, or whatever. 10ms on the 6P may translate to 200ms on a way worse phone.

I think the two options in a situation like this are to either take the hit and do it initially (predictable but slow), or do the operation on the background and be prepared to handle that (more complicated but more predictable and better UX).

Since the real hit is from loading up the cert pinning key, I think we should do that on a background thread and have the VimeoClient be aware of the state of this load operation. This will get kicked off the second the VimeoClient is initialized and then by the time they make their first request, it may already be done.
In the case where it's not done by the first request (which is when we need the key), since our network requests always get done on the background thread, we could probably just pair the load operation with the first asynchronous call we make (if it isn't done by that time) and the user will be none the wiser. Maybe with a countdown latch.

kevinzetterstrom:
@kvenn I like the approach have delaying it asynchronously in the library, but how do you handle synchronous calls? In this case we would need to block until the initialization finishes, and really then, the only place to do it effectively is in ```getInstance```.
This was a pretty quick and dirty way to allow clients to defer the initialization until ```getInstance```. Yes, I realize that may result in dropped frames, same as it would if you called initialize like we currently do, and the same as if you called the library while it still hasn't initialized asynchronously and was blocking.
My initial instinct for this approach was that we could hold off on initialization until we knew when it would have the least impact on affecting our UI, which is our first call to ```getInstance```. This proves valuable in this use-case and doesn't affect any other users (including the Vimeo app) who are still using ```initialize``` in the app onCreate. No matter what we dream up, I am sure it won't be the best solution for _every_ consumer, so we need to find a way that works great for as many as possible.

anthonycr:
Here's an idea I have... it looks bad, but does everything we want it to do:

```
private static VimeoClient sSharedInstance;

    public static synchronized VimeoClient getInstance() {
        synchronized (sSharedInstance) {
            if (sSharedInstance == null) {
                throw new AssertionError("Instance must be configured before use");
            }
        }

        return sSharedInstance;
    }

    public static void initialize(final Configuration configuration) {
        if (sSharedInstance == null) {
            Executors.newSingleThreadExecutor().execute(new Runnable() {
                @Override
                public void run() {
                    synchronized (sSharedInstance) {
                            sSharedInstance = new VimeoClient(configuration);
                        }
                    }
                }
            });
        }
    }
```

anthonycr:
The real fix would be to use a reactive approach 😎

kevinzetterstrom:
Let's discuss at our Android Sync-Up tomorrow. We need a solid fix to this problem. This solution is cheap but not the best.

PR #119, opened 2016-08-02
VA-1712 robust parsing handling

kevinzetterstrom:
#### Ticket
[VA-1712](https://vimean.atlassian.net/browse/VA-1712)

#### Ticket Summary
Add more robust logic around parsing objects from the API

#### Implementation Summary
While stag adds extra parsing protection, we need to catch errors being thrown by Gson still. The networking layer has a logging system that will use a provider logger or system.out; we can use that to log any errors encountered in stag. 

#### How to Test
Change an object to an array, or vice versa.


anthonycr:
Looks good, merge it 👍

PR #124, opened 2016-08-08
VA-1881 adding sync method for client creds

kevinzetterstrom:
#### Ticket
[VA-1881](https://vimean.atlassian.net/browse/VA-1881)

#### Ticket Summary
BlinkFeed content never loads

#### Implementation Summary
The issue is that when fresh installed and never going into the app, a client credentials token is never issued - thus leaving BlinkFeed to try and make requests using basic auth, which fail.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -356,6 +356,7 @@ public final class VimeoClient {
      * WARNING: This contains a synchronous network call. Use {@link #authorizeWithClientCredentialsGrant(AuthCallback)}
      * for asynchronous use.
      */
+    @Nullable
     public VimeoAccount authorizeWithClientCredentialsGrantSync() {
 
         Call<VimeoAccount> call = mVimeoService.authorizeWithClientCredentialsGrant(getBasicAuthHeader(),
@@ -365,7 +366,7 @@ public final class VimeoClient {
         VimeoAccount vimeoAccount = null;
         try {
             retrofit2.Response<VimeoAccount> response = call.execute();
-             if (response.isSuccessful()) {
+            if (response.isSuccessful()) {
                 vimeoAccount = response.body();
                 saveAccount(vimeoAccount, null);
             }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -348,6 +348,35 @@ public final class VimeoClient {
     }
 
     /**
+     * Authorizes users of the app who are not signed in. This call requires a client id and client secret
+     * to be set on the initial Configuration.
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -348,6 +348,35 @@ public String getCodeGrantAuthorizationURI() {
...
+            retrofit2.Response<VimeoAccount> response = call.execute();
+             if (response.isSuccessful()) {
+                vimeoAccount = response.body();
+                saveAccount(vimeoAccount, null);
+            }
+        } catch (IOException e) {
+            ClientLogger.e("Exception during authorizeWithClientCredentialsGrantSync: " + e.getMessage(), e);
+        }
+
+        return vimeoAccount;
```
kvenn: I would suggest annotating this method nullable since it'll return null in the event of an IOException. Or have the method throw an exception - I think either would get the point across.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -348,6 +348,35 @@ public String getCodeGrantAuthorizationURI() {
...
+        Call<VimeoAccount> call = mVimeoService.authorizeWithClientCredentialsGrant(getBasicAuthHeader(),
+                                                                                    Vimeo.CLIENT_CREDENTIALS_GRANT_TYPE,
+                                                                                    mConfiguration.scope);
+
+        VimeoAccount vimeoAccount = null;
+        try {
+            retrofit2.Response<VimeoAccount> response = call.execute();
+             if (response.isSuccessful()) {
+                vimeoAccount = response.body();
+                saveAccount(vimeoAccount, null);
```
kvenn: We might want to pull this block out into a helper just so it's consistent with the async version:

```java
            if (vimeoAccount.getUser() != null && (mEmail == null || mEmail.isEmpty())) {
                // We must always have a `name` field, which is used by the Android Account Manager for
                // display in the device Settings -> Accounts [KZ] 12/17/15
                String name = (vimeoAccount.getUser().name !=
                               null) ? vimeoAccount.getUser().name : vimeoAccount.getUser().uri;
                mClient.saveAccount(vimeoAccount, name);
            } else {
                mClient.saveAccount(vimeoAccount, mEmail);
            }
```

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: Nice call, adding the annotation

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: We will never have a user or email, so there is not consistency that can benefit from a helper.

kvenn:
merge 💯

PR #126, opened 2016-08-15
VA-1741 add Recommendation model, add categories connection for User

kevinzetterstrom:
#### Ticket
[VA-1741](https://vimean.atlassian.net/browse/VA-1741)

#### Ticket Summary
Add the Recommendation response, as defined [here](https://github.vimeows.com/Vimeo-API/api/issues/127).
Add a categories connection as defined [here](https://github.vimeows.com/Vimeo-API/api/issues/150).

#### Implementation Summary
Added the connection, created the model, its response list, and a simple unit test for the ```NotNull``` model method.

#### How to Test
Run the unit test.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Recommendation.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Recommendation.java
@@ -1,0 +1,73 @@
+package com.vimeo.networking.model;
+
+import com.vimeo.stag.GsonAdapterKey;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+
+/**
+ * A model that contains recommendations for a User to follow.
+ * A recommendation can either be a {@link User} or {@link Channel}
+ * <p/>
+ * Created by zetterstromk on 8/15/16.
+ */
+public class Recommendation implements Serializable {
+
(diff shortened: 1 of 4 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Recommendation.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Null if not recommended by a category
+     */
+    @Nullable
+    @GsonAdapterKey("category")
+    public Category mCategory;
+
+    @Nullable
+    @GsonAdapterKey("type")
+    public String mRecommendationType;
```
anthonycr: Minor, could use a StringDef to enforce correct assignment... although since this will be set by the server response it most likely won't do much good

anthonycr:
Looks good, check out my comment, if it doesn't see like it will provide any benefit, just merge it

vimeo-networking/src/main/java/com/vimeo/networking/model/Recommendation.java
kevinzetterstrom: Isn't that an Android Annotation... ?

anthonycr: lol yes, forgot this was networking

PR #128, opened 2016-08-18
DTA-26 add single sign on token exchange

kevinzetterstrom:
#### Ticket
[DTA-26](https://vimean.atlassian.net/browse/DTA-26)

#### Ticket Summary
When introducing multiple apps that may be authenticated, we need a way to exchange tokens from one app for another, so users do not need to enter their credentials again.

#### Implementation Summary
The API provides a single-sign-on token exchange endpoint, for use by Vimeo's iOS and Android apps. I added a client and service method to support this.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -397,6 +397,30 @@ public final class VimeoClient {
         return call;
     }
 
+    /**
+     * This method is used to exchange a user access token from app A for a token for app B.
+     * <p/>
+     * An app can use this to exchange an authenticated user's access token from app A for a new access
+     * token for app B, using the scope associated with app B. The caller of this method will be app B.
+     * <p/>
+     * <b>This method is application restricted, and not meant to be consumed by the general public.</b>
+     *
+     * @param token    the authenticated user access token from application A. This <b>cannot</b> be a client
+     *                 credentials access token.
+     * @param callback Callback invoked upon success/fail of the service
+     * @return A Call object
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -397,6 +397,30 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
...
+    public Call<VimeoAccount> singleSignOnTokenExchange(@NotNull String token, AuthCallback callback) {
+        if (callback == null) {
+            throw new AssertionError("Callback cannot be null");
+        }
+
+        Call<VimeoAccount> call =
+                mVimeoService.ssoTokenExchange(getBasicAuthHeader(), token, mConfiguration.scope);
+        call.enqueue(new AccountCallback(this, callback));
+        return call;
+    }
```
kvenn: We haven't officially found a way to prevent people from using a thing, but we could maybe do what the android folks do and throw @hide on it.

http://stackoverflow.com/a/17056643/1759443

kvenn:
Just made one comment. Merge when ready 👍

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: We will need to add https://bintray.com/bintray/jcenter/com.google.doclava:doclava and we should go through the whole repo and add ```@hide``` on methods and even entire files. I am scoping out the work for a 2.0 upgrade path this sprint, and can add this to the list.

PR #130, opened 2016-08-24
VA-1929 account for DRM files

kevinzetterstrom:
#### Ticket
[VA-1929](https://vimean.atlassian.net/browse/VA-1929)

#### Ticket Summary
We need to account for DRM files when counting files

#### Implementation Summary
Simple count++ when files exist


anthonycr:
approved 👍

PR #132, opened 2016-08-25
Merge v1.1.0 into master for release

kevinzetterstrom:
v1.1.0 is ready to go! It needs to be merged into master, after which master can be pushed to bintray/jcenter, tagged, and added as a release with notes.

**Breaking Changes**
* Removed DebugLoggerInterface and DebugLogger in favor of LogProvider and ClientLogger - no need to pass in a DebugLoggerInterface to the Configuration any longer - the library will fall back onto ClientLogger
* Removed VimeoCallback setCall method - with newer Retrofit version there is no need
* Interaction "addedDate" renamed to "addedTime" - the former would never have had data
* Privacy comments moved from String to PrivacyValue enum
* Removed VideoLog from Video, and moved the class - this is intended to only be used by the offical Vimeo apps

**New**
* Added Stag for custom type adapters
* Updated Retrofit
* Allow adding additional interceptors (can be useful for adding Stetho)
* Better debugging - new ClientLogger does not require a LogProvider (previously DebugLoggerInterface) to be provided
* Deprecated some methods
* Added models and more members to certain models


kvenn:
Merge when ready!

PR #135, opened 2016-08-26
VA-1741 adding post service that takes an array

kevinzetterstrom:
#### Ticket
[VA-1741](https://vimean.atlassian.net/browse/VA-1741)

#### Ticket Summary
batch following of users/channels

#### Implementation Summary
The API request for the batch follow takes an array as the body, so we need support  for this.


anthonycr:
m e r g e i t

PR #136, opened 2016-08-28
ANDTV-168 Channels

howardrigberg:
#### Ticket
**Required for Vimeans only**
[ANDTV-168](https://vimean.atlassian.net/browse/ANDTV-168)

#### Ticket Summary
Add support for displaying channels on User Pages.

#### Implementation Summary
Added two convenience methods for getting channel connection data.

#### How to Test
Nothing to test at this point.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
@@ -179,7 +179,7 @@ public class User implements Serializable {
     }
 
     @Nullable
-    public Connection getChannelsConnection() {
+    public Connection getFollowedChannelsConnection() {
         if (metadata != null && metadata.connections != null && metadata.connections.channels != null) {
             return metadata.connections.channels;
         }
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
```diff
@@ -179,6 +179,21 @@ public int getLikesCount() {
     }
 
     @Nullable
+    public Connection getChannelsConnection() {
+        if (metadata != null && metadata.connections != null && metadata.connections.channels != null) {
+            return metadata.connections.channels;
```
kevinzetterstrom: These are the channels that a user follows. Is this what we want to display on the user profile?... because I could have sworn we only wanted to show channels that the user helped curate. There is a ```moderated_channels``` connection for this - I think you should add that as part of this PR in addition to this, it would be good to have both available on the User model.

howardrigberg: Great catch. I'll add it.

howardrigberg:
Added moderated channels. https://github.com/vimeo/vimeo-networking-java/commit/d48ff3d57a7eddaf7f2f534dc86f56926dc52b1c

kevinzetterstrom:
👍 merge away

PR #139, opened 2016-08-31
VA-1898 Add followable interface

anthonycr:
#### Ticket
[VA-1898](https://vimean.atlassian.net/browse/VA-1898)

#### Ticket Summary
While building onboarding, I needed a common interface for an item to be "Followable" so that I could execute follows on any one of several items.

#### Implementation Summary
Added the Followable interface that is implemented by Channels, Categories, and Users. Possibly it should be implemented by more model classes. Also, let me know if there should be any more methods that are a part of the Followable interface.


kvenn:
Looks good! Very cool idea - a lot of our models have a ton of common functionality so we might be able to find more ways to unify them.

Merge when ready

PR #143, opened 2016-09-09
Smarter error handling

anthonycr:
#### Summary
Sometimes when we get an error back from a response, we have a non null but empty error response body. In this case, we throw an exception (and catch it) as we try to parse the empty error.

#### Implementation Summary
I am tired of seeing this exception in my logs, so to fix it, I am checking whether the error response is empty, and not trying to parse it if it is. The reason it is empty is that we try to pull from cache and get an error if it is empty or invalid, and since the cache is not sending any sort of response code, the error response is empty.

#### How to Test
From a cold start with cache cleared, start the app. Without this change, two or three exceptions will be logged to `System.out` as the error converter is unable to convert empty errors.


kevinzetterstrom:
👍  Nice, I am also tired of seeing this! Merge when ready

PR #148, opened 2016-09-26
VA-1963 wrapping date deserialization

anthonycr:
#### Ticket
[VA-1963](https://vimean.atlassian.net/browse/VA-1963)

#### Ticket Summary
The date deserialization is crashing sometimes because the API seems to send us bad dates.

#### Implementation Summary
Wrap the deserialization in a try catch so we can choose whether or not to crash on the client side.

#### How to Test
make sure dates still show up correctly

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/logging/LoggingInterceptor.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/logging/LoggingInterceptor.java
@@ -23,12 +23,12 @@
 package com.vimeo.networking.logging;
 
 import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import com.vimeo.networking.Configuration;
 import com.vimeo.networking.Vimeo.LogLevel;
 import com.vimeo.networking.VimeoClient;
+import com.vimeo.networking.utils.VimeoNetworkUtil;
 
 import java.io.IOException;
 
@@ -125,7 +125,7 @@ public class LoggingInterceptor implements Interceptor {
                 JsonParser parser = new JsonParser();
                 JsonObject json = parser.parse(jsonString).getAsJsonObject();
 
-                Gson gson = new GsonBuilder().setPrettyPrinting().create();
+                Gson gson = VimeoNetworkUtil.getGsonBuilder().setPrettyPrinting().create();
                 prettyString = gson.toJson(json);
             } catch (Exception e) {
                 ClientLogger.e("Error making body pretty response/request");
--- a/vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
@@ -1,0 +1,55 @@
+package com.vimeo.networking.utils;
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+import com.google.gson.internal.bind.util.ISO8601Utils;
+import com.vimeo.networking.logging.ClientLogger;
+
+import java.lang.reflect.Type;
+import java.text.ParseException;
+import java.text.ParsePosition;
+import java.util.Date;
+
+/**
+ * A wrapper class that serializes and deserializes dates
+ * using the {@link ISO8601Utils} class that catches errors
+ * and reports them to the client application, allowing
+ * them to decide if they should crash or not. If we just
+ * rely on the default adapter used by JSON, we are unable
+ * to absorb date parsing errors or log them correctly.
+ */
+@SuppressWarnings("WeakerAccess")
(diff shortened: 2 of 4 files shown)


howardrigberg:
I will look at this later today

vimeo-networking/src/main/java/com/vimeo/networking/logging/LoggingInterceptor.java
```diff
@@ -125,7 +125,7 @@ private String toPrettyFormat(String jsonString) {
                 JsonParser parser = new JsonParser();
                 JsonObject json = parser.parse(jsonString).getAsJsonObject();
 
-                Gson gson = new GsonBuilder().setPrettyPrinting().create();
+                Gson gson = VimeoNetworkUtil.getGsonBuilder().setPrettyPrinting().create();
```
kvenn: 👍

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
```diff
@@ -0,0 +1,55 @@
...
+import java.text.ParseException;
+import java.text.ParsePosition;
+import java.util.Date;
+
+/**
+ * A wrapper class that serializes and deserializes dates
+ * using the {@link ISO8601Utils} class that catches errors
+ * and reports them to the client application, allowing
+ * them to decide if they should crash or not. If we just
+ * rely on the default adapter used by JSON, we are unable
```
kvenn: Should this say Gson?

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
```diff
@@ -0,0 +1,55 @@
...
+ * to absorb date parsing errors or log them correctly.
```
kvenn: Also it wasn't just logging them incorrectly.....it was serializing it incorrectly.

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
anthonycr: yes

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
anthonycr: yeah I'll update these docs to reflect the new findings

anthonycr:
Gson bug filed here https://github.com/google/gson/issues/935

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public static JsonDeserializer<Date> getDateDeserializer() {
+        return new JsonDeserializer<Date>() {
+            @Override
+            public Date deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+                    throws JsonParseException {
+                try {
+                    return json == null ? null : ISO8601Utils.parse(json.getAsString(), new ParsePosition(0));
+                } catch (ParseException e) {
+                    ClientLogger.e("Incorrectly formatted date sent from server: " + json.getAsString(), e);
```
howardrigberg: See above, but maybe be a little more careful with getAsString?

vimeo-networking/src/main/java/com/vimeo/networking/logging/LoggingInterceptor.java
howardrigberg: Do you think it would help to actually log the exception in the log below? Does it have any useful info?

vimeo-networking/src/main/java/com/vimeo/networking/utils/ISO8601Wrapper.java
```diff
@@ -0,0 +1,55 @@
...
+                return src == null ? null : new JsonPrimitive(ISO8601Utils.format(src));
+            }
+        };
+    }
+
+    public static JsonDeserializer<Date> getDateDeserializer() {
+        return new JsonDeserializer<Date>() {
+            @Override
+            public Date deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+                    throws JsonParseException {
```
howardrigberg: It doesn't actually throw a JsonParseException, so maybe ditch the the "throws"? It'll still be a legit overrride.

Also, if you just look at the javadocs for getAsString there are other types of exceptions that can occur here due to getAsString. Maybe we should swallow more than just ParseException? But perhaps more importantly, unless I am missing something...

Here's the weird thing regarding getAsString, in the version of gson that I have, this method always throws an UnsupportedOperationException. I'm confused...

anthonycr: Regarding unsupported operation exception, it's because JsonElement is an abstract class, so the implementation is hidden

vimeo-networking/src/main/java/com/vimeo/networking/logging/LoggingInterceptor.java
anthonycr: I don't know

howardrigberg: Maybe just log it then, it couldn't hurt. It is just one parameter...

PR #149, opened 2016-10-04
VA-1984 add VOD season models, move VOD models, add VideoBadge

kevinzetterstrom:
#### Ticket
[VA-1984](https://vimean.atlassian.net/browse/VA-1984)

#### Ticket Summary
For VOD Season support, we need to add some models. Also, as a bonus, add the Video Badge representation.

#### Implementation Summary
I created 3 models - a ```Season```, ```SeasonList```, and ```VideoBadge```. Since the main model package is growing and we want to do a bunch of refactoring for 2.0.0, I made a vod package for the season models, and moved ```VodItem``` and ```VodList``` into it.

#### How to Test
We aren't using the new models anywhere yet, but verify that the app doesn't crash with the package moves - use the mobile PR for that.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/VideoBadge.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/VideoBadge.java
@@ -1,0 +1,91 @@
+package com.vimeo.networking.model;
+
+import com.vimeo.stag.GsonAdapterKey;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+
+/**
+ * A model representing a "badge" that a video can display. The badge represents
+ * an award, and is usually displayed over the thumbnail of the video, or near
+ * video data (name, description, etc)
+ * Created by zetterstromk on 10/4/16.
+ */
+public class VideoBadge implements Serializable {
+
+    private static final long serialVersionUID = -5343389171512787927L;
+
+    @Nullable
+    @GsonAdapterKey("type")
+    public String mType;
+
+    @Nullable
+    @GsonAdapterKey("festival")
+    public String mFestival;
+
+    @Nullable
+    @GsonAdapterKey("link")
+    public String mLink;
+
+    @Nullable
+    @GsonAdapterKey("text")
+    public String mText;
+
+    @Nullable
+    @GsonAdapterKey("pictures")
+    public PictureCollection mPictureCollection;
+
+    // -----------------------------------------------------------------------------------------------------
+    // Getters
+    // -----------------------------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/VideoBadge.java
```diff
@@ -0,0 +1,91 @@
...
+    @GsonAdapterKey("pictures")
+    public PictureCollection mPictureCollection;
+
+    // -----------------------------------------------------------------------------------------------------
+    // Getters
+    // -----------------------------------------------------------------------------------------------------
+    // <editor-fold desc="Getters">
+
+    /**
+     * A type of the badge, such as "staffpick", "vod", or "weekendchallenge". These are
```
kevinzetterstrom: It may not be used, however, since this is for the public API, I think we should add all of the available fields.

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoBadge.java
```diff
@@ -0,0 +1,91 @@
...
+        return mFestival;
+    }
+
+    /**
+     * @return A link to this content - may be generic (such as "https://vimeo.com/ondemand)
+     * or specific (such as "https://vimeo.com/channels/staffpicks/12345")
+     */
+    @Nullable
+    public String getLink() {
+        return mLink;
```
kevinzetterstrom: I mentioned this in the javadoc of this method: this may be a generic link, such as "https://vimeo.com/ondemand" - that would be an entirely different link than provided by the video. Again, same as with the type, I want to provide as much coverage as possible.

kevinzetterstrom:
@anthonycr about the missing ```equals``` and ```hashcode``` methods - we need a good way to determine that they are equal before implementing these methods. For most objects, like Video, and User, for example, we use the resource key or uri - something that is unique and will not change. In the case of these new objects, there is no resource key (and with VideoBadge, no uri either). So for VideoBadge, how are two badges equal? They could have the same type but have different festivals. Or same links but different types perhaps. For that reason I left them off. I could use the uri for the Season object if you feel strongly about adding it.

anthonycr:
@kevinzetterstrom could you write a ticket and for writing robust equals/hashcode methods? We will need proper functionality in the future for properly diffing lists of items. As for equals on things like the badge object, we would just compare all the fields in the object. We should eventually migrate away from using .equals to just check if two objects are the same uri and more toward actual equality with matching fields, etc. 

- [ ] add ticket for proper equals/hashcode

Looks good, merge when ready

kevinzetterstrom:
I created VA-2024 to make our comparison methods more robust

PR #151, opened 2016-10-05
VA-1983 Add season name helper method to Video

kevinzetterstrom:
#### Ticket
[VA-1983](https://vimean.atlassian.net/browse/VA-1983)

#### Ticket Summary
We want to display a VOD Season name in the VOD list, this is about creating a helper for that.

#### Implementation Summary
I added a simple method to get the season name, and I moved the Play Status into the correct fold.


anthonycr:
looks güd 👍

PR #152, opened 2016-10-11
VA-1728 Improving account manager code

kevinzetterstrom:
#### Ticket
[VA-1728](https://vimean.atlassian.net/browse/VA-1728)

#### Ticket Summary
We need to make a few adjustments to the networking code to improve auth within the apps

#### Implementation Summary
The main aspect of this ticket was to create a synchronous version of the single sign on method, that can be passed a basic auth string. This is needed since this will be called from a background thread within the Android AccountManager system (getAuthToken), and we need the basic auth header of the requesting app, not the basic auth header of the running app - the running app is always the first installed app with a Vimeo AccountManager (authenticator)

Side effect are cleaning up some of the deprecated methods around the AccountStore, and adding annotation.s

#### How to Test
Use the VA-1728 branches in vimeo-kit and VimeoMobile to test

### DIFF
--- a/example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
+++ b/example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
@@ -5,6 +5,7 @@ import android.accounts.AccountManager;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.annotation.RequiresPermission;
 
 import com.vimeo.android.networking.example.AccountPreferenceManager;
@@ -34,18 +35,20 @@ public class TestAccountStore implements AccountStore {
 //        mAccountManager = AccountManager.get(context);
     }
 
+    @Nullable
     @Override
     public VimeoAccount loadAccount() {
         return AccountPreferenceManager.getClientAccount();
     }
 
     @Override
-    public void saveAccount(@NonNull VimeoAccount vimeoAccount) {
+    public void saveNonUserAccount(@NonNull VimeoAccount vimeoAccount) {
         AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
     @Override
-    public void saveAccount(@NonNull VimeoAccount vimeoAccount, @NonNull String email) {
+    public void saveAuthenticatedUserAccount(@NonNull VimeoAccount vimeoAccount,
+                                             @NonNull String accountName) {
         AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
--- a/example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
+++ b/example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
@@ -42,13 +42,15 @@ public class TestAccountStore implements AccountStore {
     }
 
     @Override
-    public void saveNonUserAccount(@NonNull VimeoAccount vimeoAccount) {
+    public void saveAuthenticatedUserAccount(@NonNull VimeoAccount vimeoAccount,
+                                             @NonNull String accountName) {
+        // In non-demo code, you may want to store the account name. It is left off in
+        // this demo, but may be used as a key (such as in the Android Account Manager).
         AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
     @Override
-    public void saveAuthenticatedUserAccount(@NonNull VimeoAccount vimeoAccount,
-                                             @NonNull String accountName) {
+    public void saveNonUserAccount(@NonNull VimeoAccount vimeoAccount) {
         AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
@@ -44,12 +44,23 @@ public interface AccountStore {
 
     /**
      * Saves a {@link VimeoAccount} for loading later.
+     * <p>
+     * Use this method for saving authenticated users that have an account name.
      *
      * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
      * @param accountName  the <i>name</i> of the account that helps uniquely identify it in account
      *                     stores, such as the Android AccountManager
      */
-    void saveAccount(@NotNull VimeoAccount vimeoAccount, String accountName);
+    void saveAccount(@NotNull VimeoAccount vimeoAccount, @NotNull String accountName);
+
+    /**
+     * Saves a {@link VimeoAccount} for loading later.
+     * <p>
+     * Use this method for saving client credential token when there is not account name.
+     *
+     * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
+     */
+    void saveAccount(@NotNull VimeoAccount vimeoAccount);
 
     /**
      * Removes the saved {@link VimeoAccount} from storage
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
@@ -51,7 +51,7 @@ public interface AccountStore {
      * @param accountName  the <i>name</i> of the account that helps uniquely identify it in account
      *                     stores, such as the Android AccountManager
      */
-    void saveAccount(@NotNull VimeoAccount vimeoAccount, @NotNull String accountName);
+    void saveAuthenticatedUserAccount(@NotNull VimeoAccount vimeoAccount, @NotNull String accountName);
 
     /**
      * Saves a {@link VimeoAccount} for loading later.
@@ -60,7 +60,7 @@ public interface AccountStore {
      *
      * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
      */
-    void saveAccount(@NotNull VimeoAccount vimeoAccount);
+    void saveNonUserAccount(@NotNull VimeoAccount vimeoAccount);
 
     /**
      * Removes the saved {@link VimeoAccount} from storage
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
@@ -46,7 +46,7 @@ public interface AccountStore {
      * Saves a {@link VimeoAccount} for loading later.
      *
      * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
-     * @param accountName  the <i>name</i> of the account that helps unique identify it in account
+     * @param accountName  the <i>name</i> of the account that helps uniquely identify it in account
      *                     stores, such as the Android AccountManager
      */
     void saveAccount(@NotNull VimeoAccount vimeoAccount, String accountName);
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
@@ -56,7 +56,7 @@ public interface AccountStore {
     /**
      * Saves a {@link VimeoAccount} for loading later.
      * <p>
-     * Use this method for saving client credential token when there is not account name.
+     * Use this method for saving client credential token when there is no account name.
      *
      * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
      */
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
@@ -34,10 +34,27 @@ import org.jetbrains.annotations.Nullable;
  */
 public interface AccountStore {
 
+    /**
+     * Load the previously saved {@link VimeoAccount} and return it
+     *
+     * @return null if no VimeoAccount could be loaded
+     */
     @Nullable
     VimeoAccount loadAccount();
 
-    void saveAccount(@NotNull VimeoAccount vimeoAccount, String email);
+    /**
+     * Saves a {@link VimeoAccount} for loading later.
+     *
+     * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
+     * @param accountName  the <i>name</i> of the account that helps unique identify it in account
+     *                     stores, such as the Android AccountManager
+     */
+    void saveAccount(@NotNull VimeoAccount vimeoAccount, String accountName);
 
+    /**
+     * Removes the saved {@link VimeoAccount} from storage
+     *
+     * @param vimeoAccount the account to delete
+     */
     void deleteAccount(@NotNull VimeoAccount vimeoAccount);
 }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
@@ -24,6 +24,9 @@ package com.vimeo.networking;
 
 import com.vimeo.networking.model.VimeoAccount;
 
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
 /**
  * Interface responsible for handling the creation, deletion, and loading of Vimeo accounts on the client.
  * <p/>
@@ -31,18 +34,10 @@ import com.vimeo.networking.model.VimeoAccount;
  */
 public interface AccountStore {
 
+    @Nullable
     VimeoAccount loadAccount();
 
-    /**
-     * @deprecated use {@link #saveAccount(VimeoAccount, String)} instead
-     * <p/>
-     * We find no use in storing the password when you can persist the {@link VimeoAccount} across
-     * application sessions.
-     */
-    @Deprecated
-    void saveAccount(VimeoAccount vimeoAccount, String email, String password);
+    void saveAccount(@NotNull VimeoAccount vimeoAccount, String email);
 
-    void saveAccount(VimeoAccount vimeoAccount, String email);
-
-    void deleteAccount(VimeoAccount vimeoAccount);
+    void deleteAccount(@NotNull VimeoAccount vimeoAccount);
 }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -101,7 +101,7 @@ public class Configuration {
      * Persist the {@link VimeoAccount} in the {@link AccountStore}
      *
      * @param account     the account to persist
-     * @param accountName the name of the account. Required for authenticated user account,
+     * @param accountName the name of the account. Required for authenticated user accounts,
      *                    but may be null for client credential accounts
      */
     public void saveAccount(@NotNull VimeoAccount account, String accountName) {
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -114,7 +114,7 @@ public class Configuration {
 
     /**
      * Persist the {@link VimeoAccount} in the {@link AccountStore} without an account
-     * name. For authenticated users, user {@link #saveAccount(VimeoAccount, String)},
+     * name. For authenticated users, use {@link #saveAccount(VimeoAccount, String)},
      * as this is meant for client credential saving, when an account name is not available.
      *
      * @param account the account to persist
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -101,27 +101,27 @@ public class Configuration {
      * Persist the {@link VimeoAccount} in the {@link AccountStore}
      * <p>
      * Account name is <b>required</b> for authenticated user accounts; for
-     * client credential account saving use {@link #saveAccount(VimeoAccount)}
+     * client credential account saving use {@link #saveNonUserAccount(VimeoAccount)}
      *
      * @param account     the account to persist
      * @param accountName the name of the account.
      */
-    public void saveAccount(@NotNull VimeoAccount account, @NotNull String accountName) {
+    public void saveAuthenticatedUserAccount(@NotNull VimeoAccount account, @NotNull String accountName) {
         if (accountStore != null) {
-            accountStore.saveAccount(account, accountName);
+            accountStore.saveAuthenticatedUserAccount(account, accountName);
         }
     }
 
     /**
      * Persist the {@link VimeoAccount} in the {@link AccountStore} without an account
-     * name. For authenticated users, use {@link #saveAccount(VimeoAccount, String)},
-     * as this is meant for client credential saving, when an account name is not available.
+     * name. For authenticated users, use {@link #saveAuthenticatedUserAccount(VimeoAccount, String)},
+     * as this is meant for client credential or code grant saving, when an account name is not available.
      *
      * @param account the account to persist
      */
-    public void saveAccount(@NotNull VimeoAccount account) {
+    public void saveNonUserAccount(@NotNull VimeoAccount account) {
         if (accountStore != null) {
-            accountStore.saveAccount(account);
+            accountStore.saveNonUserAccount(account);
         }
     }
 
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -97,18 +97,35 @@ public class Configuration {
         return new Cache(cacheDirectory, cacheSize);
     }
 
-    public void saveAccount(@NotNull VimeoAccount account, String email) {
+    /**
+     * Persist the {@link VimeoAccount} in the {@link AccountStore}
+     *
+     * @param account     the account to persist
+     * @param accountName the name of the account. Required for authenticated user account,
+     *                    but may be null for client credential accounts
+     */
+    public void saveAccount(@NotNull VimeoAccount account, String accountName) {
         if (accountStore != null) {
-            accountStore.saveAccount(account, email);
+            accountStore.saveAccount(account, accountName);
         }
     }
 
+    /**
+     * Remove the {@link VimeoAccount} from the {@link AccountStore}
+     *
+     * @param account the account to remove
+     */
     public void deleteAccount(@NotNull VimeoAccount account) {
         if (accountStore != null) {
             accountStore.deleteAccount(account);
         }
     }
 
+    /**
+     * Load the {@link VimeoAccount} from the {@link AccountStore}
+     *
+     * @return the loaded account, or null if none could be loaded
+     */
     @Nullable
     public VimeoAccount loadAccount() {
         if (accountStore == null) {
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -26,6 +26,7 @@ import com.vimeo.networking.Vimeo.LogLevel;
 import com.vimeo.networking.logging.LogProvider;
 import com.vimeo.networking.model.VimeoAccount;
 
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.File;
@@ -96,24 +97,13 @@ public class Configuration {
         return new Cache(cacheDirectory, cacheSize);
     }
 
-    /**
-     * @deprecated use {@link #saveAccount(VimeoAccount, String)} instead
-     * <p/>
-     * We find no use in storing the password when you can persist the {@link VimeoAccount} across
-     * application sessions.
-     */
-    @Deprecated
-    public void saveAccount(VimeoAccount account, String email, String password) {
-        saveAccount(account, email);
-    }
-
-    public void saveAccount(VimeoAccount account, String email) {
+    public void saveAccount(@NotNull VimeoAccount account, String email) {
         if (accountStore != null) {
             accountStore.saveAccount(account, email);
         }
     }
 
-    public void deleteAccount(VimeoAccount account) {
+    public void deleteAccount(@NotNull VimeoAccount account) {
         if (accountStore != null) {
             accountStore.deleteAccount(account);
         }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -99,14 +99,29 @@ public class Configuration {
 
     /**
      * Persist the {@link VimeoAccount} in the {@link AccountStore}
+     * <p>
+     * Account name is <b>required</b> for authenticated user accounts; for
+     * client credential account saving use {@link #saveAccount(VimeoAccount)}
      *
      * @param account     the account to persist
-     * @param accountName the name of the account. Required for authenticated user accounts,
-     *                    but may be null for client credential accounts
+     * @param accountName the name of the account.
      */
-    public void saveAccount(@NotNull VimeoAccount account, String accountName) {
+    public void saveAccount(@NotNull VimeoAccount account, @NotNull String accountName) {
         if (accountStore != null) {
             accountStore.saveAccount(account, accountName);
+        }
+    }
+
+    /**
+     * Persist the {@link VimeoAccount} in the {@link AccountStore} without an account
+     * name. For authenticated users, user {@link #saveAccount(VimeoAccount, String)},
+     * as this is meant for client credential saving, when an account name is not available.
+     *
+     * @param account the account to persist
+     */
+    public void saveAccount(@NotNull VimeoAccount account) {
+        if (accountStore != null) {
+            accountStore.saveAccount(account);
         }
     }
 
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -126,7 +126,11 @@ public final class VimeoClient {
         ClientLogger.setLogLevel(mConfiguration.logLevel);
 
         VimeoAccount vimeoAccount = mConfiguration.loadAccount();
-        setVimeoAccount(vimeoAccount);
+        if (vimeoAccount == null) {
+            configureAccessTokenAccount();
+        } else {
+            setVimeoAccount(vimeoAccount);
+        }
     }
 
     private Retrofit createRetrofit() {
@@ -216,52 +220,55 @@ public final class VimeoClient {
     }
 
     /**
-     * Sets the current {@link VimeoAccount} on this client. If the provided account is null and
-     * an access token exists on the {@link Configuration}, then a new account will be created
-     * using that access token.
+     * Sets the current {@link VimeoAccount} on this client.
      *
      * @param vimeoAccount the account to set on this client
      */
     public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
-        if (vimeoAccount == null) {
-            if (mConfiguration.accessToken != null) {
-                // If the provided account was null but we have an access token, persist the vimeo account with
-                // just a token in it. Otherwise we'll want to leave the persisted account as null.
-                vimeoAccount = new VimeoAccount(mConfiguration.accessToken);
-                mConfiguration.saveNonUserAccount(vimeoAccount);
-            }
-        }
-
         mVimeoAccount = vimeoAccount;
+    }
+
+    /**
+     * This will configure a {@link VimeoAccount} using the access token provided to the
+     * {@link Configuration}.
+     */
+    private void configureAccessTokenAccount() {
+        if (mConfiguration.accessToken != null) {
+            // If the provided account was null but we have an access token, persist the vimeo account with
+            // just a token in it. Otherwise we'll want to leave the persisted account as null.
+            VimeoAccount vimeoAccount = new VimeoAccount(mConfiguration.accessToken);
+            mConfiguration.saveNonUserAccount(vimeoAccount);
+
+            setVimeoAccount(vimeoAccount);
+        }
     }
 
     /**
      * Sets the {@link #mVimeoAccount} field as well as triggering the saveAuthenticatedUserAccount
      * event for the account store.
      *
-     * @param vimeoAccount The account to save - this should be provided; if it is not, then one will
-     *                     be constructed using the access token set on the {@link Configuration}.
+     * @param vimeoAccount The account to save
      * @param accountName  accountName should be provided if the account is for an authenticated user. It may
      *                     be null if the account represents a client credentials grant.
      */
-    public void saveAuthenticatedUserAccount(@Nullable VimeoAccount vimeoAccount,
+    public void saveAuthenticatedUserAccount(@NotNull VimeoAccount vimeoAccount,
                                              @NotNull String accountName) {
-        setVimeoAccount(vimeoAccount);
         if (vimeoAccount != null) {
+            setVimeoAccount(vimeoAccount);
             mConfiguration.saveAuthenticatedUserAccount(vimeoAccount, accountName);
         }
     }
 
     /**
      * Sets the {@link #mVimeoAccount} field as well as triggering the saveNonUserAccount event for the
-     * account store. This is meant to be user when no user is directly associated with the account.
+     * account store. This is meant to be used when no user is directly associated with the account.
      *
      * @param vimeoAccount The account to save - this should be provided; if it is not, then one will
      *                     be constructed using the access token set on the {@link Configuration}.
      */
     public void saveNonUserAccount(@Nullable VimeoAccount vimeoAccount) {
-        setVimeoAccount(vimeoAccount);
         if (vimeoAccount != null) {
+            setVimeoAccount(vimeoAccount);
             mConfiguration.saveNonUserAccount(vimeoAccount);
         }
     }
@@ -602,7 +609,9 @@ public final class VimeoClient {
 
         VimeoAccount vimeoAccount = executeAccountCall(call);
 
-        saveAuthenticatedUserAccount(vimeoAccount, email);
+        if (vimeoAccount != null) {
+            saveAuthenticatedUserAccount(vimeoAccount, email);
+        }
 
         return vimeoAccount;
     }
@@ -702,7 +711,7 @@ public final class VimeoClient {
     private static class AccountCallback extends VimeoCallback<VimeoAccount> {
 
         private final VimeoClient mClient;
-        private String mEmail;
+        private String mAccountName;
         private final AuthCallback mCallback;
 
         public AccountCallback(VimeoClient client, AuthCallback callback) {
@@ -714,13 +723,13 @@ public final class VimeoClient {
             mCallback = callback;
         }
 
-        public AccountCallback(VimeoClient client, String email, AuthCallback callback) {
+        public AccountCallback(VimeoClient client, String accountName, AuthCallback callback) {
             if (client == null || callback == null) {
                 throw new AssertionError("Client and Callback must not be null");
             }
 
             mClient = client;
-            mEmail = email;
+            mAccountName = accountName;
             mCallback = callback;
         }
 
@@ -737,14 +746,14 @@ public final class VimeoClient {
 
         @Override
         public void success(VimeoAccount vimeoAccount) {
-            if (vimeoAccount.getUser() != null && (mEmail == null || mEmail.isEmpty())) {
+            if (vimeoAccount.getUser() != null && (mAccountName == null || mAccountName.isEmpty())) {
                 // We must always have a `name` field, which is used by the Android Account Manager for
                 // display in the device Settings -> Accounts [KZ] 12/17/15
                 String name = (vimeoAccount.getUser().name !=
                                null) ? vimeoAccount.getUser().name : vimeoAccount.getUser().uri;
                 mClient.saveAuthenticatedUserAccount(vimeoAccount, name);
             } else {
-                mClient.saveAuthenticatedUserAccount(vimeoAccount, mEmail);
+                mClient.saveAuthenticatedUserAccount(vimeoAccount, mAccountName);
             }
             mCallback.success();
         }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -445,8 +445,6 @@ public final class VimeoClient {
 
         VimeoAccount vimeoAccount = executeAccountCall(call);
 
-        saveAccount(vimeoAccount, accountName);
-
         return vimeoAccount;
     }
 
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -223,7 +223,12 @@ public final class VimeoClient {
 
     /**
      * Sets the {@link #mVimeoAccount} field as well as triggering the saveAccount event for the
-     * account store
+     * account store.
+     *
+     * @param vimeoAccount The account to save - this should be provided; if it is not, then one will
+     *                     be constructed using the access token set on the {@link Configuration}.
+     * @param email        Email should be provided if the account is for an authenticated user. It may
+     *                     be null if the account represents a client credentials grant.
      */
     public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email) {
         setVimeoAccount(vimeoAccount);
@@ -417,7 +422,6 @@ public final class VimeoClient {
     /**
      * Synchronous version of {@link #singleSignOnTokenExchange(String, AuthCallback)}. This is useful
      * when already running in a background thread, such as when using the Android AccountManager.
-     * See the documentation of {@link #singleSignOnTokenExchange(String, AuthCallback)} for more info.
      *
      * @param token           the authenticated user access token from application A. This <b>cannot</b> be a client
      *                        credentials access token.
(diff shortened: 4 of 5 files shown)


howardrigberg:
I will get to this tomorrow morning PST. I have another PR ahead of it for Anthony.

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
```diff
@@ -24,25 +24,20 @@
...
-     * @deprecated use {@link #saveAccount(VimeoAccount, String)} instead
-     * <p/>
-     * We find no use in storing the password when you can persist the {@link VimeoAccount} across
-     * application sessions.
-     */
-    @Deprecated
-    void saveAccount(VimeoAccount vimeoAccount, String email, String password);
-
-    void saveAccount(VimeoAccount vimeoAccount, String email);
+    void saveAccount(@NotNull VimeoAccount vimeoAccount, String email);
```
howardrigberg: Annotate email with @NotNull too?

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -96,24 +97,13 @@ public Cache getCache() {
...
-     * We find no use in storing the password when you can persist the {@link VimeoAccount} across
-     * application sessions.
-     */
-    @Deprecated
-    public void saveAccount(VimeoAccount account, String email, String password) {
-        saveAccount(account, email);
-    }
-
-    public void saveAccount(VimeoAccount account, String email) {
+    public void saveAccount(@NotNull VimeoAccount account, String email) {
```
howardrigberg: @NotNull for email?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -221,23 +222,14 @@ public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
...
-    @Deprecated
-    public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email, String password) {
-        saveAccount(vimeoAccount, email);
-    }
-
-    /**
      * Sets the {@link #mVimeoAccount} field as well as triggering the saveAccount event for the
      * account store
      */
     public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email) {
```
howardrigberg: Again, why ignore email in the annotations? Also, it seems a little strange to allow a null account to be passed in here. It's not really sane behavior for a caller to try to save a null account. I get that you are protecting against null below. It's also inconsistent with the rest of the annotations that you've got for saveAccount in other places.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -221,23 +222,14 @@ public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
...
         setVimeoAccount(vimeoAccount);
-        mConfiguration.saveAccount(vimeoAccount, email);
+        if (vimeoAccount != null) {
+            mConfiguration.saveAccount(vimeoAccount, email);
```
howardrigberg: Is it OK to save with a null email?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -422,6 +414,37 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
         return call;
     }
 
+    /**
+     * Synchronous version of {@link #singleSignOnTokenExchange(String, AuthCallback)}. This is useful
+     * when already running in a background thread, such as when using the Android AccountManager.
+     * See the documentation of {@link #singleSignOnTokenExchange(String, AuthCallback)} for more info.
```
howardrigberg: Use @see?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -422,6 +414,37 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
...
+
+        if (basicAuthHeader == null) {
+            basicAuthHeader = getBasicAuthHeader();
+        }
+        Call<VimeoAccount> call =
+                mVimeoService.ssoTokenExchange(basicAuthHeader, token, mConfiguration.scope);
+
+        VimeoAccount vimeoAccount = executeAccountCall(call);
+
+        saveAccount(vimeoAccount, accountName);
```
howardrigberg: It seems a little confusing that we have named the parameter email in some places and name in others. Does it have to be an email? If not, maybe we should rename all of the parameters as accountName instead.

Also, why does this method save the account too? Shouldn't that be left to the caller? If it remains here, maybe the null check of the client account should be at this level?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,53 +40,65 @@
...
     public User user;
     private String userJSON;
 
-    public VimeoAccount(){
+    public VimeoAccount() {
         //constructor for stag TypeAdapter generation
     }
 
     public VimeoAccount(@Nullable String accessToken) {
         this.accessToken = accessToken;
```
howardrigberg: Is there a reason why this class uses this. for its member fields? It's not our general coding style elsewhere. Ditto for the "m" prefix for members.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,53 +40,65 @@
...
     }
 
-    public VimeoAccount(String accessToken, String tokenType, String scope, String userJSON) {
-        if (accessToken == null || accessToken.isEmpty() || tokenType == null ||
-            tokenType.isEmpty() || scope == null || scope.isEmpty()) {
+    public VimeoAccount(@Nullable String accessToken, @NotNull String tokenType, @NotNull String scope,
+                        String userJSON) {
+        if ((accessToken != null && accessToken.isEmpty()) || tokenType == null || tokenType.isEmpty() ||
```
howardrigberg: It seems odd to have the annotation that the access token is nullable but then throw an exception if it is null. 

Also, since you are doing the null plus isEmpty check so many times, it would be good to break this out into a helper somewhere.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,53 +40,65 @@
...
+            scope == null || scope.isEmpty()) {
             throw new AssertionError("Account can only be created with token, tokenType, scope");
         }
 
         this.accessToken = accessToken;
         this.tokenType = tokenType;
         this.scope = scope;
 
-        Gson gson = VimeoNetworkUtil.getGson();
+        if (userJSON != null) {
```
howardrigberg: Check against empty? Also do we want to do any other checks or are we cool with a RuntimeException being thrown for invalid JSON here?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -95,13 +108,12 @@ public User getUser() {
...
         this.user = user;
     }
 
     @Nullable
-    public String getUserJSON() // For AccountManager.userData [AH]
-    {
+    public String getUserJSON() {
         if (this.user == null) {
             return null;
         }
```
howardrigberg: The logic for this method seems flawed. I realize that it isn't part of this PR, but if the user ever changes on this account object between calls to this method it will be returning an incorrectly cached userJSON string.

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
kevinzetterstrom: Email can be null - the case for this is a client credentials "account" in which we have scope, token type, and an auth token (a VimeoAccount) but nothing that uniquely identifies the user. The email field here is the Android Account ```name``` field that will appear in the AccountManager - I can change it to accountName for clarity.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: 👍

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: This is following all of the other authentication methods - as soon as we get a VimeoAccount, let's save it for use right away.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
kevinzetterstrom: It's not throwing an exception if ```accessToken``` is null - only if its empty. There will now be a possibility that our account store couldn't access the access token in a synchronous manner, so when instantiating a VimeoAccount, we want access token to be allowed to be null. However, if it isn't null, then it should have characters - an empty access token would always be invalid.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: I ignored email because it should be passed in when there is an authenticated user token, but can be null when there is a client credentials grant. I will add javadoc to the matter, since I think an annotation for Nullable is misleading.

Also, we allow a null account to be passed in since the methods fetching the account could return null, and this method will correctly construct one if so.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kevinzetterstrom: May be null if this doesn't represent an authenticated user. I will add javadoc.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
kevinzetterstrom: This (pun intended) was the original code written by Alfie way back when. We need the qualifier since the parameter in the constructor and the member are the same name. It has been our practice to only add "m" to new classes (until we completely refactor the library) since this is a public library and we would need to deprecate everything properly. Since this is going in to 2.0.0 and there will be breaking changes, I could do it here. I suppose I kept with our standard here rather than changing it throughout the code base, but it's a good call out.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
kevinzetterstrom: This could only be validS JSON since we are setting it ourselves when we set the user (initially), using Gson - ```userJSON``` is not a public member. I will make sure it gets set when we set the user too.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
kevinzetterstrom: 👍 nice call out - it wasn't something I was paying attention to, but I will make sure it gets done right as part of this PR.

kevinzetterstrom:
@howardrigberg changes here 7cda454 and here 2f177d6

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: Ah man, I did have to remove saveAccount here, due to the nature in which this method is called.

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
```diff
@@ -24,25 +24,37 @@
...
-     * application sessions.
+     * Saves a {@link VimeoAccount} for loading later.
+     *
+     * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
+     * @param accountName  the <i>name</i> of the account that helps uniquely identify it in account
+     *                     stores, such as the Android AccountManager
      */
-    @Deprecated
-    void saveAccount(VimeoAccount vimeoAccount, String email, String password);
+    void saveAccount(@NotNull VimeoAccount vimeoAccount, String accountName);
```
howardrigberg: Sorry to keep at this with the annotations, but I am following up on it since we were discussing this in the context of another PR (https://github.vimeows.com/MobileApps/Vimeo-Android/pull/912). We are going to try to move toward explicitly annotating null and nonnull parameters in method signatures. @alfiehanssen asked us to do it in that context. We've got it on the agenda to discuss for a future Android sync, but in the mean time it makes sense to apply the practice to this PR. It'd be great if you could quickly sweep the files that you touched for this PR and add the annotations. Since there will already be breaking changes in this version. It would be good to review the networking interfaces in general w/r/t annotations and make sure it's all explicit and correct. If it is too much to take on here, we should ticket it, but be sure it gets done before 2.0.0 goes out.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -221,23 +222,19 @@ public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
...
-     * account store
+     * account store.
+     *
+     * @param vimeoAccount The account to save - this should be provided; if it is not, then one will
+     *                     be constructed using the access token set on the {@link Configuration}.
+     * @param accountName  accountName should be provided if the account is for an authenticated user. It may
+     *                     be null if the account represents a client credentials grant.
      */
-    public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email) {
+    public void saveAccount(@Nullable VimeoAccount vimeoAccount, String accountName) {
```
howardrigberg: See above comment regarding annotations.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -422,6 +419,35 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
...
+     * Synchronous version of {@link #singleSignOnTokenExchange(String, AuthCallback)}. This is useful
+     * when already running in a background thread, such as when using the Android AccountManager.
+     *
+     * @param token           the authenticated user access token from application A. This <b>cannot</b> be a client
+     *                        credentials access token.
+     * @param accountName     the name of the account, usually the email
+     * @param basicAuthHeader a "basic" auth header to pass in; the basic auth header should be from
+     *                        the application that needs the new token. In the case of the Android
+     *                        account authenticator, which runs on the first app installed, we need
+     *                        to pass in that token
```
howardrigberg: Maybe also mention what happens if you pass null for basicAuthHeader

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -575,6 +593,19 @@ public VimeoAccount logIn(String email, String password) {
         return call;
     }
 
+    @Nullable
+    private static VimeoAccount executeAccountCall(@NotNull Call<VimeoAccount> call) {
+        try {
+            retrofit2.Response<VimeoAccount> response = call.execute();
+            if (response.isSuccessful()) {
+                return response.body();
+            }
```
howardrigberg: Is there anything reasonable to log if the call is not successful?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,81 +40,141 @@
...
+        mAccessToken = accessToken;
+        mTokenType = tokenType;
+        mScope = scope;
 
-        Gson gson = VimeoNetworkUtil.getGson();
+        if (userJSON != null) {
+            Gson gson = VimeoNetworkUtil.getGson();
 
-        this.user = gson.fromJson(userJSON, User.class);
+            mUser = gson.fromJson(userJSON, User.class);
```
howardrigberg: Shouldn't mUserJson also be saved off here?

kevinzetterstrom: Good catch - it will help prevent us from trying to create it later

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: I can add a log in.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: 👍

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
kevinzetterstrom: Again, the reason that I didn't add it is this: if the VimeoAccount represents a client credentials grant, then the accountName should be null, and we would want to show this using the Nullable annotation. 
If the VimeoAccount represents an authenticated user, than the accountName should not be null - but we shouldn't use the NotNull annotation here because of the fact that it would be null in the aforementioned case. 

I think adding either annotation would be misleading - many times a user looks at the signature of the method to determine what is required to be passed in. I think leaving it off in a scenario here is helpful - as a consumer of the method I would try and look at the javadoc or the method more closely to determine if it is needed and in what scenarios then.

I'd like to leave annotations off of method parameters like this until we discuss in Android sync the preferred approach. And since all of these files will be refactored before a 2.0.0 release, we have more than plenty of time to drop the annotations in (or not, depending on the outcome of our discussion).

kevinzetterstrom:
@howardrigberg  latest changes 8b93c16

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
howardrigberg: Perhaps the problem is better solved by providing separate interface methods with clear requirements for the parameters. Why not have a separate method to save client credentials grant and another one for authenticated users. The API would be much less ambiguous and you could annotate everything.

alfiehanssen: I think the use cases you described @kevinzetterstrom mean that the `accountName` argument is firmly in the Nullable zone. 

I agree with @howardrigberg that we should use both annotations always. If an argument is not one then it's the other. Leaving arguments blank is misleading. 

Will you please do a sweep and add for this PR?

alfiehanssen: I also think that Howard's comment about splitting the method into two makes sense as well, in order to address the separate issue/question about the various use cases for which this method will be called.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -97,28 +98,34 @@ public Cache getCache() {
...
+     * @param accountName the name of the account. Required for authenticated user accounts,
+     *                    but may be null for client credential accounts
      */
-    @Deprecated
-    public void saveAccount(VimeoAccount account, String email, String password) {
-        saveAccount(account, email);
-    }
-
-    public void saveAccount(VimeoAccount account, String email) {
+    public void saveAccount(@NotNull VimeoAccount account, String accountName) {
```
alfiehanssen: Please add nullability annotation

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
kevinzetterstrom: @howardrigberg peep the latest for those overloaded changes

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -422,6 +419,36 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
         return call;
     }
 
+    /**
+     * Synchronous version of {@link #singleSignOnTokenExchange(String, AuthCallback)}. This is useful
+     * when already running in a background thread, such as when using the Android AccountManager.
+     *
+     * @param token           the authenticated user access token from application A. This <b>cannot</b> be a client
```
alfiehanssen: Does this mean that the token must be a password-grant token? If so, consider saying so in the docs (rather than saying what it can't be), and maybe even capturing this in the method name or argument name.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -422,6 +419,36 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
         return call;
     }
 
+    /**
+     * Synchronous version of {@link #singleSignOnTokenExchange(String, AuthCallback)}. This is useful
+     * when already running in a background thread, such as when using the Android AccountManager.
```
alfiehanssen: Can you elaborate in these docs on the use case for this method? Can't hurt to get specific.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -422,6 +419,36 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
...
+     *
+     * @param token           the authenticated user access token from application A. This <b>cannot</b> be a client
+     *                        credentials access token.
+     * @param accountName     the name of the account, usually the email
+     * @param basicAuthHeader a "basic" auth header to pass in; the basic auth header should be from
```
alfiehanssen: I'm unclear as to why we need to pass in a basic auth header. Can you describe how this is used and where it comes from?

alfiehanssen: It seems strange that we wouldn't just use `getBasicAuthHeader()`

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
alfiehanssen: @kevinzetterstrom Can you elaborate on why a null accessToken is valid? It seems dangerous to create an account without one. 

If the access token is null, what do we do?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,81 +40,142 @@
...
+
+    @Nullable
+    private String mUserJson;
 
-    public VimeoAccount(){
+    public VimeoAccount() {
         //constructor for stag TypeAdapter generation
     }
 
     public VimeoAccount(@Nullable String accessToken) {
```
alfiehanssen: Why is this argument nullable? What is an account if it doesn't have an accessToken?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,81 +40,142 @@
...
+            mUserJson = userJSON;
+        }
     }
 
+    /**
+     * @return true if the access token is not empty, false if it is null or empty
+     */
     public boolean isAuthenticated() {
-        return (this.accessToken != null && !this.accessToken.isEmpty());
+        return (mAccessToken != null && !mAccessToken.isEmpty());
```
alfiehanssen: Similar to questions above. If the account doesn't have a token then I question whether it's actually an account. I.e. maybe that state should be represented as a null account rather than a nonnull account with a null token. 

And in that case, an empty token would be an invalid state.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,81 +40,142 @@
...
     }
 
+    /**
+     * Sets the auth token. This should only be used when there is no auth token set, or it otherwise
+     * needs to change. If changed, this {@link VimeoAccount} should be set and saved in the client
+     * for use.
+     *
+     * @param accessToken the new auth token
+     */
+    public void setAccessToken(@Nullable String accessToken) {
```
alfiehanssen: Why is this method necessary? We have a constructor that takes an auth token. I'm wary or making this externally settable as well as allowing it to be null.

Related to my comments above

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kevinzetterstrom: @alfiehanssen I added them in my latest commit

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: I thought the docs were details enough, but I added a little more specificity here: 99d20cf

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: It just means that it must be an authenticated user - user is the key word here. Adding more specificity here 99d20cf

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
kevinzetterstrom: All our fields should be nullable in our networking models - we shouldn't have any of them guaranteed. This is why the Nullable annotation is added on the setters and constructor like this.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
alfiehanssen: @kevinzetterstrom Trying to understand the ins and outs of this. How is it possible for an account stored in the account manager to not have an auth token?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: FWIW, this method is not consumable for general public use, and I have that stated in the javadocs.

kevinzetterstrom:
@howardrigberg @alfiehanssen I have made it so that a ```VimeoAccount``` must be created with an auth token, and have allowed the VimeoAccount member of VimeoClient to be null, which will be the case until we have a token.

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
```diff
@@ -38,19 +39,18 @@ public VimeoAccount loadAccount() {
...
     @Override
-    public void saveAccount(VimeoAccount vimeoAccount, String email) {
+    public void saveAccount(@NonNull VimeoAccount vimeoAccount, @NonNull String email) {
         AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
     @Override
-    public void deleteAccount(VimeoAccount vimeoAccount) {
+    public void deleteAccount(@NonNull VimeoAccount vimeoAccount) {
         AccountPreferenceManager.removeClientAccount();
```
howardrigberg: Should the code be checking something about the identity of the VimeoAccount before deleting it?

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -97,28 +98,49 @@ public Cache getCache() {
...
+     * Account name is <b>required</b> for authenticated user accounts; for
+     * client credential account saving use {@link #saveAccount(VimeoAccount)}
+     *
+     * @param account     the account to persist
+     * @param accountName the name of the account.
      */
-    @Deprecated
-    public void saveAccount(VimeoAccount account, String email, String password) {
-        saveAccount(account, email);
+    public void saveAccount(@NotNull VimeoAccount account, @NotNull String accountName) {
```
howardrigberg: Is there a reason not to name these saveAuthenticatedUserAccount and saveClientCredentialsAccount?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -199,45 +200,49 @@ public Retrofit getRetrofit() {
...
+    /**
+     * @return the current VimeoAccount, or null if it hasn't been set yet
+     */
+    @Nullable
     public VimeoAccount getVimeoAccount() {
-        if (mVimeoAccount == null) {
-            throw new AssertionError("Account should never be null");
-        }
-
         return mVimeoAccount;
```
howardrigberg: It seems potentially dangerous to return the reference to the actual private member. Is this a safe move or should it be cloned?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -199,45 +200,49 @@ public Retrofit getRetrofit() {
...
     }
 
+    public boolean isAuthenticated() {
```
howardrigberg: Javadoc?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -199,45 +200,49 @@ public Retrofit getRetrofit() {
...
+        return mVimeoAccount != null && mVimeoAccount.isAuthenticated();
+    }
+
     public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
```
howardrigberg: Javadoc this method too.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -199,45 +200,49 @@ public Retrofit getRetrofit() {
...
-     * account store
+     * account store.
+     *
+     * @param vimeoAccount The account to save - this should be provided; if it is not, then one will
+     *                     be constructed using the access token set on the {@link Configuration}.
+     * @param accountName  accountName should be provided if the account is for an authenticated user. It may
+     *                     be null if the account represents a client credentials grant.
      */
-    public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email) {
+    public void saveAccount(@Nullable VimeoAccount vimeoAccount, @Nullable String accountName) {
```
howardrigberg: Maybe it makes sense to split this into two versions like it is in Configuration?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,81 +40,156 @@
 public class VimeoAccount implements Serializable {
 
     private static final long serialVersionUID = -8341071767843490585L;
-    //    private static final String TOKEN_TYPE_BEARER = "bearer";
 
+    @Nullable
     @GsonAdapterKey("access_token")
-    public String accessToken;
+    public String mAccessToken;
```
howardrigberg: private members with accessors?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
```diff
@@ -39,81 +40,156 @@
...
+        }
+
+        mAccessToken = accessToken;
     }
 
-    public VimeoAccount(String accessToken, String tokenType, String scope, String userJSON) {
-        if (accessToken == null || accessToken.isEmpty() || tokenType == null ||
-            tokenType.isEmpty() || scope == null || scope.isEmpty()) {
+    /**
+     * Using this constructor is to create an account manually. It is recommended that if you cannot obtain
```
howardrigberg: Maybe it is better worded as: "Use this constructor to create an account manually. If you cannot obtain a valid access token, it is recommended to have a VimeoAccount constructed automatically using one of the authentication methods in VimeoClient rather than calling this constructor directly." If I am misunderstanding the docs, please clarify.

howardrigberg:
OK. I've reviewed this again. A few more suggestions added. Maybe @alfiehanssen should also take another look since a lot of the initial comments were his.

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
kevinzetterstrom: No dice on this until we change Stag so that it supports non-public members

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: 👍

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: I don't believe this is dangerous to return the reference here, but we could clone it to be safe.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kevinzetterstrom: No reason, this is just how they were previously named - it is a good suggestion and I can make this change.

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
```diff
@@ -38,19 +39,18 @@ public VimeoAccount loadAccount() {
...
     @Override
-    public void saveAccount(VimeoAccount vimeoAccount, String email, String password) {
-        saveAccount(vimeoAccount, email);
+    public void saveAccount(@NonNull VimeoAccount vimeoAccount) {
+        AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
     @Override
-    public void saveAccount(VimeoAccount vimeoAccount, String email) {
+    public void saveAccount(@NonNull VimeoAccount vimeoAccount, @NonNull String email) {
```
kevinzetterstrom: I think if we were to use the second parameter then we would need a lot more demo code than we have; I think the demo could be improved upon greatly, but I also don't think this is the time to do it. I will make sure the name matches though.

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
```diff
@@ -38,19 +39,18 @@ public VimeoAccount loadAccount() {
         return AccountPreferenceManager.getClientAccount();
```
kevinzetterstrom: I can add those annotation.

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
kevinzetterstrom: I don't think we need that, especially since this is a very simplistic demo. If we were to have a more in depth demo that perhaps. Something to keep in mind for 2.0.0

kevinzetterstrom:
@howardrigberg great suggestions, peep the latest commit

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
```diff
@@ -34,18 +35,20 @@ public TestAccountStore(Context context) {
...
     @Override
-    public void saveAccount(@NonNull VimeoAccount vimeoAccount) {
+    public void saveNonUserAccount(@NonNull VimeoAccount vimeoAccount) {
         AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
     @Override
-    public void saveAccount(@NonNull VimeoAccount vimeoAccount, @NonNull String email) {
+    public void saveAuthenticatedUserAccount(@NonNull VimeoAccount vimeoAccount,
+                                             @NonNull String accountName) {
```
howardrigberg: It would help to just add a line comment to this method to indicate that in non-demo code, you'd probably want to store the account name as well.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -101,27 +101,27 @@ public Cache getCache() {
...
      * Account name is <b>required</b> for authenticated user accounts; for
-     * client credential account saving use {@link #saveAccount(VimeoAccount)}
+     * client credential account saving use {@link #saveNonUserAccount(VimeoAccount)}
      *
      * @param account     the account to persist
      * @param accountName the name of the account.
      */
-    public void saveAccount(@NotNull VimeoAccount account, @NotNull String accountName) {
+    public void saveAuthenticatedUserAccount(@NotNull VimeoAccount account, @NotNull String accountName) {
         if (accountStore != null) {
```
howardrigberg: Hey, I'm sorry to bring this one up, Kevin... I feel like we've opened a whole can of worms with this PR as we touch different files. This one also still has the old member variable naming (i.e. accountStore vs mAccountStore). If it is too much of a pain in the ass, we can just ticket the rest of the renaming of the members of networking.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -205,46 +205,68 @@ public Retrofit getRetrofit() {
...
-            } else {
-                mConfiguration.saveAccount(vimeoAccount);
-            }
+            mConfiguration.saveAuthenticatedUserAccount(vimeoAccount, accountName);
+        }
+    }
+
+    /**
+     * Sets the {@link #mVimeoAccount} field as well as triggering the saveNonUserAccount event for the
+     * account store. This is meant to be user when no user is directly associated with the account.
```
howardrigberg: "This is meant to be used..."

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
```diff
@@ -33,24 +35,25 @@ public TestAccountStore(Context context) {
...
     @Override
     public VimeoAccount loadAccount() {
         return AccountPreferenceManager.getClientAccount();
     }
 
-    @Deprecated
     @Override
-    public void saveAccount(VimeoAccount vimeoAccount, String email, String password) {
-        saveAccount(vimeoAccount, email);
+    public void saveNonUserAccount(@NonNull VimeoAccount vimeoAccount) {
```
alfiehanssen: Call it `saveClientCredentialsAccount`? This seems to be more specific. Unless I'm missing another form of "non-user" account.

alfiehanssen: Similar question to @howardrigberg, should this method be checking to see if this is actually a non-user account?

howardrigberg:
Kevin, I have finished reviewing the latest commit. My suggestions are all minor and I don't feel the need to review this again. However, please get @alfiehanssen to sign off on it before committing since many of the comments were his.

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
alfiehanssen: @howardrigberg Can you elaborate?

If it's a simple change (seems like it is), it's something we should probably do, no?

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
kevinzetterstrom: @alfiehanssen a user can generate an auth token via the web API console, and use with the networking library. I was under the assumption this is not a client credentials grant. Do you have a suggestion for a name that accounts (pun intended) for a client credentials grant and that auth type?

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
```diff
@@ -24,25 +24,48 @@
...
+     * <p>
+     * Use this method for saving authenticated users that have an account name.
+     *
+     * @param vimeoAccount the VimeoAccount to save, should not be null when calling this method
+     * @param accountName  the <i>name</i> of the account that helps uniquely identify it in account
+     *                     stores, such as the Android AccountManager
      */
-    @Deprecated
-    void saveAccount(VimeoAccount vimeoAccount, String email, String password);
+    void saveAuthenticatedUserAccount(@NotNull VimeoAccount vimeoAccount, @NotNull String accountName);
```
alfiehanssen: Nit: In the test class above the methods are listed in one order and they're listed in a different order in this interface. Use consistent ordering?

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -97,28 +98,49 @@ public Cache getCache() {
...
+     *
+     * @param account     the account to persist
+     * @param accountName the name of the account.
      */
-    @Deprecated
-    public void saveAccount(VimeoAccount account, String email, String password) {
-        saveAccount(account, email);
+    public void saveAuthenticatedUserAccount(@NotNull VimeoAccount account, @NotNull String accountName) {
+        if (accountStore != null) {
+            accountStore.saveAuthenticatedUserAccount(account, accountName);
```
alfiehanssen: Do the nullability annotations just generate warnings but wont prevent compile or run-time events? In other words, do we actually need null checks everywhere we have a nonnull argument?

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -97,28 +98,49 @@ public Cache getCache() {
...
+     * Persist the {@link VimeoAccount} in the {@link AccountStore} without an account
+     * name. For authenticated users, use {@link #saveAuthenticatedUserAccount(VimeoAccount, String)},
+     * as this is meant for client credential or code grant saving, when an account name is not available.
+     *
+     * @param account the account to persist
+     */
+    public void saveNonUserAccount(@NotNull VimeoAccount account) {
         if (accountStore != null) {
-            accountStore.saveAccount(account, email);
+            accountStore.saveNonUserAccount(account);
```
alfiehanssen: Why do we make a distinction between saving a user vs. client credentials account? What is the advantage of doing it this way?

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kevinzetterstrom: I think the issue with changing the member variable names is the spider web effect - changing the name here will have implications on another class, which in turn may need name changes, and it can continue to branch out from there. 
How we handle this is something we should touch on in the next Android sync - do we ticket every model and their app - dependencies individually? Do we tackle each class when working on tickets related to the class?
I am wary of touching more variable renaming because of this spider web effect. There is a ticket to work on the Configuration/Config Builder already (https://vimean.atlassian.net/browse/VA-1473) so if anything, I would think that ticket would be more appropriate than this one to do that clean up.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -199,47 +200,73 @@ public Retrofit getRetrofit() {
...
 
+    /**
+     * Sets the current {@link VimeoAccount} on this client. If the provided account is null and
+     * an access token exists on the {@link Configuration}, then a new account will be created
+     * using that access token.
+     *
+     * @param vimeoAccount the account to set on this client
+     */
     public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
         if (vimeoAccount == null) {
```
alfiehanssen: Otherwise we're mixing concerns and it's not transparent to anyone outside of this method.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -199,47 +200,73 @@ public Retrofit getRetrofit() {
...
+     *                     be constructed using the access token set on the {@link Configuration}.
+     * @param accountName  accountName should be provided if the account is for an authenticated user. It may
+     *                     be null if the account represents a client credentials grant.
      */
-    @Deprecated
-    public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email, String password) {
-        saveAccount(vimeoAccount, email);
+    public void saveAuthenticatedUserAccount(@Nullable VimeoAccount vimeoAccount,
+                                             @NotNull String accountName) {
+        setVimeoAccount(vimeoAccount);
```
alfiehanssen: Similar comment here, it seems strange to be about to "save an authenticated user account that is null". What does saving a null account mean? Can this API be clarified so the method name(s) and implementation(s) are crystal clear?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -199,47 +200,73 @@ public Retrofit getRetrofit() {
...
-     * account store
+     * Sets the {@link #mVimeoAccount} field as well as triggering the saveNonUserAccount event for the
+     * account store. This is meant to be user when no user is directly associated with the account.
+     *
+     * @param vimeoAccount The account to save - this should be provided; if it is not, then one will
+     *                     be constructed using the access token set on the {@link Configuration}.
      */
-    public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email) {
+    public void saveNonUserAccount(@Nullable VimeoAccount vimeoAccount) {
         setVimeoAccount(vimeoAccount);
```
alfiehanssen: Seems like we're relying on similar logic here. In this case with `setVimeoAccount` doing some opaque stuff in response to a null account object.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -535,17 +600,9 @@ public VimeoAccount logIn(String email, String password) {
...
-            if (response.isSuccessful()) {
-                vimeoAccount = response.body();
-            }
-        } catch (IOException e) {
-            ClientLogger.e("Exception during logIn: " + e.getMessage(), e);
-        }
+        VimeoAccount vimeoAccount = executeAccountCall(call);
 
-        saveAccount(vimeoAccount, email);
+        saveAuthenticatedUserAccount(vimeoAccount, email);
```
alfiehanssen: Seems like we call this `email` here but something like username or account name above. Can we be consistent about how we refer to this piece of data?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
alfiehanssen: @kevinzetterstrom Is there a ticket or other doc that captures the need for this?

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kevinzetterstrom: @alfiehanssen this was the argument brought up in the initial review of this PR - that it would be clearer if we had two separate methods for saving an account rather that one with optional parameters. Both you and @howardrigberg thought that was the preferred approach. It was also suggested I rename "saveAccount" so that it is more pointed as to what is being saved. Are you saying that you are against this approach now?

vimeo-networking/src/main/java/com/vimeo/networking/model/VimeoAccount.java
kevinzetterstrom: Yes, it was recognized when stag was released and is listed in the future enhancements section of its readme. Also, it is ticketed for us to take on here: https://vimean.atlassian.net/browse/VA-1926

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: I can change it to account name here - this stemmed from us using this callback inside login/join methods, which require an email for our API.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kevinzetterstrom: _that_ is the big issue @howardrigberg and I have with the NonNull/NotNull annotations - it only generates warnings. It's on the board for us to discuss at the next Android sync.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: 👍

kevinzetterstrom: I can clean this up for authenticating a config access token, but I think we should be able to null out the account somehow. Do you propose a different way of clearing out the account rather than allowing null to be passed in?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: I cleaned up the opaqueness by having setVimeoAccount only be a setter.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: 👍

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
kevinzetterstrom: 👍

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kevinzetterstrom: 👍

example/src/main/java/com/vimeo/android/networking/example/vimeonetworking/TestAccountStore.java
```diff
@@ -33,24 +35,25 @@ public TestAccountStore(Context context) {
...
+        AccountPreferenceManager.setClientAccount(vimeoAccount);
     }
 
     @Override
-    public void saveAccount(VimeoAccount vimeoAccount, String email) {
+    public void saveAuthenticatedUserAccount(@NonNull VimeoAccount vimeoAccount,
```
kevinzetterstrom: I do not think we need to confirm that it is a user account because 1. this is a simple demo that is just meant to show that you can respond to these callbacks, and 2. Now that we have split the method into two, this one should always be representative of a user account.

kevinzetterstrom:
@alfiehanssen responded here: fa92e97, let me know if you have any additional comments.

alfiehanssen:
Thanks @kevinzetterstrom, I left a few comments questions on the new commit. 👍

kevinzetterstrom:
@alfiehanssen responded here: 2d24549

PR #153, opened 2016-10-15
ANDTV-216 Adding watched videos

howardrigberg:
#### Ticket
**Required for Vimeans only**
[ANDTV-216](https://vimean.atlassian.net/browse/ANDTV-216)

#### Ticket Summary
Adding watched_videos to user.metadata.connections to support watched history.

#### Implementation Summary
A trivial one-liner.

#### How to Test
Nothing to test.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
@@ -234,6 +234,15 @@ public class User implements Serializable, Followable {
         }
         return null;
     }
+
+    @Nullable
+    public Connection getWatchedVideosConnection() {
+        if ((metadata != null) && (metadata.connections != null) &&
+            (metadata.connections.watchedVideos != null)) {
+            return metadata.connections.watchedVideos;
+        }
+        return null;
+    }
     // </editor-fold>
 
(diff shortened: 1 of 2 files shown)


kevinzetterstrom:
Looks good. My only suggestions is that you add a helper on the User to grab the connection

howardrigberg:
Added the accessor

vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
```diff
@@ -234,6 +234,15 @@ public Connection getWatchLaterConnection() {
         }
         return null;
     }
+
+    @Nullable
+    public Connection getWatchedVideosConnection() {
+        if ((metadata != null) && (metadata.connections != null) &&
+            (metadata.connections.watchedVideos != null)) {
+            return metadata.connections.watchedVideos;
```
kevinzetterstrom: Nit: Since this method could return null, this if could be simplified to 
```
if ((metadata != null) && (metadata.connections != null)) {
   return metadata.connections.watchedVideos;
}

return null;
```

kevinzetterstrom:
This is super clean. 💯 👍 🚵

PR #154, opened 2016-10-17
ANDTV-224 Vod Seasons 

kevinzetterstrom:
#### Ticket
[ANDTV-224](https://vimean.atlassian.net/browse/ANDTV-224)

#### Ticket Summary
In order to dynamically add seasons to the rows fragment on tv, we will need to get the seasons uri from the API. Also we should have simple helper accessing methods for this.

#### Implementation Summary
I added in the connection and some simple helpers for ```VodItem``` and ```Season```

#### How to Test
Nothing to test


howardrigberg:
Looks good 🍔

PR #158, opened 2016-11-08
ANDTV-81 changes for 3.3.1

kevinzetterstrom:
#### Ticket
[ANDTV-81](https://vimean.atlassian.net/browse/ANDTV-81)

#### Ticket Summary
We are changing the logging system to make things simpler. This will add changes to a few other areas, as we need to upgrade several things at once - we will be targeting 3.3.1

#### Implementation Summary
* ```PictureCollection``` now has a ```type``` field, rather than a ```default``` boolean
* ```User.preferences.videos``` now has the standard ```Privacy``` type, rather than a string
* ```VideoLog``` is removed, in favor of a single string endpoint for logging
* Some constants are added for constructing play logging requests

I updated a few classes to not use public members, and changes their names to be more forward. I did not update every class I touched, because that would be insane.

#### How to Test
Use the same branch in mobile/vimeo-kit and Stetho to test that play logs send


kevinzetterstrom:
@howardrigberg 3d6bac5 should address all your comments of this PR. I only added javadocs on methods of VideosPreference, Preferences, and PictureCollection, since clean up of files really goes beyond the scope of this PR and I don't want to get into a rabbit hole.

PR #160, opened 2016-11-11
DTA-39 adding category icon

kevinzetterstrom:
#### Ticket
[DTA-39](https://vimean.atlassian.net/browse/DTA-39)

#### Ticket Summary
For Day Tripper, we want to show the category icon on the cards. We should do this dynamically, as the API recently provided support for this. This PR is about adding that API support. This is not about updating Category and all of its members and methods, this is just about the additional field.

#### Implementation Summary
Added in the additional field for the icon, along with a getter.


kvenn:
Looks good! Nice work adding annotations and such.

MWR

PR #163, opened 2016-12-15
VA-2055 add subscription models and service

kevinzetterstrom:
#### Ticket
[VA-2055](https://vimean.atlassian.net/browse/VA-2055)

#### Ticket Summary
Create new models for subscriptions, and a way to edit them

#### Implementation Summary
Created new models, and added a new service method and client method to edit the subscriptions

#### How to Test
Use the Vimeo-Android branch to test


howardrigberg:
I will get to this now.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -949,6 +950,27 @@ public void failure(VimeoError error) {
...
+     *
+     * @param subscriptionMap a map of the subscriptions you wish to change
+     * @param callback        callback to be invoked when the request finishes
+     * @return A Call so that the request can be cancelled if need be
+     */
+    @Nullable
+    public Call<SubscriptionCollection> editSubscriptions(@NotNull Map<String, Boolean> subscriptionMap,
+                                                          @NotNull
+                                                                  ModelCallback<SubscriptionCollection> callback) {
+        if (callback == null) {
```
howardrigberg: Null check not needed if you are annotating. 

If you feel ambitious, you could change the other methods in this class that take a callback to have the  @NotNull annotation too and remove their null checks. Obviously that extra work is not required.

vimeo-networking/src/main/java/com/vimeo/networking/model/SubscriptionCollection.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.vimeo.stag.GsonAdapterKey;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+import java.util.Date;
+
+/**
+ * A representation of the subscriptions that a user has.
```
howardrigberg: {@link Subscriptions}

vimeo-networking/src/main/java/com/vimeo/networking/model/Subscriptions.java
```diff
@@ -0,0 +1,108 @@
...
+ * has both getters and setters since a user can update their subscriptions
+ * using a Patch.
+ * <p>
+ * Created by zetterstromk on 12/15/16.
+ */
+public class Subscriptions implements Serializable {
+
+    private static final long serialVersionUID = 3088065484753327987L;
+
+    public static final String KEY_COMMENT = "comment";
```
howardrigberg: So these are fixed as opposed to dynamic key/value pairs? How are they going to be changed going forward?

vimeo-networking/src/main/java/com/vimeo/networking/model/SubscriptionCollection.java
kevinzetterstrom: I have a ```@see Subscriptions``` on the next line, but can add that here too

vimeo-networking/src/main/java/com/vimeo/networking/model/Subscriptions.java
kevinzetterstrom: These are fixed, just like every other property we receive from our api - if our API wishes to change them in the future, then they would need to properly deprecate them or make the change behind a new API version.

kevinzetterstrom:
@howardrigberg responded here cc4149c

vimeo-networking/src/main/java/com/vimeo/networking/model/Subscriptions.java
```diff
@@ -0,0 +1,108 @@
...
+    @GsonAdapterKey(KEY_REPLY)
+    boolean mReply;
+
+    @GsonAdapterKey(KEY_FOLLOW)
+    boolean mFollow;
+
+    @GsonAdapterKey(KEY_VIDEO_AVAILABLE)
+    boolean mVideoAvailable;
+
+    public boolean isComment() {
```
howardrigberg: In looking at this again, I realized that the naming could be changed a little to more accurately represent what the methods do. How about this pattern: 
```
public void receiveThing(boolean receive)
public boolean isReceivingThing() 
```

For example: 

```
public void receiveLikes(boolean receive)
public boolean isReceivingLikes()
```

kevinzetterstrom: I can get down with that name change. I made it here: cf8ce0a

howardrigberg:
👍 looks good. mwr

PR #164, opened 2016-12-16
CI-89 buildbotting

wilrnh:


wilrnh:
@kevinzetterstrom @anthonycr or @kvenn lemme know if this is aok with you guys, particularly the gitignore in `example/src/main/res/values/`

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  desc "test"
+  lane :test do
+    
+    # gotta do some setup before running gradle
+    sh 'touch ../local.properties'
+
+    File.open("../example/src/main/res/values/secrets.xml", "w"){
```
kevinzetterstrom: Out of curiosity, why put the secrets file here instead of at the top level?

PR #166, opened 2017-01-04
Add spatial data representation to video object

anthonycr:
#### Summary

Adding spatial data to the video object for 360 videos.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
@@ -3,7 +3,6 @@ package com.vimeo.networking.model;
 import com.vimeo.stag.GsonAdapterKey;
 
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
 import java.io.Serializable;
 
@@ -13,6 +12,7 @@ import java.io.Serializable;
  * <p>
  * Created by restainoa on 1/3/17.
  */
+@SuppressWarnings("unused")
 public class Spatial implements Serializable {
 
     private static final long serialVersionUID = 5660325676029549468L;
@@ -22,13 +22,15 @@ public class Spatial implements Serializable {
         CYLINDRICAL,
         CUBICAL,
         PYRAMID,
-        DOME
+        DOME,
+        UNKNOWN
     }
 
     public enum Format {
         MONO,
         LEFT_RIGHT,
-        TOP_BOTTOM
+        TOP_BOTTOM,
+        UNKNOWN
     }
 
     @NotNull
@@ -39,7 +41,7 @@ public class Spatial implements Serializable {
     @GsonAdapterKey("stereo_format")
     protected String mStereoFormat;
 
-    @Nullable
+    @NotNull
     public Projection getProjection() {
         switch (mProjection) {
             case "equirectangular":
@@ -53,11 +55,11 @@ public class Spatial implements Serializable {
             case "dome":
                 return Projection.DOME;
             default:
-                return null;
+                return Projection.UNKNOWN;
         }
     }
 
-    @Nullable
+    @NotNull
     public Format getFormat() {
         switch (mStereoFormat) {
             case "mono":
@@ -67,7 +69,7 @@ public class Spatial implements Serializable {
             case "top-bottom":
                 return Format.TOP_BOTTOM;
             default:
-                return null;
+                return Format.UNKNOWN;
         }
     }
 }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
@@ -1,0 +1,73 @@
+package com.vimeo.networking.model;
+
+import com.vimeo.stag.GsonAdapterKey;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+
+/**
+ * The spatial object representing the format
+ * data for a 360 video object.
+ * <p>
+ * Created by restainoa on 1/3/17.
+ */
+public class Spatial implements Serializable {
+
+    private static final long serialVersionUID = 5660325676029549468L;
+
+    public enum Projection {
+        EQUIRECTANGULAR,
+        CYLINDRICAL,
+        CUBICAL,
+        PYRAMID,
+        DOME
+    }
+
+    public enum Format {
+        MONO,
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
```diff
@@ -0,0 +1,73 @@
...
+            case "cylindrical":
+                return Projection.CYLINDRICAL;
+            case "cubical":
+                return Projection.CUBICAL;
+            case "pyramid":
+                return Projection.PYRAMID;
+            case "dome":
+                return Projection.DOME;
+            default:
+                return null;
```
anthonycr: should we annotate this `@NotNull` instead and throw an exception here? That might be beneficial since the `@NotNull` annotations on the `mProjection` and `mStereoFormat` fields will ensure that they are never null. Or we should we add in an `UNSUPPORTED` enum field so that we don't have to deal with null?

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -118,68 +118,99 @@ public String toString() {
...
     @GsonAdapterKey("download")
     public ArrayList<VideoFile> download;
+
     @Nullable
     @GsonAdapterKey("badge")
     public VideoBadge videoBadge;
 
+    @Nullable
+    @GsonAdapterKey("spatial")
+    public Spatial spatial;
```
kevinzetterstrom: Want to add a getter for this?

vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
```diff
@@ -0,0 +1,73 @@
...
+        DOME
+    }
+
+    public enum Format {
+        MONO,
+        LEFT_RIGHT,
+        TOP_BOTTOM
+    }
+
+    @NotNull
```
anthonycr: I decided to annotate these fields `@NotNull` since this class should always contain these fields from what I've seen the api return

kevinzetterstrom: Dope! Loving the NotNull enforcement with Stag!

vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
```diff
@@ -0,0 +1,73 @@
...
+    public Format getFormat() {
+        switch (mStereoFormat) {
+            case "mono":
+                return Format.MONO;
+            case "left-right":
+                return Format.LEFT_RIGHT;
+            case "top-bottom":
+                return Format.TOP_BOTTOM;
+            default:
+                return null;
```
anthonycr: same comment here as above about potentially throwing an exception

vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Nullable
+    public Format getFormat() {
+        switch (mStereoFormat) {
+            case "mono":
+                return Format.MONO;
+            case "left-right":
+                return Format.LEFT_RIGHT;
+            case "top-bottom":
```
kevinzetterstrom: Can you move all of the strings to constants at the top of the file?

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -118,68 +118,99 @@ public String toString() {
...
+
     @Nullable
     @GsonAdapterKey("download")
     public ArrayList<VideoFile> download;
+
     @Nullable
     @GsonAdapterKey("badge")
     public VideoBadge videoBadge;
 
+    @Nullable
```
anthonycr: non-360 videos do not contain spatial data, so this field will be null

vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
```diff
@@ -0,0 +1,73 @@
...
+        TOP_BOTTOM
+    }
+
+    @NotNull
+    @GsonAdapterKey("projection")
+    protected String mProjection;
+
+    @NotNull
+    @GsonAdapterKey("stereo_format")
+    protected String mStereoFormat;
```
kevinzetterstrom: I must have missed when Stag started allowing for protected members, I thought they had to be public of package-private. Nice to see this!

vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
kevinzetterstrom: I think we should keep these as they are and return null - that is basically the point of having a string over an enum anyway. It lets the API add a value that the client is unaware of. We can keep it Nullable and return null as you do here, or have a ```NONE``` (or similar) type in each enum that is returned, and then have the method be NotNull.

kevinzetterstrom: Oops - I missed the ```UNSUPPORTED``` type you mentioned above. I think either case is good. I am kind of leaning towards NotNull and returning a default type that signifies it is unknown.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
kevinzetterstrom: Also, we could add a helper method here too - ```is360Video()``` or similar. We have similar helpers in ```Video``` for ```isVod()```

anthonycr: I figured I would just leave it public for now as I think the Video object needs an overhaul to make all the member variables protected/package private and then add getters, so I didn't want to just add a single getter for this field

kevinzetterstrom: Want to add a ```is360Video()```? I think it would be super beneficial, just like the isVod one is. Also, if I am looking at this class I may not fully understand what Spatial is if I am not familiar with 360 video. Having a 360 specific method could definitely help that.

anthonycr: I'll add the `is360Video` method

vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
anthonycr: I don't think moving these to constants benefits us, because it encourages using the string values rather than the enums that the strings are being mapped to. Let's discuss if you think they should be constants still

PR #167, opened 2017-01-10
Adding equals, hashcode, and toString for spatial object

anthonycr:
Adding `equals`, `hashCode`, and `toString` for `Spatial` object

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
@@ -72,4 +72,34 @@ public class Spatial implements Serializable {
                 return Format.UNKNOWN;
         }
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Spatial.java
```diff
@@ -72,4 +72,34 @@ public Format getFormat() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        Spatial spatial = (Spatial) o;
+
+        return mProjection.equals(spatial.mProjection) && mStereoFormat.equals(spatial.mStereoFormat);
```
kvenn: I see they fields are annotated NotNull - but can we guarantee that?

anthonycr: Stag will guarantee that they are `NotNull`. However, if we construct the object outside of the stag environment, then we cannot guarantee it. This is an interesting problem, I will add this to the sync agenda next week

kvenn: Good idea - MWR

PR #168, opened 2017-01-11
VA-1849 notification models

kevinzetterstrom:
#### Ticket
[VA-1849](https://vimean.atlassian.net/browse/VA-1849)

#### Ticket Summary
In order to create the notification screen within the Android app, we need to have the models. 

#### Implementation Summary
As the API is going to be using the same keys for all types, I broke the types out into an enum and a constants file. I moved the subscription classes into the notification package since they are related.

I created two API issues
* https://github.vimeows.com/Vimeo-API/api/issues/819 the uri is always null
* https://github.vimeows.com/Vimeo-API/api/issues/820 the response doesn't contain the state (new/seen). I will follow up with another networking PR once that is finalized.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
@@ -58,10 +58,10 @@ public class Connection implements Serializable {
     public int viewableTotal;
 
     @GsonAdapterKey("new_total")
-    public int newTotal;
+    int newTotal;
 
     @GsonAdapterKey("unread_total")
-    public int unreadTotal;
+    int unreadTotal;
 
     @Nullable
     @GsonAdapterKey("name")
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
@@ -57,6 +57,12 @@ public class Connection implements Serializable {
     @GsonAdapterKey("viewable_total")
     public int viewableTotal;
 
+    @GsonAdapterKey("new_total")
+    public int newTotal;
+
+    @GsonAdapterKey("unread_total")
+    public int unreadTotal;
+
     @Nullable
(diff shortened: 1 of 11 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
```diff
@@ -57,6 +57,12 @@
     @GsonAdapterKey("viewable_total")
     public int viewableTotal;
 
+    @GsonAdapterKey("new_total")
+    public int newTotal;
+
+    @GsonAdapterKey("unread_total")
+    public int unreadTotal;
```
anthonycr: since we have getters below for these fields, can we make them package local or protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/Credit.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+
+/**
+ * A model representing a credit.
+ * Created by zetterstromk on 1/11/17.
+ */
+public class Credit implements Serializable {
```
anthonycr: Can you generate `equals`, `toString`, and `hashCode` methods for this class

vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/Notification.java
```diff
@@ -0,0 +1,84 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+import java.util.Date;
+
+/**
+ * Created by zetterstromk on 1/11/17.
+ */
+public class Notification implements Serializable {
```
anthonycr: Can you generate `equals`, `toString`, and `hashCode` methods for this class

kevinzetterstrom:
@anthonycr changes here: 24e00a7

PR #170, opened 2017-01-18
VA-1850 Notification updates

kevinzetterstrom:
#### Ticket
[VA-1850](https://vimean.atlassian.net/browse/VA-1850)

#### Ticket Summary
In order to properly handle object updates, we need to add some setters to Notification and Credit

#### Implementation Summary
I added setters for the two models, and I fixed a NotificationType bug that associated Likes to Comments

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Credit.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Credit.java
@@ -54,9 +54,17 @@ public class Credit implements Serializable {
         return mVideo;
     }
 
+    public void setVideo(@Nullable Video video) {
+        mVideo = video;
+    }
+
     @Nullable
     public User getUser() {
         return mUser;
+    }
+
+    public void setUser(@Nullable User user) {
+        mUser = user;
     }
 
     @Override
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationType.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationType.java
@@ -25,7 +25,7 @@ public enum NotificationType {
             case NotificationConstants.NOTIFICATION_FOLLOW:
                 return NOTIFICATION_TYPE_FOLLOW;
             case NotificationConstants.NOTIFICATION_LIKE:
-                return NOTIFICATION_TYPE_COMMENT;
+                return NOTIFICATION_TYPE_LIKE;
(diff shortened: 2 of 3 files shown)


anthonycr:
why do we need the setters on the model objects?

kevinzetterstrom:
Sorry, I should have been more thorough in the description. We need setters so that we can modify in-memory lists. This is currently how our object tracking is set up. We don't have setters throughout the code base because most of the variables are public so we can just do something like:

```
video.user = updatedUserObject;

notifyAdapterForVideoUpdate(video, positionOfVideoInList);

```

With these newer models and the enhancements made to Stag, I have gone away from public variables to package-private. Until we come up with a better object tracking method, we will need setters so we can do the above with other types.

anthonycr:
Thanks for clearing that up

vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationType.java
```diff
@@ -25,7 +25,7 @@ public static NotificationType notificationTypeFromString(@NotNull String type)
             case NotificationConstants.NOTIFICATION_FOLLOW:
                 return NOTIFICATION_TYPE_FOLLOW;
             case NotificationConstants.NOTIFICATION_LIKE:
-                return NOTIFICATION_TYPE_COMMENT;
+                return NOTIFICATION_TYPE_LIKE;
```
anthonycr: nice bug lol

vimeo-networking/src/main/java/com/vimeo/networking/model/Credit.java
```diff
@@ -54,11 +54,19 @@ public Video getVideo() {
         return mVideo;
     }
 
+    public void setVideo(@Nullable Video video) {
```
anthonycr: Do we want to make these `NonNull` instead? I know the field itself would still be nullable, but we are using these methods to update the fields, so it seems like we shouldn't allow someone to null out a non null field. What do you think?

anthonycr: Sounds good, I have already put field nullability on the agenda so this will fit in nicely

kevinzetterstrom: Looks like we are keeping this as is based on IRL discussion.

anthonycr: 👍 merge it

PR #173, opened 2017-01-23
VA-2112 Add toString for notification analytics

kevinzetterstrom:
#### Ticket
[VA-2112](https://vimean.atlassian.net/browse/VA-2112)

#### Ticket Summary
Adding a toString method to NotificationType enum for analytics

#### Implementation Summary
I created a toString and a static method for getting a string key from an enum value for `NotificationType`. I opted for this approach versus a string constructor for the enum values due to the future expansion of types - I don't want the app to crash if we get additional types, instead, the type will be converted to Unknown using this approach.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationType.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationType.java
@@ -36,11 +36,6 @@ public enum NotificationType {
         }
     }
 
-    @Override
-    public String toString() {
-        return stringFromNotificationType(this);
-    }
-
     @Nullable
     public static String stringFromNotificationType(@NotNull NotificationType type) {
         switch (type) {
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationType.java
```diff
@@ -35,4 +36,30 @@ public static NotificationType notificationTypeFromString(@NotNull String type)
         }
     }
 
+    @Override
+    public String toString() {
+        return stringFromNotificationType(this);
```
anthonycr: I don't think this is a good idea. The default `toString` of an enum will return the name of the enum value. If the purpose of changing this is to return a serialized value of the enum, then we should create a different method than overriding toString.

kevinzetterstrom:
@anthonycr good call: fix here 6179c25

PR #174, opened 2017-01-23
VA-1890 Patch requests for notifications

kevinzetterstrom:
#### Ticket
[VA-1890](https://vimean.atlassian.net/browse/VA-1890)

#### Ticket Summary
We need to issue patch requests for notifications. One of the bodies contains a map, the other a list. Neither request has a return object.

#### Implementation Summary
I created a generic `Void` `Call` for patches. I use an `Object` as the body to let it be decided by the consumer what type to pass. I also added a few parameter constants.


howardrigberg:
📟 🇿🇲 💴 🇶🇦 Merge when ready

PR #176, opened 2017-01-31
Updating to Stag 2.0.1

anthonycr:
#### Summary
- Updating to stag 2.0.1
- Fixing bug exposed by Stag where the type we exposed in `privacy.add` and `privacy.download` was integer instead of boolean.


kevinzetterstrom:
Dope. MWR

PR #177, opened 2017-02-01
Don’t use Stag for serialization of Privacy object

anthonycr:
#### Summary
Don’t use Stag for serialization of Privacy object

#### Implementation Summary
Due to the `Privacy.add` and `Privacy.download` fields being sent from the server as integers, we need to deserialize these as integers and not as booleans. Gson and Stag don't support integer to boolean conversion, so we need to create a custom adapter that does it for us.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/utils/PrivacyDeserializer.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/utils/PrivacyDeserializer.java
@@ -1,0 +1,150 @@
+package com.vimeo.networking.utils;
+
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.reflect.TypeToken;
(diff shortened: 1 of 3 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/utils/PrivacyDeserializer.java
```diff
@@ -0,0 +1,150 @@
...
+                        object.view = mPrivacyValueTypeAdapter.read(reader);
+                        break;
+                    case "embed":
+                        object.embed = com.google.gson.internal.bind.TypeAdapters.STRING.read(reader);
+                        break;
+                    case "download":
+                        if (jsonToken == JsonToken.NUMBER) {
+                            object.download = reader.nextInt() == 1;
+                        } else {
+                            object.download = reader.nextBoolean();
```
kevinzetterstrom: :1st_place_medal:

PR #178, opened 2017-02-14
model package cleanup

anthonycr:
#### Summary
- Cleaning up the model package
    - Moving to `@UseStag` and away from `@GsonAdapterKey`
    - Adding stag to more models
    - Getting rid of any hungarian notation naming in member variables
    - Cleaning up some lint problems
    - Suppressing unused warnings
    - Adding spacing between members
    - Added getters for all fields and reduce all field visibility to non-public

#### TODO (maybe)
- `appconfiguration` package has been ignored because those classes have some problems. (See ticket ADF-100)
- Some sort of unit test validation that models are generated?
- Update README with new field names/getters

PR #181, opened 2017-02-21
ADF-3 Search suggestion models

kevinzetterstrom:
#### Ticket
[ADF-3](https://vimean.atlassian.net/browse/ADF-3)

#### Ticket Summary
We need to create some models to represent the response from `https://api.vimeo.com/suggest`

#### Implementation Summary
I created the necessary models, using a base class to contains the `text` field, since all suggestions will always have this. I also added in `toString`, `equals`, and `hashCode` methods.

#### How to Test
Using Postman or another rest client, hit `https://api.vimeo.com/suggest?query=cats&video_count=5&ondemand_title_count=5` and compare the response with the models created.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/search/BaseSuggestion.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/search/BaseSuggestion.java
@@ -1,0 +1,31 @@
+package com.vimeo.networking.model.search;
+
+import com.google.gson.annotations.SerializedName;
+import com.vimeo.stag.UseStag;
+import com.vimeo.stag.UseStag.FieldOption;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.Serializable;
+
+/**
+ * A base class for search suggestions. Each suggest must contain a <code>text</code>
+ * string.
+ * <p>
+ * Created by zetterstromk on 2/21/17.
+ */
+@UseStag(FieldOption.SERIALIZED_NAME)
+abstract class BaseSuggestion implements Serializable {
+
+    @NotNull
+    @SerializedName("text")
+    String mText;
+
+    /**
+     * Returns the title associated with this suggestion.
+     */
+    @NotNull
+    public String getText() {
+        return mText;
+    }
+}
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/search/OndemandSuggestion.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/search/OndemandSuggestion.java
@@ -67,7 +67,7 @@ public final class OndemandSuggestion extends BaseSuggestion implements Serializ
     }
 
     @UseStag(FieldOption.SERIALIZED_NAME)
-    public static class OndemandSuggestionMetadata implements Serializable {
+    public final static class OndemandSuggestionMetadata implements Serializable {
 
         private static final long serialVersionUID = 7538947801290601850L;
 
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/search/OndemandSuggestion.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/search/OndemandSuggestion.java
@@ -67,7 +67,7 @@ public class OndemandSuggestion extends BaseSuggestion implements Serializable {
     }
 
     @UseStag(FieldOption.SERIALIZED_NAME)
-    static class OndemandSuggestionMetadata implements Serializable {
+    public static class OndemandSuggestionMetadata implements Serializable {
 
         private static final long serialVersionUID = 7538947801290601850L;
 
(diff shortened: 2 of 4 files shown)


kevinzetterstrom:
Reassigning since Anthony reviewed the PR based off of this.

vimeo-networking/src/main/java/com/vimeo/networking/model/search/OndemandSuggestion.java
```diff
@@ -0,0 +1,83 @@
...
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "OndemandSuggestion{" +
+               "mText='" + mText + '\'' +
+               ", mMetadata=" + mMetadata +
+               '}';
+    }
```
anthonycr: 👍 glad to see equals/hashCode/toString

vimeo-networking/src/main/java/com/vimeo/networking/model/search/SuggestionResponse.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.Serializable;
+import java.util.ArrayList;
+
+/**
+ * A response sent from the API for search/autocomplete suggestions.
+ * <p>
+ * Created by zetterstromk on 2/21/17.
+ */
+@UseStag(FieldOption.SERIALIZED_NAME)
+public class SuggestionResponse implements Serializable {
```
anthonycr: It's too bad that this response can't extend from `BaseResponseList` but I guess it's fundamentally a different type of data (i.e. not pageable)

vimeo-networking/src/main/java/com/vimeo/networking/model/search/SuggestionResponse.java
```diff
@@ -0,0 +1,68 @@
...
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "SuggestionResponse{" +
+               "mOndemandSuggestions=" + mOndemandSuggestions +
+               ", mVideoSuggestions=" + mVideoSuggestions +
+               '}';
+    }
```
anthonycr: 👍

vimeo-networking/src/main/java/com/vimeo/networking/model/search/VideoSuggestion.java
```diff
@@ -0,0 +1,69 @@
...
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "VideoSuggestion{" +
+               "mText='" + mText + '\'' +
+               ", mMetadata=" + mMetadata +
+               '}';
+    }
```
anthonycr: 👍 for these methods

vimeo-networking/src/main/java/com/vimeo/networking/model/search/BaseSuggestion.java
```diff
@@ -0,0 +1,31 @@
...
+ * string.
+ * <p>
+ * Created by zetterstromk on 2/21/17.
+ */
+@UseStag(FieldOption.SERIALIZED_NAME)
+abstract class BaseSuggestion implements Serializable {
+
+    @NotNull
+    @SerializedName("text")
+    String mText;
```
anthonycr: I know we talked about this yesterday, but I've done some thinking and we need to make all our member variables `protected` instead of just package local so that we can extend them from other packages (I know it isn't relevant for this class since the class itself is package-local). This comment applies to only those fields added in this PR, the rest I have taken care of in https://github.com/vimeo/vimeo-networking-java/pull/178

- [ ] Change the visibility of all fields in `public` classes added in this PR to `protected`.

kevinzetterstrom: Ok, I think I can be onboard with that.

kevinzetterstrom:
@anthonycr I opted to make the classes final rather than making the members protected. 2d50d18

vimeo-networking/src/main/java/com/vimeo/networking/model/search/OndemandSuggestion.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @UseStag(FieldOption.SERIALIZED_NAME)
+    public static class OndemandSuggestionMetadata implements Serializable {
```
anthonycr: This class should also be final

vimeo-networking/src/main/java/com/vimeo/networking/model/search/VideoSuggestion.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @UseStag(FieldOption.SERIALIZED_NAME)
+    public static class VideoSuggestionMetadata implements Serializable {
```
anthonycr: This class should also be final

PR #184, opened 2017-02-21
Only run Danger when a Github token exists

wilrnh:
External PRs break Travis because env vars aren't available to them.


wilrnh:
cc @kevinzetterstrom @anthonycr @howardrigberg @kvenn

PR #186, opened 2017-02-22
VA-2203 Notification connection for type counts

kevinzetterstrom:
#### Ticket
[VA-2203](https://vimean.atlassian.net/browse/VA-2203)

#### Ticket Summary
We need to add a new object onto the notification connection for type counts.

Also, somehow my copyright profile got messed up and wasn't adding copyrights, so I went back and added them to files I recently added.

#### Implementation Summary
I decided that since the new fields only applied to the notification `Connection`, that it'd be better to remove the fields from `Connection` and make a specialized `NotificationConnection` object. This has all the great fields of `Connection` with the additional properties sent specific to the `notification` connection.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
@@ -57,12 +57,6 @@ public class Connection implements Serializable {
     @GsonAdapterKey("viewable_total")
     public int viewableTotal;
 
-    @GsonAdapterKey("new_total")
-    int newTotal;
-
-    @GsonAdapterKey("unread_total")
-    int unreadTotal;
-
     @Nullable
     @GsonAdapterKey("name")
     public String name;
@@ -95,14 +89,6 @@ public class Connection implements Serializable {
 
     public int getViewableTotal() {
         return viewableTotal;
-    }
-
-    public int getNewTotal() {
-        return newTotal;
-    }
-
-    public int getUnreadTotal() {
-        return unreadTotal;
     }
 
     @Nullable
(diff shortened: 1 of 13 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationConnection.java
```diff
@@ -0,0 +1,199 @@
...
+ * <p>
+ * Created by zetterstromk on 2/22/17.
+ */
+@UseStag(FieldOption.SERIALIZED_NAME)
+public class NotificationConnection extends Connection {
+
+    private static final long serialVersionUID = 4908222195478449252L;
+
+    @SerializedName("new_total")
+    protected int newTotal;
```
anthonycr: should be `mNewTotal`

vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationConnection.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @SerializedName("unread_total")
+    protected int unreadTotal;
```
anthonycr: should be `mUnreadTotal`

vimeo-networking/src/main/java/com/vimeo/networking/model/Connection.java
```diff
@@ -57,12 +57,6 @@
     @GsonAdapterKey("viewable_total")
     public int viewableTotal;
 
-    @GsonAdapterKey("new_total")
-    int newTotal;
-
-    @GsonAdapterKey("unread_total")
-    int unreadTotal;
-
```
anthonycr: Nice to be able to remove these from Connection since they aren't applicable anywhere but in Notifications

vimeo-networking/src/main/java/com/vimeo/networking/model/notifications/NotificationConnection.java
kevinzetterstrom: 🙈

kevinzetterstrom:
@anthonycr renamed 315fe4a

PR #193, opened 2017-03-15
ADF-118 SVOD Model Changes

kevinzetterstrom:
#### Ticket
[ADF-118](https://vimean.atlassian.net/browse/ADF-118)

#### Ticket Summary
We need to make changes to models for SVOD. The API side is not yet done, but the blueprint is here: https://github.vimeows.com/Vimeo-API/blueprint/pull/13/files

#### Implementation Summary
I added a new `Interaction` subclass, `SvodInteraction` to contain the interaction data specific to the SVOD representation.

I added helpers to access this interaction on `Video` and `User`

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
@@ -27,7 +27,6 @@ import com.vimeo.networking.Vimeo;
 import com.vimeo.networking.model.Privacy.PrivacyValue;
 import com.vimeo.networking.model.UserBadge.UserBadgeType;
 import com.vimeo.networking.model.notifications.NotificationConnection;
-import com.vimeo.networking.model.vod.SvodInteraction;
 import com.vimeo.stag.UseStag;
 
 import org.jetbrains.annotations.NotNull;
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
@@ -27,6 +27,7 @@ import com.vimeo.networking.Vimeo;
 import com.vimeo.networking.model.Privacy.PrivacyValue;
 import com.vimeo.networking.model.UserBadge.UserBadgeType;
 import com.vimeo.networking.model.notifications.NotificationConnection;
+import com.vimeo.networking.model.vod.SvodInteraction;
 import com.vimeo.stag.UseStag;
 
 import org.jetbrains.annotations.NotNull;
@@ -160,7 +161,7 @@ public class User implements Serializable, Followable {
     @Nullable
     private ConnectionCollection getMetadataConnections() {
         if (mMetadata != null) {
-            return mMetadata.mConnections;
+            return mMetadata.getConnections();
         }
         return null;
     }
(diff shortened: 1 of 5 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -734,6 +743,34 @@ public Date getSubscriptionExpiration() {
...
+     * or if the user is not subscribed to SVOD
+     */
+    @Nullable
+    public Date getSvodExpiration() {
+        if (isSvod()) {
+            // isSubscription will validate and prevent possible npes
+            assert mMetadata.getInteractions().getSvod() != null;
+            return mMetadata.getInteractions().getSvod().getExpiration();
+        }
+        return null;
```
anthonycr: The underlying implementation of `isSvod` is subject to change, so using it as a null check here is incorrect.

This method could be simplified to:

```java
SvodInteraction svodInteraction = getSvodInteraction();
return svodInteraction != null ? svodInteraction.getExpiration() : null;
```

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -734,6 +743,34 @@ public Date getSubscriptionExpiration() {
...
+     * or if the user is not subscribed to SVOD
+     */
+    @Nullable
+    public Date getSvodPurchaseDate() {
+        if (isSvod()) {
+            // isSubscription will validate and prevent possible npes
+            assert mMetadata.getInteractions().getSvod() != null;
+            return mMetadata.getInteractions().getSvod().getPurchaseDate();
+        }
+        return null;
```
anthonycr: The same comment applied to `getSvodExpiration` about the faulty null check also applies to this method.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -756,6 +793,26 @@ public boolean isSubscription() {
...
     }
 
+    /**
+     * Helper to determine if this video is part of the SVOD library and included in the subscription. If wanting
+     * to know if a user also has access to the video, use {@link #canAccessVideoInSvod()}
+     *
+     * @return true if this video can be accessed with an SVOD subscription
+     */
+    public boolean isSvod() {
+        return mMetadata.getInteractions().getSvod() != null;
```
anthonycr: Should just be:
```java
getSvodInteraction() != null;
```

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -638,6 +638,15 @@ public long getPlayProgressMillis() {
     // <editor-fold desc="VOD">
 
     /**
+     * @return the {@link SvodInteraction}. Will be null if this video is not part of SVOD
+     */
+    @Nullable
+    public SvodInteraction getSvodInteraction() {
+        InteractionCollection interactions = mMetadata.getInteractions();
```
anthonycr: Do you think we should check the metadata object for nullity, or can we safely assume it is not null? It's not marked as either nonnull or nullable, and it seems that we sometimes check to see if it's null and sometimes don't in the code.

vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
```diff
@@ -160,102 +160,122 @@ public boolean isFollowing() {
...
         ConnectionCollection connections = getMetadataConnections();
-        return connections != null ? connections.mWatchlater : null;
+        return connections != null ? connections.getWatchlater() : null;
     }
 
     @Nullable
     public Connection getWatchedVideosConnection() {
         ConnectionCollection connections = getMetadataConnections();
-        return connections != null ? connections.mWatchedVideos : null;
+        return connections != null ? connections.getWatchedVideos() : null;
```
anthonycr: 👍 good changes to move away from referencing the member variables themselves.

kevinzetterstrom:
Thanks for the review, here are the changes d296acf

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -756,6 +793,26 @@ public boolean isSubscription() {
...
+        return mMetadata.getInteractions().getSvod() != null;
+    }
+
+    /**
+     * Helper to determine if the video is part of the SVOD library and that the user currently has access to it.
+     * This will be true if the user has purchased an SVOD subscription that is active.
+     *
+     * @return true if this video can be accessed with an SVOD subscription and the user's subscription is active
+     */
+    public boolean canAccessVideoInSvod() {
```
kevinzetterstrom: Since they are helpers around metadata and pertain to all videos, I think it is OK to stay in this class.

PR #195, opened 2017-03-30
Backwards compatibility

anthonycr:
#### Summary
Due to a problem where we didn't use the stag serializer in some code, the field names were all broken because of the `@GsonAdapterKey` annotation. Since we have now replaced that annotation with the `@SerializedName` annotation which is picked up by gson regardless of whether the stag plugin is used, we need to add backwards compatibility for the broken cached json. The only classes that were affected were those that are referenced by `Video` and `VideoFile`.

#### Implementation Summary
By using the alternate name scheme, we can add these optional names as json keys to pick up what was broken. When these classes are serialized again, the `value` is used and the `alternate` values are ignored.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/PictureCollection.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/PictureCollection.java
@@ -65,18 +65,18 @@ public class PictureCollection implements Serializable {
     }
 
     @Nullable
-    @SerializedName("uri")
+    @SerializedName(value = "uri", alternate = "m_uri")
     protected String mUri;
 
-    @SerializedName("active")
+    @SerializedName(value = "active", alternate = "m_is_active")
     protected boolean mIsActive;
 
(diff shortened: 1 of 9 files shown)


kevinzetterstrom:
What about for values serialized before we had changed the names to hungarian? Is that an issue?

anthonycr:
@kevinzetterstrom So in our app's version 2.4.0, we had some fields using hungarian notation, but most as not using it. As of our app's 2.5.0 version, all fields are using hungarian notation. Anything we serialized in version 2.4.0 without using the stag serializer will default to using the field name (camel case separated by underscores). Since our naming scheme was basically the reversal of that, we don't have to bother with fields that were not using hungarian notation (actually there are a few that I should fine).

Tl;dr; it's not an issue, except for a few stray fields who's keys don't match their variable names

anthonycr:
I'm going to wait to merge this until some turnstile work is complete

PR #196, opened 2017-03-30
ANDTV-581 Making a few modifications to support SVOD

howardrigberg:
#### Ticket
[ANDTV-581](https://vimean.atlassian.net/browse/ANDTV-581)
[ANDTV-582](https://vimean.atlassian.net/browse/ANDTV-582)
[ANDTV-583](https://vimean.atlassian.net/browse/ANDTV-583)
[ANDTV-594](https://vimean.atlassian.net/browse/ANDTV-594)

#### Ticket Summary
Adding a few modifications to support the SVOD prototype. 

#### Implementation Summary
In general, this change involves ironing out a few wrinkles around the fact that SVOD and TVOD now co-exist. There were some methods like isTrailer that needed to take that into account as naming and commenting some methods to distinguish the two types. Renaming the public API was ticketed for a future date as I didn't want to introduce any breaking changes at this point since they are not necessary to get the prototype to work. Here is the future ticket:
https://vimean.atlassian.net/browse/ADF-128

#### How to Test
Testing will happen in the related app-level prototype branch when it is ready.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -811,7 +811,7 @@ public class Video implements Serializable {
      * @return true if this is a trailer that is associated with an SVOD title, false otherwise
      */
     public boolean isSvodRelatedTrailer() {
-        return isTrailer() && getSvodInteraction() != null;
+        return isTrailer() && isSvod();
     }
 
 
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -688,7 +688,7 @@ public class Video implements Serializable {
         return VodVideoType.NONE;
     }
 
-    private static boolean isPurchased(Interaction interaction) {
+    private static boolean isInteractionPurchased(Interaction interaction) {
         return (interaction != null && interaction.mStream == Stream.PURCHASED);
     }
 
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -796,21 +796,44 @@ public boolean isSvod() {
...
+    /**
+     * Helper to determine if this is an SVOD video (not a trailer) that has not been purchased.
+     * @return true if it is an unpurchased video, false otherwise
+     */
+    public boolean isUnpurchasedSvod() {
+        return !isTrailer() &&
+               isSvod() &&
+               !isPlayable() &&
+               getSvodInteraction() != null &&
+               getSvodInteraction().getPurchaseDate() == null;
```
kevinzetterstrom: Could `getSvodInteraction() != null && getSvodInteraction().getPurchaseDate() == null` be removed in favor of `getSvodPurchaseDate() == null` since we also confirm that it is an Svod?

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -796,21 +796,44 @@ public boolean isSvod() {
...
+        return getPlayStatus() == Play.Status.PLAYABLE;
     }
 
+    /**
+     * Helper to determine if this is a trailer that is associated with an SVOD title
+     *
+     * @return true if this is a trailer that is associated with an SVOD title, false otherwise
+     */
+    public boolean isSvodRelatedTrailer() {
+        return isTrailer() && getSvodInteraction() != null;
```
kevinzetterstrom: Since the method is named `isSvodRelatedTrailer `, a suggestion here would be to write this as `isTrailer() && isSvod()`. This way if the logic behind `isSvod()` changes from `getSvodInteractins() != null`, we are a little more protected.

howardrigberg: 👍

PR #202, opened 2017-04-20
Unit tests

anthonycr:
#### Summary
Adding unit tests to verify that type adapters are or are not generated for each model.


build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.3.1'
```
kevinzetterstrom: 👌

vimeo-networking/src/main/java/com/vimeo/networking/model/Credit.java
```diff
@@ -1,3 +1,25 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
```
kevinzetterstrom: 🎉

vimeo-networking/src/test/java/com/vimeo/networking/model/appconfiguration/AppConfigurationTest.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Unit tests for {@link AppConfiguration}.
+ * <p>
+ * Created by restainoa on 4/20/17.
+ */
+public class AppConfigurationTest {
+
+    @Test
+    public void verifyTypeAdapterWasNotGenerated() throws Exception {
+        Utils.verifyNoTypeAdapterGeneration(AppConfiguration.class);
```
kevinzetterstrom: Why doesn't `AppConfiguration` use Stag? It has `SerializedName` annotated members...

This question applies to other classes in this package.

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
```diff
@@ -46,6 +46,7 @@
     private static final String STREAM_AVAILABLE = "available";
     private static final String STREAM_UNAVAILABLE = "unavailable";
 
+    @UseStag
```
kevinzetterstrom: Should we go ahead an add this to other enums while we're at it?

vimeo-networking/src/test/java/com/vimeo/networking/model/appconfiguration/AppConfigurationTest.java
anthonycr: A while ago, I went through to see what could be added and what shouldn't to stag and learned that basically all the data returned from the `/configs` endpoint is broken https://vimean.atlassian.net/browse/ADF-100. The only reason we didn't notice it until that ticket is that we aren't actually using a lot of the configuration stuff in the client app. Because the API endpoint is busted, I opted to not migrate those classes to stag. It's been low priority since we don't really care about the configs endpoint (I can up the priority if you feel it needs to change).

kevinzetterstrom: 👍

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
anthonycr: Done, it doesn't make sense to do this for all enums as many are not parsed directly during serialization, but I added a few UseStag annotations and added tests for all enums: https://github.com/vimeo/vimeo-networking-java/pull/202/commits/c9889e6cc86e2977cbc2f594d0cf1cea0f08acac

kevinzetterstrom: Awesome!

kevinzetterstrom:
Merge after the build passes

PR #204, opened 2017-04-26
[FEATURE] ADF-126 Drakov support

anthonycr:
#### Ticket
[ADF-126](https://vimean.atlassian.net/browse/ADF-126)

#### Ticket Summary
We would like to start using Drakov (https://github.com/Aconex/drakov) to support mocking API calls. We would like to support changing the host on a per request basis.

#### Implementation Summary
There were a few approaches to implementing this support. 

The first would have been to add a URL parameter to every request method in `VimeoService` and to propagate that change outward. This seemed like an extremely disruptive change to make for something that is not going to be used frequently. Additionally, adding it as a parameter on service calls would affect client code negatively, making it more complicated.

A second would be to recreate the retrofit client whenever we needed to change the base URL. This approach was not friendly for consumers that would be changing the base URL a lot and would have created some very confusing logic in the client in order to support this feature on a per request basis.

The solution chosen was to use a custom interceptor to change the base URL and set that on the OkHttpClient used by the Retrofit instance. See here for discussion on this being the correct approach for dynamically changing the base URL (https://github.com/square/retrofit/issues/1404#issuecomment-207408548, https://gist.github.com/swankjesse/8571a8207a5815cca1fb) The interceptor holds a map of endpoints that should be redirected to a different base URL, and access to this interceptor is regulated by the `VimeoClient`. This allows for a good balance of adaptability and low side affects. It has almost no impact on the client code, and allows different API calls to be changed. As an example, the code to change the base URL for all requests to the list of staff picks videos and the request to the staff picks channel itself is as follows:

```java
HttpUrl newBaseUrl = HttpUrl.parse("https://api.vimeo-staging.com/");
VimeoClient.getInstance().includePathsForBaseUrl(newBaseUrl, "/channels/927", "/channels/927/videos");
```

This method `includePathsForBaseUrl` can be called from anywhere, and after it is called, every call to the supplied path will be changed to the new host. Currently, the URI paths supplied must match exactly, which is why two calls for both the channel URI and the URI for the videos in that channel must be made. A potential solution to this problem would be allowing a more lenient matching mechanism. I wanted this to be specific as possible though, and if we decide to make it more lenient then it should be able to be easily changed.

#### How to Test
Drop the above lines of code anywhere in the client app and observe that the base URL is changed in the requests.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -102,11 +102,9 @@ public final class VimeoClient {
      */
     private VimeoAccount mVimeoAccount;
 
-    /**
-     * -----------------------------------------------------------------------------------------------------
-     * Configuration
-     * -----------------------------------------------------------------------------------------------------
-     */
+    // -----------------------------------------------------------------------------------------------------
+    // Configuration
+    // -----------------------------------------------------------------------------------------------------
     // <editor-fold desc="Configuration">
     @Nullable
     private static VimeoClient sSharedInstance;
@@ -204,6 +202,7 @@ public final class VimeoClient {
         }
     }
 
+    @NotNull
     public Retrofit getRetrofit() {
         return mRetrofit;
     }
@@ -255,11 +254,9 @@ public final class VimeoClient {
     }
     // </editor-fold>
 
-    /**
-     * -----------------------------------------------------------------------------------------------------
-     * Authentication
-     * -----------------------------------------------------------------------------------------------------
-     */
+    // -----------------------------------------------------------------------------------------------------
+    // Authentication
+    // -----------------------------------------------------------------------------------------------------
     // <editor-fold desc="Authentication">
 
     /**
@@ -694,6 +691,7 @@ public final class VimeoClient {
      */
     private static class PinCodeAccountCallback extends AccountCallback {
 
+        @NotNull
         private final Timer mTimer;
 
         public PinCodeAccountCallback(@NotNull VimeoClient client, @NotNull AuthCallback callback,
@@ -704,9 +702,7 @@ public final class VimeoClient {
 
         private void cancelPolling() {
             VimeoClient.sContinuePinCodeAuthorizationRefreshCycle = false;
-            if (mTimer != null) {
-                mTimer.cancel();
-            }
+            mTimer.cancel();
         }
 
         @Override
@@ -854,11 +850,9 @@ public final class VimeoClient {
     }
     // </editor-fold>
 
-    /**
-     * -----------------------------------------------------------------------------------------------------
-     * Editing (Video, User)
-     * -----------------------------------------------------------------------------------------------------
-     */
+    // -----------------------------------------------------------------------------------------------------
+    // Editing (Video, User)
+    // -----------------------------------------------------------------------------------------------------
     // <editor-fold desc="Editing (Video, User)">
     @Nullable
     public Call<Object> editVideo(String uri, String title, String description, @Nullable String password,
@@ -1035,11 +1029,9 @@ public final class VimeoClient {
     }
     // </editor-fold>
 
-    /**
-     * -----------------------------------------------------------------------------------------------------
-     * Pictures
-     * -----------------------------------------------------------------------------------------------------
-     */
+    // -----------------------------------------------------------------------------------------------------
+    // Pictures
+    // -----------------------------------------------------------------------------------------------------
     // <editor-fold desc="Pictures">
 
     /**
@@ -1100,11 +1092,9 @@ public final class VimeoClient {
 
     // </editor-fold>
 
-    /**
-     * -----------------------------------------------------------------------------------------------------
-     * Video actions (Like, Watch Later, Commenting)
-     * -----------------------------------------------------------------------------------------------------
-     */
+    // -----------------------------------------------------------------------------------------------------
+    // Video actions (Like, Watch Later, Commenting)
+    // -----------------------------------------------------------------------------------------------------
     // <editor-fold desc="Video actions (Like, Watch Later, Commenting)">
     public Call<Object> updateFollow(boolean follow, String uri, ModelCallback callback) {
         return updateFollow(follow, uri, callback, true);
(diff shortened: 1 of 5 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -85,28 +86,33 @@
...
     private final Cache mCache;
+    @Nullable
     private String mCurrentCodeGrantState;
 
     @NotNull
     private final Retrofit mRetrofit;
 
     @NotNull
     private final Gson mGson;
+    @Nullable
```
kvenn: Nice cleanup of adding annotations 🔥

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -85,28 +86,33 @@
...
     private VimeoAccount mVimeoAccount;
 
-    /**
-     * -----------------------------------------------------------------------------------------------------
-     * Configuration
-     * -----------------------------------------------------------------------------------------------------
-     */
+    // -----------------------------------------------------------------------------------------------------
+    // Configuration
+    // -----------------------------------------------------------------------------------------------------
```
kvenn: More nice cleanup

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -137,13 +143,28 @@ private VimeoClient(@NotNull Configuration configuration) {
...
+     * by the VimeoClient for specific paths. If a null
+     * base URL is provided, the default base URL will be
+     * used by the client.
+     *
+     * @param path    the path for which to change the base URL.
+     * @param baseUrl the new base URL to use, may be null.
+     */
+    public void setBaseUrlForRequest(@NotNull String path, @Nullable String baseUrl) {
+        mBaseUrlInterceptor.setBaseUrlForRequest(path, baseUrl);
+    }
```
kvenn: Another cool feature which I think we should add as part of this is is the ability set the baseUrl for all requests without needing to reinitialize. The Vimeo flagship app would get some benefit out of this with an example from our Admin panel. This also makes it more consistent from a consumer's standpoint since they won't have to use two separate techniques for baseUrl modification. It doesn't cover cert pinning disabling but we could maybe do that through `ignoreCertificates()`?

```java
Configuration configuration = VimeoClient.getInstance().getConfiguration();
configuration.setBaseURLString(newUrl);
configuration.setCertPinningEnabled(false); // disable cert pinning or endpoints will most likely fail
VimeoClient.initialize(configuration);
```

Another feature which I don't think needs to be done here is a blacklist of urls for the new url. So we hit every request with the new URL _except_ a select few. Might help get around that pattern matching issue a little.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -840,7 +860,7 @@ public void success(PinCodeInfo pinCodeInfo) {
                 VimeoClient.sContinuePinCodeAuthorizationRefreshCycle = true;
                 mPinCodeAuthorizationTimer.scheduleAtFixedRate(
                         new PinCodePollingTimerTask(pinCodeInfo, mPinCodeAuthorizationTimer,
-                                                    pinCodeInfo.getExpiresIn(), authCallback, sSharedInstance,
+                                                    pinCodeInfo.getExpiresIn(), authCallback, getInstance(),
```
kvenn: This change isn't functionally equivalent since `getInstance` might throw an error. I think it's fine - and probably even better - just don't know if we're making any assumptions elsewhere that will cause problems.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
kvenn: Also being able to clear this url map to go back to the default would be beneficial - I'm looking at this from an "Admin Panel" perspective with a "Reset to defaults" type option.

vimeo-networking/src/main/java/com/vimeo/networking/utils/BaseUrlInterceptor.java
```diff
@@ -0,0 +1,84 @@
...
+
+        mBaseUrlMap.put(path, baseUrl);
+    }
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Request request = chain.request();
+        String baseUrl = mBaseUrlMap.get(request.url().encodedPath());
+        if (baseUrl != null) {
+            HttpUrl httpUrl = HttpUrl.parse(baseUrl);
```
kvenn: What will happen if baseUrl is malformed?

vimeo-networking/src/main/java/com/vimeo/networking/utils/BaseUrlInterceptor.java
```diff
@@ -0,0 +1,84 @@
...
+            HttpUrl newUrl = request.url().newBuilder()
+                    .host(httpUrl.host())
+                    .scheme(httpUrl.scheme())
+                    .username(httpUrl.username())
+                    .password(httpUrl.password())
+                    .port(httpUrl.port())
+                    .build();
```
kvenn: Are these all necessary and will they potentially overwrite a default we want if they're null?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: You're right. I changed it to this because the parameter is annotated `@NotNull` and it was giving a lint warning that the instance could have been null. I assumed that there wouldn't be a consequence because it would crash when it was null. Looking into the code deeper though, it looks like the client is immediately placed in a `WeakReference`, so there was no harm passing null to the not-null annotated parameter. So, I'm going to revert back to using the field and change the annotation on the parameter.

vimeo-networking/src/main/java/com/vimeo/networking/utils/BaseUrlInterceptor.java
anthonycr: Well if the base url is malformed, the whole thing will break... the assumption is that the developer consuming this method knows how to construct correctly formed URLs. Worst case scenario it crashes, and we will never be using this in production so it doesn't seem like a problem. That said, it would be better to make the parameter an `HttpUrl` instead of a `String` to make it more clear that the url must be safe, so I'll do that.

vimeo-networking/src/main/java/com/vimeo/networking/utils/BaseUrlInterceptor.java
anthonycr: Strictly speaking for the example I gave, only `host` and `scheme` are required, however to support all potential URLs, including ones that point to localhost, I'm adding all these parts of the URI in accordance with the URI syntax: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

The only part of the URI we don't care about is the path and the query parameters

anthonycr:
Responded to PR comments: https://github.com/vimeo/vimeo-networking-java/pull/204/commits/5fee4ec59c4ce346d15914e1910578d5d05eca62 and https://github.com/vimeo/vimeo-networking-java/pull/204/commits/a2d8d26542ddf3ba819b8005a4b4568772b7e7a6

PR #205, opened 2017-04-27
ADF-128 TVOD vs SVOD

anthonycr:
#### Ticket
[ADF-128](https://vimean.atlassian.net/browse/ADF-128)

#### Ticket Summary
Because of the introduction of SVOD types into the code, it became necessary to distinguish between TVOD and SVOD rather than just naming things VOD.

#### Implementation Summary
Everything that was either TVOD or SVOD was renamed to reflect that. Unit tests were also added for several `Video` helper methods.

#### How to Test
Run the tests, see the corresponding PRs on client code.


anthonycr:
@howardrigberg since you wrote this ticket... do you think I should write more unit tests or is testing the few basic SVOD/TVOD methods I tested enough?

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -648,62 +653,62 @@ public SvodInteraction getSvodInteraction() {
...
      * 5) rental
      *
-     * @return the VodVideoType of the video or {@code VodVideoType.NONE} if it is not a VOD video or
-     * {@code VodVideoType.UNKNOWN} if it is a VOD video that is not marked as rented, subscribed or bought
+     * @return the TvodVideoType of the video or {@code TvodVideoType.NONE} if it is not a TVOD video or
+     * {@code TvodVideoType.UNKNOWN} if it is a TVOD video that is not marked as rented, subscribed or bought
      */
-    public VodVideoType getVodVideoType() {
-        if (isVod()) {
+    public TvodVideoType getTvodVideoType() {
```
kevinzetterstrom: Could be annotated NonNull

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -764,26 +769,26 @@ public Date getSvodPurchaseDate() {
...
-     * Videos that are VODs are part of Seasons. Included on a Season
+     * Videos that are TVODs are part of Seasons. Included on a Season
      * Connection is the name of that season.
      *
      * @return the Season name, or null if this Video is not part of a Season
      */
     @Nullable
-    public String getSeasonName() {
+    public String getTvodSeasonName() {
         if (mMetadata != null && mMetadata.mConnections != null &&
```
kevinzetterstrom: Not sure I agree with this name change. I think we want to leave this open enough that all videos can be in seasons, not just tvod videos.

kevinzetterstrom: On second thought, maybe I do agree with it. Since the URI will undoubtably return a list of `Season`s, which currently is TVOD related only. If we do open it up in the future we will need to make a bunch of changes anyway. 👍

anthonycr: I was having the same thoughts, but I figured until we have seasons for non vod content that it should be named accordingly. I'll keep it as is 👍

PR #206, opened 2017-05-01
ADF-154 consistent user agent

anthonycr:
#### Ticket
[ADF-154](https://vimean.atlassian.net/browse/ADF-154)

#### Ticket Summary
In order to begin publicly supporting beta versions of the networking API, the networking library needs to begin identifying itself as the source of requests using the user agent string.

See https://github.vimeows.com/Vimeo-API/api/issues/1141#issuecomment-143397

#### Implementation Summary
The format of the user agent is as follows:
```
[Optional consumer supplied UA] VimeoNetworking/version (Java)
```

The version is retrieved at runtime from a generated `BuildConfig` class.

#### How to Test
Make a request with the VimeoClient and ensure that the user agent has at least `VimeoNetworking/1.1.0 (Java)` in it.

### DIFF
--- a/vimeo-networking/build.gradle
+++ b/vimeo-networking/build.gradle
@@ -6,6 +6,9 @@ compileJava {
     targetCompatibility = 1.7
 }
 
+sourceCompatibility = '1.7'
+targetCompatibility = '1.7'
+
 buildscript {
     repositories {
         maven {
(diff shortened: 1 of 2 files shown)


vimeo-networking/build.gradle
```diff
@@ -14,10 +17,12 @@ buildscript {
     }
     dependencies {
         classpath 'net.ltgt.gradle:gradle-apt-plugin:0.6'
+        classpath 'gradle.plugin.de.fuerstenau:BuildConfigPlugin:1.1.8'
```
anthonycr: Is everyone fine with adding this plugin? it appears to be the only build config plugin for java projects https://github.com/mfuerstenau/gradle-buildconfig-plugin

kevinzetterstrom: I don't have any objections to it.

kvenn: Nor do I. Do you happen to have a method count on hand?

PR #207, opened 2017-05-01
Follow User Agent RFC Spec

anthonycr:
#### Summary
Based on the RFC spec for user agents, the primary identifier should be the first thing in the UA. The VimeoNetworking portion of the UA was being appended, which is not in accordance to the spec. This PR ensures that the implementation is up to spec.

Original PR: https://github.com/vimeo/vimeo-networking-java/pull/206
RFC Spec: https://tools.ietf.org/html/rfc7231#section-5.5.3


kvenn:
I'm waiting on a response to https://github.vimeows.com/Vimeo-API/api/issues/1141#issuecomment-143560

Merge once a decision is made there.

anthonycr:
@kvenn seems like aaron doesn't care about the which comes first so maybe we don't need this

kvenn:
@anthonycr then it's up to you and Gavin - I'd say just be consistent.

anthonycr:
Gavin and I have decided to put VimeoNetworking last, so I am closing this PR.

PR #209, opened 2017-05-08
Creating an example app in Kotlin

anthonycr:
#### Summary
The example app, but in Kotlin.


anthonycr:
@wilrnh I renamed the `example` module, and now it can't find the `secrets.xml` file so now the build is failing

wilrnh:
@anthonycr all good; would you mind updating this path to wherever the new module is?

https://github.com/vimeo/vimeo-networking-java/blob/28a1477b73d50bfb4e5a6e177e9392eccca4c98a/fastlane/Fastfile#L37

anthonycr:
@wilrnh will do, thanks. I did try to do a global search for that string throughout the project but I must have been doing something wrong because I couldn't find this line lol.

example-kotlin-android/src/main/kotlin/com/vimeo/android/networking/example/kotlin/AccountPreferenceManager.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    fun removeClientAccount() {
+        sSharedPreferences!!.edit().remove(CLIENT_ACCOUNT_JSON).apply()
+    }
+
+    // NOTE: This happens on the main thread, don't do this
+    // NOTE: This happens on the main thread, don't do this
+    var clientAccount: VimeoAccount?
+        get() {
+            val accountJSON = sSharedPreferences!!.getString(CLIENT_ACCOUNT_JSON, null) ?: return null
```
kvenn: Such elvis. Much terse.

anthonycr:
@wilrnh according to the build log now, there's a problem with accepting licenses, any ideas?

wilrnh:
@anthonycr checking; its missing Android SDK Platform 25 for some reason

.travis.yml
```diff
@@ -3,7 +3,7 @@ jdk:
 - oraclejdk8
 android:
   components:
-  - android-23
```
wilrnh: this should be `platform;android-25` @anthonycr

anthonycr: why is that, we don't do that in any other projects?

kvenn:
I'd like to take a quick look at this tomorrow

example-kotlin-android/src/main/kotlin/com/vimeo/android/networking/example/kotlin/AccountPreferenceManager.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.vimeo.networking.utils.VimeoNetworkUtil
+
+/**
+ * Class for persisting the the auth access token (and possibly the user)
+ *
+ * Created by anthonyr on 5/8/17.
+ */
+object AccountPreferenceManager {
+
+    private var sSharedPreferences: SharedPreferences? = null
```
kvenn: Since we use `!!` when accessing this field, I'd assume we'd want it to be annotated `NonNull`. Any use of `!!` would imply non-nullness in my mind.

But also with the `?`, that's saying it's nullable - but you're accessing it like it's not? Not sure what the best thing to do here is.

anthonycr: Okay, so my reasoning for using `!!` instead of `?` so that we fail if it is null is because that's how the class was designed to behave in the `java` module. The assumption is that you will call `initializeInstance` from the `onCreate` of the Application object. So in reality the class will always be initialized. I would say there is more harm than good in switching to `?` because if the preferences aren't initialized it will just fail silently and you wouldn't know why. The best case would be to change the `AccountPreferenceManager` so that the preferences are marked non null but given that this is a sample app, rearchitecting it to use proper dependency injection seems like a low priority at the moment

kvenn: Fair reasoning 👍  MWR 🥇

PR #225, opened 2017-06-28
VimeoClient Not Final

msya:
#### Ticket

In order to be able to mock `VimeoClient` during tests with Mockito, the PR makes `Vimeo` non final.


anthonycr:
Just fyi, not sure if you care, looks like your git settings are messed up because your commits aren't mapping to your github account.

PR #226, opened 2017-07-05
Adding more documentation around TVOD/SVOD methods

anthonycr:
Adding more documentation around TVOD/SVOD methods


anthonycr:
Yes this would definitely produce null warnings if everything was annotated correctly. I'll just leave it for now, since fixing null problems is a bigger task that should be undertaken as you pointed out.

PR #229, opened 2017-07-07
Add Text Tracking Request

msya:
#### Implementation
- Added endpoint  for `/videos/video_id/texttracks` in `VimeoClient`.
- Added DTOs `TextTrackList` and `TextTrack`.
- Added text tracks connection on video object.

#### Usage
```
final String uri = video.getTexttracksConnection().getUri();
VimeoClient.getInstance().getTextTrackList(uri, new ModelCallback<TextTrackList>() {
            @Override
            public void success(TextTrackList textTrackList) {
                    // show captions
            }

            @Override
            public void failure(VimeoError error) {
                       // handle error
            }
        }).
```

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
@@ -3,6 +3,8 @@ package com.vimeo.networking.model;
 import com.google.gson.annotations.SerializedName;
 import com.vimeo.stag.UseStag;
 
+import org.jetbrains.annotations.Nullable;
+
 import java.io.Serializable;
 import java.util.Date;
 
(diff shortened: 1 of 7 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
```diff
@@ -0,0 +1,116 @@
...
+@SuppressWarnings({"unused", "WeakerAccess"})
+public class TextTrack implements Serializable {
+
+    private static final long serialVersionUID = 3019422318347933808L;
+
+    @SerializedName("uri")
+    protected String mUri;
+
+    @SerializedName("active")
+    protected Boolean mActive;
```
anthonycr: I know this is a habit from using Kotlin, but maybe we should use a primitive boolean type here

vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
```diff
@@ -0,0 +1,116 @@
...
+ * Created by Mohit Sarveiya on 6/29/17.
+ */
+@UseStag
+@SuppressWarnings({"unused", "WeakerAccess"})
+public class TextTrack implements Serializable {
+
+    private static final long serialVersionUID = 3019422318347933808L;
+
+    @SerializedName("uri")
+    protected String mUri;
```
anthonycr: Stag supports private visibility on fields as long as a setter/getter is present, and since I see that you have getters and setters for all fields, this should work fine as private. Also, while you're at it, could you update stag to version 2.3.1 (it's currently at 2.2.0).

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -483,12 +490,14 @@ public Connection getRelatedConnection() {
         return null;
     }
 
-    public int likeCount() {
-        if (getLikesConnection() != null) {
-            return getLikesConnection().mTotal;
+    @Nullable
+    public Connection getTexttracksConnection() {
+        if ((mMetadata != null) && (mMetadata.mConnections != null) && (mMetadata.mConnections.mTexttracks != null)) {
+            return mMetadata.mConnections.mTexttracks;
```
anthonycr: Could you rename `mTexttracks` to `mTextTracks` as well as the method name, to match the class that we have? I'd like to see consistent naming around `Texttracks` vs `TextTracks`.

vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
```diff
@@ -0,0 +1,116 @@
...
+    protected Boolean mActive;
+
+    @SerializedName("type")
+    protected String mType;
+
+    @SerializedName("language")
+    protected String mLanguage;
+
+    @SerializedName("link")
+    protected String mLink;
```
kevinzetterstrom: Should these all be annotated Nullable?

vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
msya: I updated all the fields to private and updated stag to version 2.3.1.

vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
msya: I updated it to `boolean`.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
msya: Updated to camel case.

vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
msya: I added the Nullabe annotations.

vimeo-networking/src/main/java/com/vimeo/networking/model/TextTrack.java
```diff
@@ -0,0 +1,134 @@
...
+        this.mHlsLinkExpiresTime = hlsLinkExpiresTime;
+    }
+
+    @Nullable
+    public String getName() {
+        return mName;
+    }
+
+    public void setName(@Nullable String name) {
+        this.mName = name;
```
kevinzetterstrom: I know this PR has been merged, but our style guide mentions that we should not use the `this` keyword in methods. If you are making future edits, can you revisit this and make the change?

msya: Why shouldn't `this` be used?

kevinzetterstrom: It was an agreed upon standard that came about before you had joined. There are times when it will have to be used, such as if the class field matches the method field, but with hungarian notation, that will pretty much never happen.

PR #230, opened 2017-07-12
Added getters and setters to BaseReponseList

msya:
In order to be able to mutate a response list, getters and setters were added to it. The fields in `BaseResponseList` were made private as Stag now supports it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.1.2-3'
+    ext.kotlin_version = '1.1.3-2'
     repositories {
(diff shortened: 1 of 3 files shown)


anthonycr:
It looks like there's a bug in stag where the equality of generic parameterized type is not working as expected (javac being dumb??????) and is causing some internal check around
`ValueCallback<T>` to fail. I'll be fixing it and issuing an update soon so that we can proceed on our way with this PR. I'll ping you on the PR when I have a fix ready.

anthonycr:
Blocked by https://github.com/vimeo/stag-java/pull/121

anthonycr:
@msya you should be able to proceed with this PR by updating to https://github.com/vimeo/stag-java/releases/tag/v2.3.2

PR #232, opened 2017-07-17
Updating retrofit to version 2.3.0

anthonycr:
#### Summary
Updating retrofit to 2.3.0

- [ ] Test that cert pinning with charles proxy still works


anthonycr:
Replaces https://github.com/vimeo/vimeo-networking-java/pull/198

kvenn:
This PR gets at my question of when we want to update libraries and what our policy should be. This should probably get a 1pt ticket since it involves checking cert pinning.

PR #234, opened 2017-07-26
ADF-32 Fix generics

howardrigberg:
#### Ticket
[ADF-32](https://vimean.atlassian.net/browse/ADF-32)

#### Ticket Summary
There is a lot of usage of generics throughout the app. Often, the usage is unsafe and unchecked, which causes compiler warnings and additionally opens up classes/methods in the app for misuse and exposes us to subtle, difficult to fix bugs.

The purpose of this ticket is to rid the app of all "Unchecked warnings" related to generics/raw types misuse. This will provide us with a safer architecture, and make the app less prone to bugs.

#### Implementation Summary
Added specific typed calls for all network api calls. Added a helper class called GetRequestType to contain a series of typed callers that can be used in association with a single method called getContent in VimeoClient. This allows a similar network calling approach to what we had before with the addition of type safety. Removed GsonDeserializer.

#### How to Test
Oh jeez... all network calls are affected. So test everything. No problem, right?


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -922,10 +953,13 @@ public void failure(VimeoError error) {
...
-                                  @Nullable Privacy.PrivacyValue privacyValue,
-                                  @Nullable HashMap<String, Object> parameters,
-                                  ModelCallback<Object> callback) {
+    public Call<Video> editVideo(String uri,
+                                 String title,
+                                 String description,
+                                 @Nullable String password,
+                                 @Nullable Privacy.PrivacyValue privacyValue,
+                                 @Nullable HashMap<String, Object> parameters,
+                                 VimeoCallback<Video> callback) {
```
kevinzetterstrom: If we are throwing exceptions when things like `uri` or `callback` are null, why not just enforce `NotNull` on those parameters and remove the checks? I feel like this is a nice juncture to do so. This applies to most methods in this class. I don't think it's a ton of work to do, but I know you have already spent a lot of time on this, so I leave it up to your discretion. If you don't want to tackle it now, perhaps add a ticket so we can revisit.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1126,117 +1162,109 @@ public void failure(VimeoError error) {
...
     }
-
-    /**
-     * Delete a picture resource
-     * This is helpful if your activation or upload step fails - you can delete the resource allocated
-     *
-     * @param uri The Uri that is found in the PictureResource returned from
-     *            {@link #createPictureResource(String, ModelCallback)}
-     */
-    public void deletePictureResource(String uri, ModelCallback callback) {
```
kevinzetterstrom: I disagree with removing this. We may not use it in our app, but it could be used by other consumers - this is a valid thing to do with the API.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1126,117 +1162,109 @@ public void failure(VimeoError error) {
...
         } else {
-            return unfollow(uri, callback, enqueue);
+            return unfollow(uri, callback);
         }
     }
 
-    public Call<Object> follow(String uri, ModelCallback callback, boolean enqueue) {
-        return putContent(uri, null, callback, enqueue);
+    private Call follow(String uri, IgnoreResponseVimeoCallback callback) {
+        return putContent(uri, null, callback);
```
kevinzetterstrom: This is an interesting approach to cases in which we don't care about the response.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1376,50 +1371,18 @@ public void failure(VimeoError error) {
...
-
-        Map<String, String> queryMap = createQueryMap(query, refinementMap, fieldFilter);
-
-        Call<Object> call = mVimeoService.GET(getAuthHeader(), uri, queryMap, cacheHeaderValue);
-        try {
-            return call.execute();
-        } catch (IOException e) {
-            return null;
-        }
+    public void fetchCurrentUser(@NotNull VimeoCallback<User> callback, @Nullable String filter) {
```
kevinzetterstrom: Should this be renamed to `getCurrentUser` ?

vimeo-networking/src/main/java/com/vimeo/networking/Search.java
```diff
@@ -200,36 +199,39 @@ public static Facet fromString(String text) {
...
                                               @Nullable Map<String, String> refinementMap,
-                                              @Nullable List<Facet> facets, @Nullable String containerFilter,
+                                              @Nullable List<Facet> facets,
+                                              @Nullable String containerFilter,
                                               @Nullable String fieldFilter) {
-        if (refinementMap == null) {
-            refinementMap = new HashMap<>();
+        Map<String, String> searchRefinementMap = new HashMap<>();
+        if (refinementMap != null) {
+            searchRefinementMap = new HashMap<>(refinementMap);
```
anthonycr: Agreed, current code is needlessly allocating memory in certain scenarios.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: I would also agree that we should use `NotNull` wherever possible. It's much more explicit. We might also look into https://www.jetbrains.com/help/idea/parametersarenonnullbydefault-annotation.html if people don't want to have to specify NotNull manually.

vimeo-networking/src/main/java/com/vimeo/networking/callbacks/IgnoreResponseVimeoCallback.java
```diff
@@ -22,24 +24,18 @@
...
 /**
- * Abstract callback class for {@code T} type of model.
+ * Use this callback when you don't care about the server response and therefore don't need to deal
+ * with typed network calls.
  * <p>
- * Created by hanssena on 4/27/15.
+ * Created by rigbergh on 7/20/17.
  */
-public abstract class ModelCallback<T> extends VimeoCallback<T> {
+public abstract class IgnoreResponseVimeoCallback extends VimeoCallback<Object> {
```
anthonycr: Why should a consumer use this class over `VimeoCallback`? It seems like an unnecessary duplication which introduces complexity when consuming any api since the existence of the object in the success method shouldn't cause problems if the consumer doesn't use it.

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestType.java
```diff
@@ -0,0 +1,362 @@
...
+
+/**
+ * This is a collection of classes that should be used for get requests to
+ * {@link VimeoClient#getContent(String, CacheControl, VimeoCallback, String, Map, String, Caller)} or
+ * {@link VimeoClient#getContentSync(String, CacheControl, String, Map, String, Caller)}
+ * Created by rigbergh on 7/9/17.
+ */
+public final class GetRequestType {
+
+    public interface Caller<DataType_T> {
```
anthonycr: Not that I'm really excited about making VimeoClient any bigger, but it would make more sense to declare this interface in the class that requires it. Then this `GetRequestType` class would strictly be a sort of factory.

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestType.java
```diff
@@ -0,0 +1,362 @@
...
+                                            @NotNull String uri,
+                                            @NotNull Map<String, String> queryMap,
+                                            @NotNull String cacheHeader,
+                                            @NotNull VimeoService vimeoService) {
+                    return vimeoService.getVideoList(authHeader, uri, queryMap, cacheHeader);
+                }
+            };
+
+    private GetRequestType() {
+
```
anthonycr: should be `private GetRequestType() {}` instead... styled on one line

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
howardrigberg: It is essentially the same thing as deleteContent with another name. It serves no purpose. I am not taking any functionality away. If someone wants to delete a picture resource, they can call deleteContent.

Arguably, I should have gone farther in the opposite direction and removed more methods like follow and unfollow which are thin wrappers around methods that actually do the work. However, I decided to limit the scope since we plan to revisit the entire networking api in the near future and come up with a complete redesign. Any thoughts @kvenn, @anthonycr, @brentwatson, @msya?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
howardrigberg: Probably, but if you search the VimeoClient class, you will notice there are also other cases of "fetch" being used than this one. I could change all of these to be consistent with "get". It is easy enough to do, but I think we should wait to tackle this when we carefully consider the entirety of the networking api in a redesign. Any thoughts @kvenn, @anthonycr, @brentwatson, @msya?

vimeo-networking/src/main/java/com/vimeo/networking/callbacks/IgnoreResponseVimeoCallback.java
howardrigberg: Good question. Primarily it solves a practical problem with the put, post and delete calls in VimeoClient. In our codebase, we are ignoring the response object in all of uses of the API when onSuccess is called. In the instances where we need an object, we are referencing an in-memory version of the object that is established before the networking request is made rather than using the object returned from the server (which would be a better practice). The alternative would be to blow out typed versions of put, post and delete calls (as I did for the get requests) when our app doesn't actually need them. It reduced the scope of the problem dramatically. I don't think that this is a great long-term solution for the networking API, but it solves the problem for now. After this PR is merged, I am going to create a series of tickets regarding TBD issues that we need to address, and this is one of them. Some of this overlaps with the object tracking initiative, so we will return to this issue one way or another.

howardrigberg:
Responded:
https://github.com/vimeo/vimeo-networking-java/commit/6eb036dbb911efb5de8693190c156cb202547ec1

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -116,6 +116,16 @@
     @Nullable
     private VimeoAccount mVimeoAccount;
 
+    public interface Caller<DataType_T> {
+
+        @NotNull
+        retrofit2.Call<DataType_T> call(@NotNull String authHeader,
```
kevinzetterstrom: Does this need to have `retrofit2.` preface?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -113,6 +116,16 @@
     @Nullable
     private VimeoAccount mVimeoAccount;
 
+    public interface Caller<DataType_T> {
```
brentwatson: Could just be `Caller<T>`.  The idea that T == "Type" is pretty ubiquitous.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -80,7 +83,7 @@
  * @see <a href="https://developer.vimeo.com/api">The Vimeo API Docs</a>
  */
 @SuppressWarnings("unused")
-public class VimeoClient {
+final public class VimeoClient {
```
brentwatson: In another PR I'd suggest some refactoring here:
- Move Retrofit setup code to another object.
- If we don't care about PUT/POST/DELETE response objects we could maybe return `Call<Boolean>` instead of returning `Call` or `Call<Object>`.
- Move cache & corresponding util methods into it's own class.
- Move auth logic into it's own class.
- This object has very low cohesion (it's responsible too much).  We could either move bits of logic into other classes (eg call the PinCode logic stuff), or we could spit each logical endpoint into it's own class and have this class be a bunch of single-line methods.

General questions:
- Some objects return a `Call<T>` and some take a `VimeoCallback`.  Is there a reason we don't stick with 1 over the other?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -133,9 +150,26 @@
      */
     // <editor-fold desc="Editing">
     @PATCH
-    Call<Object> edit(@Header("Authorization") String authHeader, @Url String uri,
+    Call<Object> edit(@Header("Authorization") String authHeader,
```
brentwatson: If every single request requires an auth header we should move this an `Interceptor` do add the header there (could be done in another PR).

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
```diff
@@ -0,0 +1,350 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Map;
+
+import okhttp3.CacheControl;
+import retrofit2.Call;
+
+/**
+ * This is a collection of classes that should be used for get requests to
```
brentwatson: "used for get requests"-> "used for GET requests"

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
```diff
@@ -0,0 +1,350 @@
...
+ * {@link VimeoClient#getContent(String, CacheControl, VimeoCallback, String, Map, String, Caller)} or
+ * {@link VimeoClient#getContentSync(String, CacheControl, String, Map, String, Caller)}
+ * Created by rigbergh on 7/9/17.
+ */
+public final class GetRequestCaller {
+
+    public static final Caller<AppConfiguration> APP_CONFIGURATION =
```
brentwatson: Should all of these methods have some javadocs about the endpoint being called (and/or a link to their public docs)?

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
```diff
@@ -0,0 +1,350 @@
...
+import okhttp3.CacheControl;
+import retrofit2.Call;
+
+/**
+ * This is a collection of classes that should be used for get requests to
+ * {@link VimeoClient#getContent(String, CacheControl, VimeoCallback, String, Map, String, Caller)} or
+ * {@link VimeoClient#getContentSync(String, CacheControl, String, Map, String, Caller)}
+ * Created by rigbergh on 7/9/17.
+ */
+public final class GetRequestCaller {
```
brentwatson: Many methods are showing "unused" warnings in the IDE.  Maybe add `@SuppressWarnings("unused")` to the class.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
brentwatson: No strong preference.  The API docs do show follow/unfollow endpoints though, so might make sense to keep these in so users can more easily find what they are looking for.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
brentwatson: Consistency is good 👍 .   I'm with @kevinzetterstrom rename any `fetchXXXX` -> `getXXXX`.

howardrigberg:
@kevinzetterstrom Sorry I missed your overarching comment in my earlier pass at responding to your review. I have the readme and code example updates on my list of follow up tickets to be created. We'll make sure that they are put in the 2.0 release requirements and up the priority. Sound good?

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
howardrigberg: Yes. I will add javadocs. Good callout. However, just to be clear, my approach here was to try to apply type safety to the existing API with minimal impact to the calling approach. As it stood, most of our API get request calls were funneled through a single method where the URI is passed in rather than baking the URI into a declared retrofit endpoint. So, there is primarily a single method (VimeoClient.getContent) that is used in association with all of these Caller objects. Regardless, I will document all of them.

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
howardrigberg: They probably have unused warnings because you are not viewing the library in the context of our app. For example, they shouldn't have warnings if you view the file in the context of the mobile project. Regardless, I will be creating a follow up ticket to write unit tests for the new networking API code that I have added. In general, I think we should be striving to create unit tests for our public APIs rather than suppressing the warnings whenever possible.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
howardrigberg: Fantastic suggestion. I will add it to my follow up list of tickets to be created.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
howardrigberg: This is following a style pattern that we agreed on in one of our previous Android syncs:
https://github.vimeows.com/MobileApps/Style-Guides/blob/master/Android/README.md

Open for debate of course. Feel free to weigh in on it or add it to a future sync agenda if you strongly disagree.

anthonycr: Our style is itself a modified version of google's java style around type names (https://google.github.io/styleguide/javaguide.html#s5.2.8-type-variable-names). Note that their style also allows for use of a single letter, something that we haven't built into our style guide... yet

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
howardrigberg: Good suggestions. I will add them to my list of follow up items. Regarding your question, I doubt that there is a good reason for the difference, but I don't know for sure since I didn't write them. Regardless, the inconsistency is one of many reasons why the entire networking API needs to be refactored in a major way. It's a high priority for the near future of the library. I think that it is prudent to wait to address all of these issues at once with a complete coherent refactor.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -318,42 +311,46 @@ public Builder(String accessToken) {
...
+         * @param accountStore  (Optional, Recommended) An implementation that can be used to interface with Android's
+         *                      <a href="https://goo.gl/QZ7rm">Account Manager</a>
          */
         @Deprecated
-        public Builder(@Nullable String baseURLString, String clientId, String clientSecret, String scope,
-                       @Nullable AccountStore accountStore, @Nullable GsonDeserializer deserializer) {
+        public Builder(@Nullable String baseURLString,
+                       String clientId,
+                       String clientSecret,
+                       String scope,
```
anthonycr: While we are here the non null parameters could be annotated as such

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: My understanding is that the reason for the custom callback was to provide a hook for the old bad custom deserialization as well as error conversion and providing a simpler API for consumption (`success(T)` over `onResponse(Call<T>, Response<T>)`) and enforcing this pattern. The problem is that it does not provide a way to cancel and so really exposing the `Call<T>` is still necessary which makes for this ambiguous API.

That said, I think the above reasons mentioned are unnecessary and could be better served by just exposing the call object.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: I would say since this method provides nothing over the fall through method other than throwing a failure with an empty URI it should be deleted. I'm on board with deleting it. It would be nice if we had specific delete methods in the future though only for content which is deletable and instead of taking in a URI they would accept the object to delete and then grab the URI from the object.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: Also good moving the get 👍

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
howardrigberg: Great suggestion @anthonycr regarding future API for delete.

vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
```diff
@@ -0,0 +1,472 @@
...
+                public Call<VideoList> call(@NotNull String authHeader,
+                                            @NotNull String uri,
+                                            @NotNull Map<String, String> queryMap,
+                                            @NotNull String cacheHeader,
+                                            @NotNull VimeoService vimeoService) {
+                    return vimeoService.getVideoList(authHeader, uri, queryMap, cacheHeader);
+                }
+            };
+
+    private GetRequestCaller() {}
```
kvenn: This class seems like it would be a great candidate for an annotation processor for generating code. @anthonycr, next weekend project :) ?????

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1431,84 +1400,71 @@ public void fetchCurrentUser(@NotNull ModelCallback<User> callback, @Nullable St
...
-                                     @NotNull ModelCallback callback, @Nullable String query,
-                                     @Nullable Map<String, String> refinementMap,
-                                     @Nullable String fieldFilter) {
+    public <DataType_T> Call<DataType_T> getContent(@NotNull String uri,
+                                                              @NotNull CacheControl cacheControl,
+                                                              @NotNull VimeoCallback<DataType_T> callback,
+                                                              @Nullable String query,
+                                                              @Nullable Map<String, String> refinementMap,
+                                                              @Nullable String fieldFilter,
+                                                              @NotNull Caller<DataType_T> caller) {
```
kvenn: It appears as though the formatting is a little wonky on this file (after pulling it down).

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1431,84 +1400,71 @@ public void fetchCurrentUser(@NotNull ModelCallback<User> callback, @Nullable St
...
+     *                      {@link RequestRefinementBuilder}
+     * @param caller        The {@link Caller} for the expected response type
      * @see <a href="https://developer.vimeo.com/api/spec#common-parameters">Vimeo API Field Filter Docs</a>
      */
     @Nullable
-    public Call<Object> fetchContent(@NotNull String uri, CacheControl cacheControl,
-                                     @NotNull ModelCallback callback, @Nullable String query,
-                                     @Nullable Map<String, String> refinementMap,
-                                     @Nullable String fieldFilter) {
+    public <DataType_T> Call<DataType_T> getContent(@NotNull String uri,
```
kvenn: I can't really think of a way around this - but as it stands now, you need to declare your type in the VimeoCallback and also with the caller (on top of the fact that both are necessary as of right now).

It would be cool if you only needed one of these - and you could figure out which caller was needed based on the type that was supplied to the callback. I think we've talked about passing the class in and switching off that, but that had its problems. It just seems like potentially avoidable boilerplate since those two types will always have to be the same, and we only have one caller object per type.

Not sure if you have any other thoughts on this (pros/cons of Caller being it's own argument, alternatives you've thought of).

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -433,18 +449,19 @@ public String getCodeGrantAuthorizationURI() {
     @Nullable
     public VimeoAccount authorizeWithClientCredentialsGrantSync() {
```
howardrigberg: @WorkerThread is in android.support.annotation, so no. :(

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1282,13 +1316,16 @@ public void deletePictureResource(String uri, ModelCallback callback) {
...
+                                                  CacheControl cacheControl,
+                                                  @Nullable String fieldFilter) {
+        final String cacheHeaderValue = createCacheControlString(cacheControl);
         try {
-            return mVimeoService.getVideo(getAuthHeader(), cacheHeaderValue, uri,
-                                          createQueryMap(null, null, fieldFilter)).execute();
-        } catch (IOException e) {
+            return mVimeoService.getVideo(getAuthHeader(),
+                                          uri,
+                                          createQueryMap(null, null, fieldFilter),
```
howardrigberg: I agree that it is good practice to pull arguments out into a variable and pass them in to keep the call to one line. I will change the call appropriately.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
howardrigberg: From a bigger picture perspective, a better solution would be to remove the Caller from the networking library entirely. The need to dynamically switch out the call is an artifact of how streams is architected in our app, but it makes calling the networking library more confusing than it needs to be for someone that knows they want to call a specific endpoint and get a specific typed result. My goal in this refactor was to find a minimal-impact way to introduce type-safety so I was trying to keep a close analogy between fetchContent and getContent with the introduction of the Caller objects. I'd like to revisit this architecture in the near future with an eye to revising or removing how this works, but it was the best idea I could come up with for solving the type-safety problem with minimal scope.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1547,7 +1503,8 @@ public void fetchCurrentUser(@NotNull ModelCallback<User> callback, @Nullable St
      * @param callback The callback for the specific model type of the resource
      */
     @Nullable
-    public Call<Void> emptyResponsePost(String uri, @Nullable HashMap<String, String> postBody,
+    public Call<Void> emptyResponsePost(String uri,
+                                        @Nullable HashMap<String, String> postBody,
                                         VimeoCallback<Void> callback) {
```
howardrigberg: I added a follow up ticket to annotate all of the method parameters of VimeoClient:
https://vimean.atlassian.net/browse/ADF-199

vimeo-networking/src/main/java/com/vimeo/networking/Search.java
```diff
@@ -200,36 +199,40 @@ public static Facet fromString(String text) {
...
+    private static final String FILTER_CATEGORY = "filter_category";
+    private static final String FILTER_UPLOADED = "filter_uploaded";
+    private static final String FILTER_DURATION = "filter_duration";
+    private static final String FILTER_TYPE = "filter_type";
+    private static final String FILTER_FEATURED_COUNT = "featured_clip_count";
+    private static final String PARAMETER_GET_FACETS = "facets";
 
     public static Call<SearchResponse> search(@NotNull String query, @NotNull FilterType type,
-                                              @NotNull ModelCallback<SearchResponse> callback,
+                                              @NotNull VimeoCallback<SearchResponse> callback,
```
anthonycr: Can we move all callbacks to become the last parameter in the method to allow for in inlining the callback? You can use the refactoring tools in android studio to do this automatically. I would suggest that all methods in VimeoClient that take a callback should take the callback as their last parameter.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
howardrigberg: Ticketed for later: https://vimean.atlassian.net/browse/ADF-199

PR #238, opened 2017-08-02
ADF-39 Remove Deprecated classes/methods/fields

kevinzetterstrom:
#### Ticket
[ADF-39](https://vimean.atlassian.net/browse/ADF-39)

#### Ticket Summary
We should remove deprecated classes, methods, and fields in preparation for 2.0.0

#### Implementation Summary
This is straightforward, so I am just listing the notables:
* I kept `Embed` as Deprecated, since the API hasn't released the new Embed representation (even though we have it in this library). Once they do, it will require a version bump. I clarified in javadocs about this.
* I bumped the version to 3.3.5 based on the API changelog around the reason we added a privacy deserializer, https://vimean.atlassian.net/wiki/display/API/Changelog and then removed the custom deserializer
* I modified the Configuration.Builder so that you cannot pass in a base URL. You can change the base url only by grabbing the current Configuration first.
* I made VideoFile abstract, and subclassed it into DashVideoFile, HlsVideoFile, and ProgressiveVideoFile
* I updated the Readme to reflect changes

#### How to Test
Test via Android PR forthcoming:
* Verify fetching a video without field filters (edit the video settings) does not crash, since the VideoFileDeserializer was removed
* Test Privacy values are correctly deserialized (remove the filtering of privacy add/download)
* Test changing the base url via the admin panel, use stetho to sniff


cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Tests have failed, see below for more information.</td>
    </tr>
  </tbody>
</table>

### Tests: 

Name | Classname | Time|
--- | --- | ---|
verifyTypeAdapterWasNotGenerated | com.vimeo.networking.model.VideoFileTest | 0.003|


<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">danger</a>
</p>

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -332,24 +309,16 @@ public Builder(String clientId, String clientSecret, String scope,
          * @param accountStore  (Optional, Recommended) An implementation that can be used to interface with Android's
          *                      <a href="https://goo.gl/QZ7rm">Account Manager</a>
          */
-        @Deprecated
-        public Builder(@Nullable String baseURLString,
-                       String clientId,
-                       String clientSecret,
-                       String scope,
+        public Builder(@NotNull String clientId,
```
msya: I think the base url should still be passed into the builder. This change would allow the client to create a builder without a base url. If the base url changes, the client should create a new configuration object and set it on the `VimeoClient` object.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -512,12 +503,12 @@ public Integer playCount() {
     // -----------------------------------------------------------------------------------------------------
     // <editor-fold desc="Files">
     @Nullable
-    public VideoFile getVideoFileForMd5(String md5) {
+    public ProgressiveVideoFile getVideoFileForMd5(String md5) {
```
msya: This method is utility method. I think it should be moved to a video utils class.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kevinzetterstrom: We had marked the Builder that allowed the base URL to be set as `Deprecated` so that consumers except for us would not use it. The reasoning is that all consumers should always use `api.vimeo.com` as the base url and would never have any reason to change it. We were the only ones that should have been passing the base url in. What I found from personal experience was that I rarely ever passed anything other than `api.vimeo.com` in. And in order to pass anything else in, I had to make a change to the code anyway. So why not just change it here? 

Also what I found, was that we had some bad logic flow. Even when passing anything other than `api.vimeo.com` in, it was always overwritten after we received a response to `/configs`. This caused this to not work as intended. I felt that it was too confusing to follow so I tried to simplify that, which is why I removed the ability to set the base url when creating a configuration object.

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
```diff
@@ -24,21 +24,22 @@
...
+ * Base representation of a Video stream/playback file
  * <p>
  * Created by alfredhanssen on 4/25/15.
  */
 @SuppressWarnings("unused")
-// TODO: Remove the VideoFileDeserializer and use Stag instead, once API corrects log issue behind version 2/16/17 [KZ]
-// @UseStag(FieldOption.SERIALIZED_NAME)
-public class VideoFile implements Serializable {
+@UseStag(FieldOption.SERIALIZED_NAME)
+public abstract class VideoFile implements Serializable {
```
kevinzetterstrom: Keeping `VideoFile` not abstract and as a single class (rather than subclassing it for dash/hls/progressive) definitely worked for us in the past, especially when we received all of the items in a single array. I think specialized VideoFile classes add benefit, now that we receive specialized representations within the Play response. Dash and HLS files do not receive the information contained within the progressive response, like size, width, height, md5, mimetype, etc. Likewise, Dash contains extra stuff for DRM that none of the other receive. And whenever we access a `VideoFile` we would have to compare it to a location on the parent Play object to determine which file type it was. This was my rationale for this change.

We can easily do a `instanceof` check or use `getQuality()` to determine the file type now. Since these DTOs are created through deserialization using Stag/Gson, in order to properly have them know their quality without receiving it in JSON, I created concrete methods that are on the subclasses. 

Do you have a suggestion on how to accomplish this another way?

vimeo-networking/src/test/java/com/vimeo/networking/model/ProgressiveVideoFileTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Unit tests for {@link ProgressiveVideoFile}.
+ * <p>
+ * Created by zetterstromk on 8/3/17.
+ */
+public class ProgressiveVideoFileTest {
+
+    @Test
+    public void verifyTypeAdapterWasGenerated() throws Exception {
```
kevinzetterstrom: Not needed. And we did agree to remove them. I updated my method template to not add them, but I copy/pasted this from another test which is why it's there. I can remove it.

vimeo-networking/src/main/java/com/vimeo/networking/model/ProgressiveVideoFile.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        ProgressiveVideoFile that = (ProgressiveVideoFile) o;
+
+        return (mLink != null && that.getLink() != null) && mLink.equals(that.getLink());
+    }
+
+    @Override
+    public int hashCode() {
+        return mLink != null ? mLink.hashCode() : 0;
```
kevinzetterstrom: I would rather not add the dependency on those classes when the current hashCode/equals work just fine the way they are.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
kevinzetterstrom: Totally. Nice call out.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
```diff
@@ -85,8 +84,6 @@ public static Gson getGson() {
     public static GsonBuilder getGsonBuilder() {
         // Example date: "2015-05-21T14:24:03+00:00"
         return new GsonBuilder().registerTypeAdapterFactory(new Stag.Factory())
-                .registerTypeAdapterFactory(new PrivacyDeserializer.TypeAdapterFactory())
-                .registerTypeAdapterFactory(new VideoFileDeserializer.TypeAdapterFactory())
```
kevinzetterstrom: I didn't touch that method, and I didn't create it to begin with, so I'm not sure. @kvenn why is it done this way?

vimeo-networking/src/test/java/com/vimeo/networking/model/HlsVideoFileTest.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Unit tests for {@link HlsVideoFile}.
+ * <p>
+ * Created by zetterstromk on 8/3/17.
+ */
+public class HlsVideoFileTest {
+
+    @Test
+    public void verifyTypeAdapterWasGenerated() throws Exception {
+        Utils.verifyTypeAdapterGeneration(HlsVideoFile.class);
```
kevinzetterstrom: I have mixed feelings on this. If we do want to change the way this is done, I would prefer to do that in a separate PR, as it leads to scope creep that I don't feel strongly about tackling at the moment. Interesting with how the others feel on moving them to a centralized class. cc @howardrigberg @anthonycr @kvenn @brentwatson

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
```diff
@@ -63,18 +64,12 @@ public String toString() {
         }
     }
 
-    @Deprecated
-    @UseStag
     public enum VideoQuality {
```
kevinzetterstrom: Totally can be. We have a ticket just for that though, and since I didn't introduce any new fields, I'd rather wait for that ticket.

kevinzetterstrom: I actually think `MimeType` should be within the `ProgressiveVideoFile` file. I'd vote for `VideoQuality` to stay within this file. They are small enums and aren't used by multiple model types as represented by the API responses.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
kvenn: I can't remember for the life of me. There's synchronized access on the call objects so maybe it was an optimization around if we had a really long list of calls to cancel...?

vimeo-networking/src/test/java/com/vimeo/networking/model/HlsVideoFileTest.java
kvenn: That does feel a bit like an optimization - and in the interest of keeping this PR scoped towards just removing deprecated methods/classes (in the cleanest way possible), I'd say it could be ticketed separately and put in the vimeo-networking backlog (to be prioritized by TPMs). Maybe just quoting Mo's original suggestion in the description (so we don't lose the intent).

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
brentwatson: Sounds like the choices are the client either needs to check an enum value then somehow know which fields are available based on the enum val or option 2; do and `instanceof` check  + a cast and then they can safely know which fields to use.  I prefer option 2 (keeping changes made in this PR).

vimeo-networking/src/test/java/com/vimeo/networking/model/HlsVideoFileTest.java
brentwatson: The disadvantage of doing in 1 place is:
- Breaks package name matching for sub packages (unless we have 1 Test class for each subpackage under `model` -- but that is sort of weird.
- You lose the cmd + shift + T IDE shortcut to navigate to the corresponding test for a class.

I think either strategy is ok as the test setup here is very simple.

vimeo-networking/src/main/java/com/vimeo/networking/model/ProgressiveVideoFile.java
brentwatson: Actually, since they all just compare the `link` value the subclasses do not need a full implementations of `equals` and `hashCode`. 

`equals` can keep the `if (o == null || !(o instanceof ProgressiveVideoFile)) { return false; }` check then just do a `return super.equals(o)`.  

`hashCode()` could either be removed from these 3 subclasses or you could just `return super.hashCode()` if you want to callout the fact that there is a custom implementation in `VideoFile`.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -134,21 +134,19 @@ public String toString() {
...
-    public ArrayList<VideoFile> mVideoFiles;
-
     @SerializedName("width")
     public int mWidth;
 
     @SerializedName("height")
     public int mHeight;
 
+    /**
+     * The Embed field is in the process of moving on to the {@link #mPlay} field
```
brentwatson: Awesome docs.  Maybe add `@deprecated` annotation to the start.  Note; capital-D `@Deprecated` marks method as deprecated, and lowercase `@deprecated` shows javadocs related to the depreciation of the method..

vimeo-networking/src/test/java/com/vimeo/networking/model/ProgressiveVideoFileTest.java
```diff
@@ -0,0 +1,42 @@
...
+        Utils.verifyTypeAdapterGeneration(ProgressiveVideoFile.class);
+    }
+}
```
brentwatson: Minor: a couple of the files are missing EOF newlines.

vimeo-networking/src/main/java/com/vimeo/networking/model/ProgressiveVideoFile.java
kevinzetterstrom: nice call out. I could also remove them completely and have the base VideoFile use `getClass()` instead of `instanceof`

vimeo-networking/src/main/java/com/vimeo/networking/model/VideoFile.java
howardrigberg: I can see how the inheritance hierarchy makes sense in the way that @kevinzetterstrom refactored it, especially since the server is providing the different objects as such. However, the casting of the objects that is happening at a higher level (in VimeoPlayer) suggests that the design of the object consumption on the app level is not ideal. We should try to address that issue in the related vimeo-android ticket.

One related issue here is that the MimeType should not be in the VideoFile class. It is only relevant to the ProgressiveVideoFileClass.

kevinzetterstrom:
@howardrigberg there is already a [ticket about cleaning up the configuration builder](https://vimean.atlassian.net/browse/ADF-34), and it is pointed as a 3. Knowing we pointed it as a 3 shows that the level of effort involved in completely cleaning up the class. I would rather not go there as part of this PR, since this is much more involved and also only pointed as a 3. I've already scope creeped with the VideoFile abstraction, and I'd like to keep it to that.

vimeo-networking/src/main/java/com/vimeo/networking/model/ProgressiveVideoFile.java
kevinzetterstrom: Also, I created a [ticket to use EqualsVerifier](https://vimean.atlassian.net/browse/ADF-215) for testing our models' equals and hashCode methods. I previously vetted this library against others in a smart lock unit testing PR.

howardrigberg:
@kevinzetterstrom https://vimean.atlassian.net/browse/ADF-34 sounds like a good place to tackle my issues with the builder/configuration. I moved most of my comments to the description over there and I moved it into vn-v2.0.0 for the fix version. However, can you please fix the issues around mBaseURLString in this PR? That would involve points #2 and #4 above.  Since you altered the logic around that part of the initialization in this PR, it makes sense to go all of the way with fixing that member's access completely here.

kevinzetterstrom:
@howardrigberg I already tackled 4, and I just did 2 with 9fb9e08

kevinzetterstrom:
@msya I've responded to all comments on here. I haven't updated vimeo-kit or Vimeo-Android yet as I wanted to lock this down first. Once this PR is locked down, then I will focus on the others. (I won't merge this PR until the others are ready to merge)

vimeo-networking/src/test/java/com/vimeo/networking/model/HlsVideoFileTest.java
howardrigberg: FWIW, I see both sides of this. I guess it could get annoying to follow this pattern in networking when we will have a ton of model classes; however, Brent's point about the sub-packages makes me lean toward keeping this as-is. Otherwise, we'd probably end up with inconsistencies in our tests for models as some of them would warrant breaking out into dedicated classes and some wouldn't and it could be hard to find all of the relevant tests.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -52,6 +53,7 @@
     private static final int DEFAULT_TIMEOUT = 60; // seconds
     private static final String DEFAULT_USER_AGENT = "sample_user_agent";
 
+    @NotNull
```
howardrigberg: Can you change the name of this to match the Builder member name: mBaseUrl?

kevinzetterstrom: You got it. See b106dc7 and fbf936d

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
kvenn: I agree with all of those points. And it also looks like there is definitely some scope creep (which Kevin says below) if we were to get to all of them.

I would suggest perhaps just doing option (2) for now (since that directly effects a change made in this PR), and then adding the remainder of Howard's comments to the ticket regarding conf builder cleanup.

vimeo-networking/src/test/java/com/vimeo/networking/model/HlsVideoFileTest.java
msya: I am okay with doing this in a separate PR if we go with a different strategy I am suggesting if (OpenPOJO)[https://github.com/oshoukry/openpojo] could be used to test all models in once place similar to this strategy:

https://github.com/percolate/percolate-java-sdk/blob/master/core/src/test/java/com/percolate/sdk/dto/DtosTest.kt

msya:
@kevinzetterstrom Would you resolve the merge conflicts in the PR?

kevinzetterstrom:
@msya resolved :)

PR #242, opened 2017-08-11
Allow VimeoClient to be mocked

anthonycr:
#### Summary
We want to allow VimeoClient to be mocked.

#### Implementation Summary
Made VimeoClient non final.


anthonycr:
Great, thanks for the link!

PR #243, opened 2017-08-17
Google auth endpoints

brentwatson:
#### Summary
Added endpoints for authenticating a user using a Google `id_token` auth token.

#### Implementation Summary
The new Google auth endpoints match the existing Facebook auth endpoints almost exactly other than passing `id_token` instead of `token` param.

Note: these endpoints will only work with the vimeo app not external users of the API.

#### How to Test
Verify that a user can:
- Login with a Google auth token
- Join with a Google auth token

#### Internal Ticket
[VA-2451](https://vimean.atlassian.net/browse/VA-2451)

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -466,18 +466,12 @@ final public class VimeoClient {
     }
 
     @Nullable
-    public Call<VimeoAccount> joinWithFacebookToken(String facebookToken, String email, AuthCallback callback) {
-        if (callback == null) {
-            throw new AssertionError("Callback cannot be null");
-        }
-
-        if (facebookToken == null || facebookToken.isEmpty()) {
+    public Call<VimeoAccount> joinWithFacebookToken(@NotNull final String facebookToken, @NotNull final String email,
+                                                    @NotNull final AuthCallback callback) {
+        if (facebookToken.isEmpty()) {
             final VimeoError error = new VimeoError("Facebook authentication error.");
-
-            if (facebookToken == null || facebookToken.isEmpty()) {
-                error.addInvalidParameter(Vimeo.FIELD_TOKEN, ErrorCode.UNABLE_TO_LOGIN_NO_TOKEN,
-                                          "An empty or null Facebook access token was provided.");
-            }
+            error.addInvalidParameter(Vimeo.FIELD_TOKEN, ErrorCode.UNABLE_TO_LOGIN_NO_TOKEN,
+                                      "An empty or null Facebook access token was provided.");
             callback.failure(error);
             return null;
         }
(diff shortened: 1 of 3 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -491,6 +491,40 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
         return call;
     }
 
+    /**
+     * Register the user using Google authentication token.
```
kevinzetterstrom: May read better as "Register the user using a Google authentication token."

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -491,6 +491,40 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
...
+     *
+     * @param googleToken {@code id_token} value received by Google after authenticating.
+     * @param email User email address.
+     * @param callback This callback will be executed after the request succeeds or fails.
```
kevinzetterstrom: Formatting will likely align these params

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -491,6 +491,40 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
...
+     * @return a retrofit {@link Call} object, which <b>has already been enqueued</b>.
+     */
+    @Nullable
+    public Call<VimeoAccount> joinWithGoogleToken(String googleToken, String email, AuthCallback callback) {
```
kevinzetterstrom: Can you annotate these parameters as NonNull?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -491,6 +491,40 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
...
+        if (callback == null) {
+            throw new AssertionError("Callback cannot be null");
+        }
+
+        if (googleToken == null || googleToken.isEmpty()) {
+            final VimeoError error = new VimeoError("Google authentication error.");
+
+            if (googleToken == null || googleToken.isEmpty()) {
```
kevinzetterstrom: Why is this same `if` check nested inside another one of the same?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -586,6 +620,39 @@ public VimeoAccount logIn(String email, String password) {
         return call;
     }
 
+    /**
+     * Allow a user to login to their account using a Google authentication token.
+     *
+     * @param googleToken {@code id_token} value received by Google after authenticating.
+     * @param email User email address.
+     * @param callback This callback will be executed after the request succeeds or fails.
```
kevinzetterstrom: Formatting will likely align these params

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -586,6 +620,39 @@ public VimeoAccount logIn(String email, String password) {
...
+     * @return a retrofit {@link Call} object, which <b>has already been enqueued</b>.
```
kevinzetterstrom: I really like the callout on the enqueue both here and above

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -586,6 +620,39 @@ public VimeoAccount logIn(String email, String password) {
...
+     */
+    @Nullable
+    public Call<VimeoAccount> loginWithGoogleToken(String googleToken, String email, AuthCallback callback) {
```
kevinzetterstrom: Can you add annotations to these?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -586,6 +620,39 @@ public VimeoAccount logIn(String email, String password) {
...
+        if (callback == null) {
+            throw new AssertionError("Callback cannot be null");
+        }
+
+        if (googleToken == null || googleToken.isEmpty()) {
+            final VimeoError error = new VimeoError("Google authentication error.");
+
+            if (googleToken == null || googleToken.isEmpty()) {
```
kevinzetterstrom: Nested if check that's the same as the outer if?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
brentwatson: 👍  Updated these and the Facebook methods that are similar.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
brentwatson: Good catch.  I used the `loginWithFacebookToken`  / `joinWithFacebookToken` methods as a starting point for these and they had the same issue.  Fixed in all 4 places.

PR #245, opened 2017-08-23
Live: Api version update 3.3.10

kevinzetterstrom:
#### Ticket
[ADF-235](https://vimean.atlassian.net/browse/ADF-235)
#### Ticket Summary
We need to update our API version to 3.3.10 for Live

#### Implementation Summary
Updated the version, and handled the changes needed in https://developer.vimeo.com/api/changelog

#### How to Test
Checkout this branch, and using the Vimeo App, do a regression test. This will all be heavily tested by QA before merging into our v2.0.0 branch.

### DIFF
--- a/vimeo-networking/build.gradle
+++ b/vimeo-networking/build.gradle
@@ -17,7 +17,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'net.ltgt.gradle:gradle-apt-plugin:0.6'
+        classpath 'net.ltgt.gradle:gradle-apt-plugin:0.11'
         classpath 'gradle.plugin.de.fuerstenau:BuildConfigPlugin:1.1.8'
     }
 }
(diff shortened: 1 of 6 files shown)


vimeo-networking/build.gradle
```diff
@@ -165,3 +165,11 @@ allprojects {
 task wrapper(type: Wrapper) {
     gradleVersion = '2.3.0'
 }
+
+gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+        aptOptions.processorArgs = [
+                stagAssumeHungarianNotation: "true",
```
anthonycr: trailing comma here is unnecessary

kevinzetterstrom:
Nice call outs. I originally left `activities` alone, since the changelog only says it was removed from a User connection collection. I talked with Aaron, and he says it was been removed from all other object types a long time ago, so we are good to remove it.

In 3.3.9, folders are added. I added this connection too. I did not create a Folder object, or add it to a video, since Aaron recommended I do not, until folders are live on site. I added a video to a folder, and then requested this video, and the folder response is missing. Since there isn't any documentation on our site for folders, and since I am not getting the response, I am leaving it off for now, even though it exists in the change log.

PR #246, opened 2017-08-23
Add Live model

kevinzetterstrom:
#### Ticket
[ADF-236](https://vimean.atlassian.net/browse/ADF-236)

#### Ticket Summary
We need to add the `live` model representation, according to the API blueprint

#### Implementation Summary
I added the representation using the `Live` class, and added it to the `Video` model.


vimeo-networking/src/main/java/com/vimeo/networking/model/Live.java
```diff
@@ -0,0 +1,216 @@
...
+ * a delay between the time that the video ends, and the time that it is archived, due
+ * to transcoding.
+ * <p>
+ * Created by zetterstromk on 8/23/17.
+ */
+@SuppressWarnings("unused")
+@UseStag
+public final class Live implements Serializable {
+
+    private static final long serialVersionUID = -293744442903838563L;
```
brentwatson: Needs to be updated?  (my serialVersionUID plugin generates a different value)

vimeo-networking/src/main/java/com/vimeo/networking/model/Live.java
```diff
@@ -0,0 +1,216 @@
...
+    public enum LiveStatus {
+        ENDED,
+        READY,
+        STREAMING,
+        UNAVAILABLE,
+        UNKNOWN;
+
+        @NotNull
+        static LiveStatus getLiveStatusFromString(@NotNull String status) {
+            switch (status) {
```
brentwatson: Optional: you could use `status.toLowerCase()` to be a bit more bullet-proof.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -213,6 +213,10 @@ public String toString() {
     @SerializedName("spatial")
     public Spatial mSpatial;
 
+    @Nullable
+    @SerializedName("live")
+    private Live mLive;
```
brentwatson: Everything else is public, might as well stay consistent.  FWIW, I have a branch open that makes all fields private: https://github.com/vimeo/vimeo-networking-java/tree/make-fields-private however there are a couple issues with Stag that are preventing me from opening up a PR for this.

vimeo-networking/src/test/java/com/vimeo/networking/model/LiveTest.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Created by zetterstromk on 8/23/17.
+ */
+class LiveTest {
+
+    @Test
+    fun verifyTypeAdapterWasGenerated() {
+        Utils.verifyTypeAdapterGeneration(Live::class.java)
+    }
+}
```
brentwatson: minor: new line.

vimeo-networking/src/main/java/com/vimeo/networking/model/Live.java
kevinzetterstrom: Yes, I must have added this before the fields.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
kevinzetterstrom: I thought about that, but since it's the direction we want to go in, and it's a new field, I thought this would be better practice. How about me leaving the accessors and making the field public?

brentwatson: 👍  good idea.  Some other fields are doing the same (public + accessors).

PR #251, opened 2017-09-05
Adding the channel membership interaction

anthonycr:
#### Summary
There is an interaction on the Video object that provides contextual information about the interaction between a Video and a Channel when a video is embedded in a channel stream. It allows the consumer to delete the video from the channel by using a DELETE on the URI of the interaction.

#### Implementation Summary
I added the `channel` interaction and called it the "channel membership" interaction.

#### How to Test
No testing necessary.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
@@ -455,6 +455,17 @@ public class Video implements Serializable {
     }
     // </editor-fold>
 
+    // <editor-fold desc="Channel">
+    @Nullable
+    public Interaction getChannelMembershipInteraction() {
+        if (mMetadata != null && mMetadata.getInteractions() != null && mMetadata.getInteractions().getChannelMembership() != null) {
+            return mMetadata.getInteractions().getChannelMembership();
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -455,6 +455,17 @@ public Connection getWatchLaterConnection() {
     }
     // </editor-fold>
 
+    // <editor-fold desc="Channel">
+    @Nullable
+    public Interaction getChannelMembershipInteraction() {
+        if (mMetadata != null && mMetadata.getInteractions() != null && mMetadata.getInteractions().getChannelMembership() != null) {
+            return mMetadata.getInteractions().getChannelMembership();
+        }
```
kevinzetterstrom: Small nit 1: Github does horizontal scrolling on this line: does it exceed 120 characters?
Small nit 2: You could pull out the InteractionCollection and the Interaction into variables. FWIW, this would be so much cleaner in Kotlin!

anthonycr: Good suggestions.

PR #256, opened 2017-09-14
Make VimeoClient non-final

brentwatson:
#### Implementation Summary
`VimeoClient` in many cases should be injectable into a class so that a mock can be used instead for Unit Testing.  Having to mock networking code is a common case.  This PR makes the `VimeoClient` class non-final so that it can more easily mocked during testing.  

- Mockito does have support for mocking final classes but it's an incubating feature that we're told will change its syntax, so probably best not rely on it.
- PowerMock could also be used, but that is know to be a bit of a code-smell in many cases.  Again, not a good thing to force across a large set of tests.

Ideally `VimeoClient` may even be turned into interface vs a class, but that's a much bigger refactoring.

#### How to Test
N/A

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -75,7 +75,7 @@ import retrofit2.Retrofit;
  * @see <a href="https://developer.vimeo.com/api">The Vimeo API Docs</a>
  */
 @SuppressWarnings("unused")
-final public class VimeoClient {
+public class VimeoClient {
 
     private static volatile boolean sContinuePinCodeAuthorizationRefreshCycle;
 


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -75,7 +75,7 @@
  * @see <a href="https://developer.vimeo.com/api">The Vimeo API Docs</a>
  */
 @SuppressWarnings("unused")
-final public class VimeoClient {
+public class VimeoClient {
```
kevinzetterstrom: This is OK with me for your intended purposes!

howardrigberg: I think that we should use Mockito's incubating feature for now. It supports mocking final classes so this should be test-able without the change. @anthonycr has already given it a test drive and he said that it works as advertised in the context or our code. It might incur a little tech debt in our testing code if they change the syntax of Mockito around this feature in the future, but it shouldn't be that big of a deal to rectify.

It is easy enough to make the change here and your points about the future of this class are accurate; however I don't think we should be changing the ideal declaration of our classes for the sake of testing unless we have to.

Since we need a solution for this problem anyway, and we already have one in the form of Mockito's incubating feature, that seems like the right thing to use in this case. If it isn't, at least it seems like a good juncture to make a firm decision about what testing tool to use going forward when dealing with final classes.

brentwatson: @howardrigberg I'll push back on this, as I believe the right thing to do is here make the class non-final.
In addition to the reasons stated in the description we should also do this for the benefit of other people using this library.  Mocking or extending this class is something we should allow.  Needing to mock network requests is a very common use-case in unit tests and we are making this hard & potentially buggy not just for ourselves but also for others who include this lib.  The idea of having the class open for extension is also not a bad design (the whole "Open-closed principle" thing).

Some additional comments about this I found online: 
- https://stackoverflow.com/a/13139670
 https://www.reddit.com/r/androiddev/comments/5pz68h/is_there_a_consensus_on_making_everything_final/dcv4irn/

howardrigberg: You make some valid points. I am not going to go to the mat for this one in this context :) Part of the reason why it doesn't actually matter in this particular case is that VimeoClient isn't exactly a well thought out design in the first place and is likely to change again in the near future. Feel free to make the change.

That being said, you are bringing up a broader design question: should classes never be marked as final (in Java) or always be marked as open (in Kotlin)? I think that there are plenty of good reasons to block class inheritance. The designers of Kotlin obviously agree with this principle as it is inherent in the language design. I think it is a strong design principle to follow and we should have our testing tools accommodate the design rather than vice-versa. I added it as a discussion topic in the next Android sync and we can move from there.

brentwatson: @howardrigberg A couple things you may find interesting:
- Kotlin recently added an `all-open` compiler plugin. 
 https://kotlinlang.org/docs/reference/compiler-plugins.html
- Long convo about classes being final by default here: https://discuss.kotlinlang.org/t/classes-final-by-default/166/79
- Don't get me wrong; I like the idea of classes being `final`, but I also think that Unit Tests should be 1st class citizens of your code, and changing things to accommodate for easier/better/less-fragile tests should be encouraged.

PR #257, opened 2017-09-26
Hot fixing issue with live time data

howardrigberg:
Business live users have a monthly maximum (in seconds) that can't fit in an int. The side effect of this problem is that business live users are unable to sign into the app.


kvenn:
Had someone used my converter, this wouldn't have happened!

http://www.kylesconverter.com/time/months-to-seconds

MWR

PR #268, opened 2017-10-30
IAP endpoints

brentwatson:
#### Ticket
[VA-2723](https://vimean.atlassian.net/browse/VA-2723)

#### Ticket Summary
Added `/products`, `/products/123`, `/me/purchase/123` endpoints

#### Implementation Summary
Updated VimeoClient, VimeoService, and created/updated models to support IAP endpoints.

#### How to Test
Unit test + internal drakov server testing complete.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -1335,7 +1335,7 @@ public class VimeoClient {
      * @return a {@link Call} with the {@link Products}. This can be used for request cancellation.
      */
     @NotNull
-    public Call<Products> getProducts( @NotNull VimeoCallback<Products> callback) {
+    public Call<Products> getProducts(@NotNull VimeoCallback<Products> callback) {
         Call<Products> call = mVimeoService.getProducts(getAuthHeader());
         call.enqueue(callback);
         return call;
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -40,6 +40,7 @@ import com.vimeo.networking.model.Video;
 import com.vimeo.networking.model.VimeoAccount;
 import com.vimeo.networking.model.error.ErrorCode;
 import com.vimeo.networking.model.error.VimeoError;
+import com.vimeo.networking.model.iap.Product;
 import com.vimeo.networking.model.iap.Products;
 import com.vimeo.networking.model.notifications.SubscriptionCollection;
 import com.vimeo.networking.model.search.SearchResponse;
@@ -1339,6 +1340,17 @@ public class VimeoClient {
         Call<Products> call = mVimeoService.getProducts(getAuthHeader());
         call.enqueue(callback);
         return call;
+    }
+
+    /**
+     * Get a single {@link Product} item from a given product URI.
+     *
+     * @param uri      URI for a product. Can be obtained from {@link Product#getUri()}.
+     * @param callback Callback that will be executed once a {@link Product} has been loaded.
+     * @return {@link Call} that can be used to cancel network requests.
+     */
+    public Call<Product> getProduct(String uri, VimeoCallback<Product> callback) {
+        return getContent(uri, CacheControl.FORCE_NETWORK, GetRequestCaller.PRODUCT, null, null, null, callback);
     }
 
     /**
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -40,7 +40,6 @@ import com.vimeo.networking.model.Video;
 import com.vimeo.networking.model.VimeoAccount;
 import com.vimeo.networking.model.error.ErrorCode;
 import com.vimeo.networking.model.error.VimeoError;
-import com.vimeo.networking.model.iap.Product;
 import com.vimeo.networking.model.iap.Products;
 import com.vimeo.networking.model.notifications.SubscriptionCollection;
 import com.vimeo.networking.model.search.SearchResponse;
@@ -1099,64 +1098,6 @@ public class VimeoClient {
     // </editor-fold>
 
     // -----------------------------------------------------------------------------------------------------
-    // IAP
-    // -----------------------------------------------------------------------------------------------------
-    // <editor-fold desc="In-App Purchasing">
-
-    /**
-     * Gets a list of {@link Products} that can be purchased, such as Vimeo subscriptions.
-     *
-     * @param callback the {@link VimeoCallback} to be invoked when the request finishes
-     * @return a {@link Call} with the {@link Products}. This can be used for request cancellation.
-     */
-    @NotNull
-    public Call<Products> getProducts( @NotNull VimeoCallback<Products> callback) {
-        Call<Products> call = mVimeoService.getProducts(getAuthHeader());
-        call.enqueue(callback);
-        return call;
-    }
-
-    /**
-     * Get details for a single {@link Product} that can be purchased.
-     *
-     * @param productId the product ID
-     * @param callback the {@link VimeoCallback} to be invoked when the request finishes
-     * @return a {@link Call} with the {@link Product}. This can be used for request cancellation.
-     */
-    @NotNull
-    public Call<Product> getProduct(@NotNull String productId, @NotNull VimeoCallback<Product> callback) {
-        Call<Product> call = mVimeoService.getProduct(getAuthHeader(), productId);
-        call.enqueue(callback);
-        return call;
-    }
-
-    /**
-     * Once a purchase has been made through Google Play Billing, this endpoint can be used to inform
-     * Vimeo that the user has purchased the given item.  An updated {@link User} object will be returned if
-     * the purchase was found to be be valid and successful.
-     *
-     * @param productId the product the user purchased
-     * @param purchaseToken The purchase token provided by Google Play after a successful purchase
-     * @param orderId The order ID provided by Google Play after a successful purchase
-     * @param callback the {@link VimeoCallback} to be invoked when the request finishes
-     * @return a {@link Call} object that contains an updated User object, with the purchase applied.
-     */
-    @NotNull
-    public Call<User> purchaseSubscription(@NotNull String productId,
-                                           @NotNull String purchaseToken,
-                                           @NotNull String orderId,
-                                           @NotNull VimeoCallback<User> callback) {
-        final Map<String, String> params = new HashMap<>();
-        params.put("purchase_token", purchaseToken);
-        params.put("order_id", orderId);
-
-        Call<User> call = mVimeoService.purchaseSubscription(getAuthHeader(), productId, params);
-        call.enqueue(callback);
-        return call;
-    }
-    // </editor-fold>
-
-    // -----------------------------------------------------------------------------------------------------
     // Pictures
     // -----------------------------------------------------------------------------------------------------
     // <editor-fold desc="Pictures">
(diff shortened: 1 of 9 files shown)


howardrigberg:
I will get to this tomorrow morning.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -392,6 +395,23 @@
     // </editor-fold>
 
     // -----------------------------------------------------------------------------------------------------
+    // IAP
+    // -----------------------------------------------------------------------------------------------------
+    // <editor-fold desc="In-App Purchases">
+    @GET("products")
```
howardrigberg: Are the available products connected to the User in any way or are they generally available to anyone?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -392,6 +395,23 @@
...
+    Call<Products> getProducts(@Header("Authorization") String authHeader);
+
+    @GET("products/{product_id}")
+    Call<Product> getProduct(@Header("Authorization") String authHeader,
+                             @Path("product_id") String productId);
+
+    @PUT("/me/purchases/{product_id}")
```
howardrigberg: This endpoint probably should not be hard-coded here. Is it available in the metadata of the User object?

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
```diff
@@ -72,6 +72,28 @@ public String toString() {
         }
     }
 
+    @UseStag
+    public enum IapStatus {
```
howardrigberg: Maybe break this out into another file?

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
```diff
@@ -88,6 +110,10 @@ public String toString() {
     protected Stream mStream;
 
     @Nullable
+    @SerializedName("status")
+    protected IapStatus mStatus;
```
howardrigberg: Why protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/InteractionCollection.java
```diff
@@ -56,6 +56,10 @@
     protected Interaction mBuy;
 
     @Nullable
+    @SerializedName("purchase")
+    protected Interaction mPurchase;
```
howardrigberg: Why protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/InteractionCollection.java
```diff
@@ -134,6 +143,10 @@ public void setBuy(@Nullable Interaction buy) {
         mBuy = buy;
     }
 
+    public void setPurchase(@Nullable Interaction purchase) {
```
howardrigberg: Why is this public?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+@SuppressWarnings("unused")
+@UseStag
+public class Product implements Serializable {
+
+    private static final long serialVersionUID = 27901675080110352L;
+
+    @Nullable
+    @SerializedName("product_id")
+    protected String mProductId;
```
howardrigberg: Why protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Nullable
+    @SerializedName("name")
+    protected String mName;
```
howardrigberg: Why protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Nullable
+    @SerializedName("description")
+    protected String mDescription;
```
howardrigberg: Why protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Nullable
+    @SerializedName("uri")
+    protected String mUri;
```
howardrigberg: Why protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Nullable
+    @SerializedName("metadata")
+    protected Metadata mMetadata;
```
howardrigberg: Why protected?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public int hashCode() {
+        return mProductId != null ? mProductId.hashCode() : 0;
+    }
+
+    public String getProductId() {
+        return mProductId;
+    }
+
+    public void setProductId(String productId) {
```
howardrigberg: Why public?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+        mProductId = productId;
+    }
+
+    public String getName() {
+        return mName;
+    }
+
+    public void setName(String name) {
```
howardrigberg: Why public?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+        mName = name;
+    }
+
+    public String getDescription() {
+        return mDescription;
+    }
+
+    public void setDescription(String description) {
```
howardrigberg: Why public?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+        mDescription = description;
+    }
+
+    public String getUri() {
+        return mUri;
+    }
+
+    public void setUri(String uri) {
```
howardrigberg: Why public?

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
```diff
@@ -0,0 +1,110 @@
...
+        mUri = uri;
+    }
+
+    @Nullable
+    public Metadata getMetadata() {
+        return mMetadata;
+    }
+
+    public void setMetadata(@Nullable Metadata metadata) {
```
howardrigberg: Why public?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
brentwatson: They are specific to platform not user.

howardrigberg: Cool. I guess my bigger-picture question still applies here though... Is this an endpoint that can be configurable from somewhere in the API or is it a true top-level endpoint?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
brentwatson: Good question.  There was no mention of this in any of the docs I've reviewed yet, but that would probably make sense.  I've asked the BE dev.

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
brentwatson: My original plan was to have this a separate file, but I found that other than SearchType and ErrorCode...  all 32 other enums are in the file in which they are used.  So I kept with that for consistency.

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
brentwatson: All other fields in this model are protected.  I actually have a jam branch to change *everything* to private w/ public accessors.  Going to keep with consistency here vs a big refactor as it's outside of scope here I think.

vimeo-networking/src/main/java/com/vimeo/networking/model/InteractionCollection.java
brentwatson: Same as above -- because everything else is protected! :)

vimeo-networking/src/main/java/com/vimeo/networking/model/InteractionCollection.java
brentwatson: Because the following setters in this model are also `public`:
```java
    public void setLike(@Nullable Interaction like)
    public void setWatchLater(@Nullable Interaction watchLater)
    public void setFollow(@Nullable Interaction follow)
    public void setBuy(@Nullable Interaction buy)
    public void setRent(@Nullable Interaction rent)
    public void setSubscribe(@Nullable Interaction subscribe)
```

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
brentwatson: Since `Product` is a new object I'll update to use `private` fields. 👍.  I originally them as `private` this but other models seem to use protected fields (I think a legacy stag limitation?) so I had changed to be consistent.  I see a couple other models do have private fields though, so I guess a mix is OK for now.

vimeo-networking/src/main/java/com/vimeo/networking/model/iap/Product.java
brentwatson: Changing this and other setters to be default scope (I thought stag needed them to be public).

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -357,6 +355,9 @@
 
     @GET
     Call<TextTrackList> getTextTrackList(@Header("Authorization") String authHeader, @Url String uri);
+
```
howardrigberg: Don't you still need to declare an endpoint for getting a product object?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1327,6 +1328,18 @@ private Call unwatchLaterVideo(@Nullable String uri,
...
 
+    /**
+     * Gets a list of {@link Products} that can be purchased, such as Vimeo subscriptions.
+     *
+     * @param callback the {@link VimeoCallback} to be invoked when the request finishes
+     * @return a {@link Call} with the {@link Products}. This can be used for request cancellation.
+     */
+    @NotNull
+    public Call<Products> getProducts( @NotNull VimeoCallback<Products> callback) {
+        Call<Products> call = mVimeoService.getProducts(getAuthHeader());
```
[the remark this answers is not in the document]
brentwatson: - Initial version of this API has no filtering.  Was discussed IRL with Nicky.  Might be added for v1 if they have time.  If it's added I'll update the endpoint with another PR.
- I'm thinking we don't want clients messing with cache control for an endpoint that returns products(?).  Eg: If prices change or products are dropped/added they should always get the latest data.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
brentwatson: Not sure what we usually do here...

Above you said:
> If the new endpoints aren't top level, but are available off of the metadata of the User object, you should take the uri as a string parameter rather than hard-coding the endpoint.

Since `Product` contains `uri`field to get product details should we expect clients to use that, or provide a hardcoded endpoint?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -354,6 +355,9 @@
...
+    @GET("products")
+    Call<Products> getProducts(@Header("Authorization") String authHeader);
```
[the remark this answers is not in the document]
brentwatson: See above.  Let me know if you disagree.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
howardrigberg: Follow up happened IRL.

PR #271, opened 2017-11-13
Quota error codes and content length check

kvenn:
#### Issue
The developer message wasn't getting set properly on the VimeoError object because the content length of the response was set to 0.
Additionally, we didn't have the error codes we needed in the enum.

#### Summary
I removed the content length check because it's not necessary. The following block is wrapped in a try catch so it shouldn't cause any issues (even in the worst case).
I also added in the error codes we needed.

#### How to Test
Make a POST request to `/me/videos` with the size parameter passed in (with a value that exceeds the weekly limit). See that you get a 4101 returned.


kvenn:
That's a really good question. I got these codes from the actual file where they are being thrown. In the past we've used a spreadsheet which the API team internally maintained.

But now I don't think there is a full process for documenting the error codes. In the meantime we may just have to be very mindful of adding them with new features. And perhaps we could work with some of the other backend engineers to get a process in place/some documentation for them.

I think there may already be a ticket for it, but I'll double check.

PR #274, opened 2017-12-06
Reuse VimeoClient for Initialization

msya:
When calling `VimeoClient.initialize` with a configuration object, a new instance of `VimeoClient` is created every time. This PR uses an existing instance if there is one and sets the configuration object. If there isn't an existing VimeoClient object, then a new one is created with the configuration object.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -86,17 +86,17 @@ public class VimeoClient {
     @NotNull
     private Configuration mConfiguration;
     @NotNull
-    private final VimeoService mVimeoService;
+    private VimeoService mVimeoService;
     @Nullable
-    private final Cache mCache;
+    private Cache mCache;
     @Nullable
     private String mCurrentCodeGrantState;
 
     @NotNull
-    private final Retrofit mRetrofit;
+    private Retrofit mRetrofit;
 
     @NotNull
-    private final String mUserAgent;
+    private String mUserAgent;
 
     @Nullable
     private Timer mPinCodeAuthorizationTimer;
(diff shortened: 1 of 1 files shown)


msya:
@brentwatson I update `VimeoClient` to reinitialize retrofit, API service and `VimeoAccount` when a new `Configuration` object is given. Back to you.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -138,13 +138,23 @@ public static VimeoClient getInstance() {
 
     public static void initialize(@NotNull Configuration configuration) {
         if (sSharedInstance != null) {
-            sSharedInstance.mConfiguration = configuration;
+            sSharedInstance.mConfiguration.deleteAccount(sSharedInstance.mVimeoAccount);
+            sSharedInstance.setVimeoAccount(null);
```
brentwatson: Is this needed since `setVimeoAccount(vimeoAccount)` is done in `setup`?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -248,6 +262,10 @@ public void saveAccount(@Nullable VimeoAccount vimeoAccount, String email) {
         mConfiguration.saveAccount(vimeoAccount, email);
     }
 
+    public void setConfiguration(@NotNull Configuration configuration) {
```
brentwatson: I guess new public method should probably include javadoc.

brentwatson: Sorry, a couple more things :)...
- Should calling `setConfiguration` also call `setup`?  If not I think you're in the same situation as review #1 where various things will be out-of-sync.
-If so, then maybe we should rename this to `updateConfiguration`?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
msya: I removed it.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
msya: I removed this method I had added. In the current implementation, I think the client should create the configuration object through the builder and only provide the object through the initialize method. There should be one way to do so.

msya: See comment above.

msya:
@brentwatson I answered the questions above. Back to you.

PR #280, opened 2018-01-19
Creating typed PUT method that returns User

anthonycr:
#### Summary
Certain `PUT` calls can return a `User` in the success callback. Since our `PUT` doesn't return a proper type, it requires us to make a second request for the `User`. We can eliminate one network request by properly returning the `User` when appropriate.

#### How to Test
Make a `PUT` request to an endpoint that returns a `User` in its response.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -1593,9 +1593,11 @@ public class VimeoClient {
      * Certain endpoints will return the current user in the response when you perform a PUT on them. Use this method to
      * interact with them.
      *
-     * @param uri      the URI to connect to.
-     * @param options  the options that can be sent with the request.
-     * @param callback the callback that will be invoked.
+     * @param uri          the URI to connect to.
+     * @param cacheControl allows the consumer to force a request from cache or network.
+     * @param options      the options that can be sent with the request.
+     * @param body         The body of the PUT request
+     * @param callback     the callback that will be invoked.
      * @return a call with the request, or null if the parameters passed are invalid.
      */
     @Nullable
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -1589,6 +1589,33 @@ public class VimeoClient {
         return call;
     }
 
+    /**
+     * Certain endpoints will return the current user in the response when you perform a PUT on them. Use this method to
+     * interact with them.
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1589,6 +1589,37 @@ private boolean isSuccessfulResponse(@Nullable retrofit2.Response response) {
         return call;
     }
 
+    /**
+     * Certain endpoints will return the current user in the response when you perform a PUT on them. Use this method to
+     * interact with them.
+     *
+     * @param uri      the URI to connect to.
+     * @param options  the options that can be sent with the request.
```
brentwatson: Missing some `@param` docs.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -367,6 +367,18 @@
     Call<Products> getProducts(@Header("Authorization") String authHeader);
     // </editor-fold>
 
+    @PUT
+    Call<User> putContentWithUserResponse(@Header("Authorization") String authHeader,
```
brentwatson: Is it weird that we have a generic URI that was return a `User`?  Should we use the `Object` versions and do a cast instead?

anthonycr: If we make it return a `Call<Object>` retrofit won't deserialize the JSON response into an `User`. It will just return a `LinkedTreeMap` of the JSON content, which we will then need to manually deserialize into a `User`.

anthonycr:
@brentwatson responded to request for changes: https://github.com/vimeo/vimeo-networking-java/pull/280/commits/9e70e095456f20491437ecece98275bb389f0ad4

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
brentwatson: Approving, but maybe pass this by one of the TPMs first. Since we already have the `<Object>` versions not sure if we should use them and pay the performance cost to Gson-ize from `LinkedTreeMap`->DTO in the client.

anthonycr: @howardrigberg @msya do either of you think this method should be removed?

howardrigberg: It is a little weird as-is and clutters the API further, but until we've gotten our networking API into shape for realz I think it is a practical and simple way to solve the problem at hand.

PR #281, opened 2018-01-24
Fix logout issue when reinitializing

msya:
#### Summary
When a new configuration object was passed into the initialize method, the `deleteAccount` method was being called. It caused the app to logout when initializing VimeoClient. Therefore, the call to `deleteAccount` has been removed.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -138,7 +138,10 @@ public class VimeoClient {
 
     public static void initialize(@NotNull Configuration configuration) {
         if (sSharedInstance != null) {
-            sSharedInstance.setup(configuration);
+            if (!sSharedInstance.mConfiguration.equals(configuration)) {
+                sSharedInstance.mConfiguration.deleteAccount(sSharedInstance.mVimeoAccount);
+                sSharedInstance.setup(configuration);
+            }
         } else {
             sSharedInstance = new VimeoClient(configuration);
         }
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -138,7 +138,6 @@ public class VimeoClient {
 
     public static void initialize(@NotNull Configuration configuration) {
         if (sSharedInstance != null) {
-            sSharedInstance.mConfiguration.deleteAccount(sSharedInstance.mVimeoAccount);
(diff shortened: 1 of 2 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -138,7 +138,6 @@ public static VimeoClient getInstance() {
 
     public static void initialize(@NotNull Configuration configuration) {
         if (sSharedInstance != null) {
-            sSharedInstance.mConfiguration.deleteAccount(sSharedInstance.mVimeoAccount);
```
brentwatson: We'll still be calling `setup` with the new config object each time though, which could cause other things.
And I think we do want to delete the stored account if the config object has changed, no?

Should the logic remain but instead be wrapped in a `configuration != existing config object` check?

brentwatson: @msya After speaking IRL, we discussed maybe adding an `equals()` to the `Configuration` so you can only reset things if the config actually changes.  If this won't work let's get this merged to fix the issues we're seeing in dev.

msya:
@brentwatson I made the changes. Back to you.

brentwatson:
Tested in the debugger and it looks like this is working -- config `equals` returns true and we don't reset things.  I wouldn't mind another set of 👀 on this though @howardrigberg or @kvenn -- are there any cases where the config object would ever modify one of it's values being used in `equals` when the same user is remaining logged in?  I fear is that if that happens we've made this a bug that happens all the time to a bug that happens infrequently that is harder to recreate or even notice.

kvenn:
@brentwatson I think that's a very valid concern. Perhaps `equals` on the config object isn't what we want, but rather an `equals` check on the VimeoAccount. CacheSize could change but that wouldn't necessarily mean we should delete the account.

VimeoAccount isn't totally correct either, but it's likely closer. It depends _why_ we're deleting the account. Anything that would cause the access token to be different would likely mean we should invalidate the account, but I doubt that's actually what we'd want to happen since that results in a user being logged out, and we might be rotating a secret. I think those fields are:
```
    @NotNull
    protected String mBaseUrl;
    protected String mClientID;
    protected String mClientSecret;
    protected String mScope;

    protected String mAccessToken;

    @Nullable
    private AccountStore mAccountStore;
```

I would say a compromise would be checking the account store in the configuration.

msya:
From IRL conversation with @kvenn and @anthonycr, we decided to revert back to what we had initially before #274. This change requires further investigation into the best way to achieve this. Therefore, I'm closing this PR and will open another PR that reverts the changes to before #274.

PR #294, opened 2018-03-26
API Version 3.4 UploadQuota Representation

anthonycr:
#### Issue
The representation returned by the API has changed for the `UploadQuota` object. The representation has been updated to the following:
```
upload_quota: 
    space: // These values will reflect the lowest of lifetime or period for free and max. These fields are intended to know "Can I upload my next video"
        free: x
        max: x
        used: x
        showing: periodic|lifetime
    periodic: // These values will only reflect the current quota period. If there is no periodic quota for this user they will be present and each value will be null
        free: x
        max: x
        used: x
        reset_date: "" //ISO8601 of the time the quota will reset
    lifetime: // These values will only reflect the lifetime quota period. If there is no lifetime quota for this user they will be present and each value will be null
        free: x
        max: x
        used: x
```

#### Summary
- `Lifetime` and `Periodic` classes have been created, and the `Quota` object has been removed.
- Documentation has been added and updated to detail these representations.
- The upload quota classes have been moved to a different pacakge `.uploadquota` for clarity.


vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Lifetime.java
```diff
@@ -22,40 +22,60 @@
...
-    public long getFree() {
+    @Nullable
+    public Long getFree() {
         return mFree;
     }
 
-    public long getMax() {
+    void setFree(@Nullable Long free) {
+        mFree = free;
+    }
```
kvenn: (Optional) How do you feel about removing setters for new models? Our plan is to have immutable models in future iterations of networking, so we could start with new models.

That makes it so you'd have to find some way to clone on your own, but currently we're the largest consumer of this lib and I don't think we need to mutate this object. 

cc: @howardrigberg @msya

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Space.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @SerializedName("max")
+    private long mMax;
+
+    @SerializedName("used")
+    private long mUsed;
+
+    @NotNull
+    @SerializedName("showing")
+    private Type mShowing;
```
kvenn: (Optional) The name "showing" probably isn't the best. Are we comfortable breaking away from the API naming convention? Or should we always name our models the same thing as the API?

Pro:
It's easy to look at the API documentation and cross it with the library
Con:
We'll have to inherit their (potentially) legacy names for things - like "clip"

cc: @howardrigberg @msya

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/UploadQuota.java
```diff
@@ -0,0 +1,103 @@
...
+ * The representation of the user's upload quota. It consists of three fields that provided details about the user's
+ * quota.
+ * <p>
+ * "space": The amount of space the user has left before they will be stopped from uploading. Use this field to
+ * determine if a file the user is about to upload will breach their quota. The quota returned in this field may be
+ * either the users periodic or lifetime quota depending on their account type.
+ * <p>
+ * "periodic": The quota that resets after a certain period of time. For example, basic users are only allowed to upload
+ * ~500 MB per week, but their total storage space is much larger. If the user does not have the restriction of a
+ * periodic quota, the field will still be populated, but object will be internally empty.
```
kvenn: (Optional) Maybe: "...the field will still be non-null, ..."

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Periodic.java
```diff
@@ -22,33 +22,60 @@
...
 
 /**
- * Created by kylevenn on 8/19/15.
+ * Represents the periodic quota limits imposed on the user. If there are no periodic limits imposed on the user, then
+ * the fields will all be {@code null}.
  */
-@SuppressWarnings("unused")
 @UseStag
-public class Quota implements Serializable {
+public class Periodic implements Serializable {
```
kvenn: (Optional)(Brainstorm)

Each one of these objects has a "free", "max", and "used". Could it be useful to have an interface on these objects so that we could have a system that doesn't care if it's a periodic object, a lifetime object, or a space object? I haven't put too much thought into this, but you might want to be able to use these objects interchangeably.

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Lifetime.java
anthonycr: I have made all the setters package private for this reason. No classes outside the `com.vimeo.networking.model.uploadquota` package can access the setter. We could remove the setters all-together, but then the fields need to be made package private or protected in order for stag to generate type adapters.

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Space.java
anthonycr: I would be fine with naming this something else. As you might have guessed, I called the enum `Type` and not `Showing` because it's a very weird name for a property IMO.

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Periodic.java
anthonycr: I would be okay with creating a `Quota` or `UploadLimit` interface. The only concern I have is regarding the nullability of each field. `Periodic` and `Lifetime` both have nullable fields, but `Space` is non null.

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Lifetime.java
kvenn: I've forgotten how to java. Package private is 👌

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Space.java
kvenn: I also like the idea of calling it something else. `type` or `quotaType` perhaps.

`space.getType()` or `space.getQuotaType()` would be more readable than `space.getShowing()`. I would assume this is okay since API conventions will eventually diverge from platform conventions. As long as we have "serialized name", a consumer should always know how to find true north.

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Periodic.java
kvenn: Well that sucks. Not really much you can do about that.
Well the fields could still be non null, but the interface methods are nullable. But then that's taking a hit to the getters of Space.

If it'll clean up the code downstream, then I'd say it's worth it. If it won't, then we can hold off. No need to over architect. If the use case calls for it and it'll look prettier if these three have an interface then I'd vote a common interface with nullable fields.

anthonycr:
See https://github.vimeows.com/Vimeo-API/blueprint/blob/master/docs/3.4/representations/Clip.apib for the updated clip representation which contains the upload representation.

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Space.java
anthonycr: Going with `type`

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Periodic.java
anthonycr: I'll add it. I forgot that you can change the nullability of the implementation. So we have

```java
public interface Quota {

    @Nullable
    Long getFree();

    @Nullable
    Long getUsed();

    @Nullable
    Long getMax();

}
```
but the `Space` class implements non null versions of those, so users of `Space` don't need to handle nullability, but users of the `Quota` interface do, which is fair.

kvenn: Agreed, that's rad.

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -229,6 +230,19 @@ public String toString() {
...
+    @Nullable
+    @SerializedName("upload")
+    private Upload mUpload;
+
+    @Nullable
+    public Upload getUpload() {
+        return mUpload;
+    }
+
+    public void setUpload(@Nullable Upload upload) {
```
kvenn: Can it be package-private?

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -220,7 +221,7 @@ public String toString() {
 
     @Nullable
     @SerializedName("live")
-    public com.vimeo.networking.model.live.Live mLive;
+    public Live mLive;
```
kvenn: 👍

vimeo-networking/src/main/java/com/vimeo/networking/upload/Upload.java
```diff
@@ -0,0 +1,136 @@
...
+     * The status of the upload.
+     */
+    public enum Status {
+        @SerializedName("complete")
+        COMPLETE,
+        @SerializedName("error")
+        ERROR,
+        @SerializedName("in_progress")
+        IN_PROGRESS
+    }
```
kvenn: Does status on the Video object still hold this information as well? 🤞  If so, we won't have to do anything.

anthonycr: `status` still holds this information, but it is deprecated so I guess we can leave it for now.

anthonycr: There is an additional `transcoding` object on the clip representation which is not documented in blueprint from what I can see

kvenn:
Pointing at the most recent commit and testing on the Android app is giving me the following error:

```

04-02 20:04:12.640 23621-23621/com.vimeo.android.videoapp W/System.err: java.io.IOException: getForm() cannot be null
04-02 20:04:12.640 23621-23621/com.vimeo.android.videoapp W/System.err:     at com.vimeo.networking.upload.Upload$TypeAdapter.read(Upload$TypeAdapter.java:154)
04-02 20:04:12.640 23621-6060/com.vimeo.android.videoapp D/TURNSTILE TurnstileLogger.d (29): UPLOAD_TASK: Upload at: 60
04-02 20:04:12.641 23621-23621/com.vimeo.android.videoapp W/System.err:     at com.vimeo.networking.upload.Upload$TypeAdapter.read(Upload$TypeAdapter.java:12)
04-02 20:04:12.642 23621-23621/com.vimeo.android.videoapp W/System.err:     at com.vimeo.networking.model.Video$TypeAdapter.read(Video$TypeAdapter.java:431)
04-02 20:04:12.642 23621-23621/com.vimeo.android.videoapp W/System.err:     at com.vimeo.networking.model.Video$TypeAdapter.read(Video$TypeAdapter.java:18)
04-02 20:04:12.642 23621-23621/com.vimeo.android.videoapp W/System.err:     at retrofit2.converter.gson.GsonResponseBodyConverter.convert(GsonResponseBodyConverter.java:37)
04-02 20:04:12.642 23621-23621/com.vimeo.android.videoapp W/System.err:     at retrofit2.converter.gson.GsonResponseBodyConverter.convert(GsonResponseBodyConverter.java:25)
04-02 20:04:12.642 23621-23621/com.vimeo.android.videoapp W/System.err:     at retrofit2.ServiceMethod.toResponse(ServiceMethod.java:117)
04-02 20:04:12.642 23621-23621/com.vimeo.android.videoapp W/System.err:     at retrofit2.OkHttpCall.parseResponse(OkHttpCall.java:211)
04-02 20:04:12.643 23621-23621/com.vimeo.android.videoapp W/System.err:     at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:106)
04-02 20:04:12.643 23621-23621/com.vimeo.android.videoapp W/System.err:     at okhttp3.RealCall$AsyncCall.execute(RealCall.java:135)
04-02 20:04:12.643 23621-23621/com.vimeo.android.videoapp W/System.err:     at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
04-02 20:04:12.643 23621-23621/com.vimeo.android.videoapp W/System.err:     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
04-02 20:04:12.643 23621-23621/com.vimeo.android.videoapp W/System.err:     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
04-02 20:04:12.643 23621-23621/com.vimeo.android.videoapp W/System.err:     at java.lang.Thread.run(Thread.java:761)
04-02 20:04:12.647 23621-23621/com.vimeo.android.videoapp V/ANALYTICS Analytics.event (161): Event: VideoUpload_App with Attributes:
                                                                                             {"network":"cellular","Action":"Failure","file path":"\/storage\/emulated\/0\/Download\/An all-new Vimeo app for Android-HD.mp4","error message":"getForm() cannot be null","video uri":"\/videos\/262873070","file size":"25MB+ to 50MB"}
```

Relevant part:
```
java.io.IOException: getForm() cannot be null
at com.vimeo.networking.upload.Upload$TypeAdapter.read(Upload$TypeAdapter.java:154)
```

It's marked as NotNull. Not sure if there's a field filtering issue (making it a client side PR thing), or if the API sometimes doesn't actually respond with form. I'm checking postman now.

kvenn:
I'm seeing it in postman, probably field filters then?

As a side note, a GET request failing is probably fine to throw an error for parsing. But a POST throwing a parsing error might not be great. This is a good example where the uploads actually go live, but we though the post was a failure. Maybe we have a separate branch for parsing errors....? Or treat it on a case by case basis? Or shut the download down if we see a failure, which we're not currently doing.

This is the case where it succeeds, but the client thinks it fails.

anthonycr:
@kvenn ready for re-review

vimeo-networking/src/main/java/com/vimeo/networking/model/uploadquota/Space.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @NotNull
+    @SerializedName("max")
+    private Long mMax;
+
+    @NotNull
+    @SerializedName("used")
+    private Long mUsed;
+
+    @Nullable
```
anthonycr: Maybe it would be better to assign a default rather than make this nullable? What do you think, @kvenn ?

kvenn: Hmm that could break the UI though or put it into an invalid state. I think it'd be better to keep it nullable for now and keep our eye on the prize of having correctly annotated API blueprints since I'm assuming this is actually nonnull (but this will also have to come with the addition of versioning since that would also give us problems).

PR #299, opened 2018-05-15
Video settings Comment/Download privacy options

msya:
Added privacy options for commenting and downloading a video.


vimeo-networking/src/main/java/com/vimeo/networking/utils/PrivacySettingsParams.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Builder for setting privacy params for a video.
+ *
+ * Created by Mohit Sarveiya on 3/20/18.
+ */
+public final class PrivacySettingsParams {
```
anthonycr: Considering that this class creates a map that is not type safe, it would be good to have some unit tests that verify that the map contains the expected values.

vimeo-networking/src/main/java/com/vimeo/networking/model/Privacy.java
```diff
@@ -42,35 +42,110 @@
...
     private static final String PRIVACY_TVOD = "ptv";
     private static final String PRIVACY_CONTACTS = "contacts";
     private static final String PRIVACY_PASSWORD = "password";
     private static final String PRIVACY_DISABLE = "disable";
     private static final String PRIVACY_UNLISTED = "unlisted";
+    private static final String PRIVACY_WHITE_LIST = "white_list";
 
     @UseStag
-    public enum PrivacyValue {
+    public enum PrivacyCommentValue {
```
anthonycr: I can't remember if this has been discussed before, but do you think it would be a good idea to remove the `Privacy` prefix from all the names of the enums in this class? Referencing each of these enums seems a little redundant with `Privacy.PrivacyCommentValue`, when `Privacy.CommentValue` seems just as clear. The only downside is that you would always want to use the qualified name for the context.

vimeo-networking/src/main/java/com/vimeo/networking/model/Privacy.java
```diff
@@ -42,35 +42,110 @@
...
+
+        @NotNull
+        private final String text;
+
+        PrivacyCommentValue(@NotNull String text) {
+            this.text = text;
+        }
+
+        @Nullable
+        public static PrivacyCommentValue privacyValueFromString(@Nullable final String string) {
```
anthonycr: could/should there be a unit test for this?

vimeo-networking/src/main/java/com/vimeo/networking/model/Privacy.java
```diff
@@ -42,35 +42,110 @@
...
+
+        @NotNull
+        private final String text;
+
+        PrivacyEmbedValue(@NotNull String text) {
+            this.text = text;
+        }
+
+        @Nullable
+        public static PrivacyEmbedValue privacyValueFromString(@Nullable final String string) {
```
anthonycr: could/should there be a unit test for this?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -940,20 +940,19 @@ public void failure(VimeoError error) {
...
-        if (privacyValue != null) {
-            if (privacyValue == Privacy.PrivacyValue.PASSWORD) {
-                if (password == null) {
-                    callback.failure(new VimeoError("Password cannot be null password privacy type"));
+        if (privacySettingsParams != null && !privacySettingsParams.getParams().isEmpty()) {
+            final Map<String, Object> privacyMap = privacySettingsParams.getParams();
+            final PrivacyViewValue viewPrivacyValue = (PrivacyViewValue) privacyMap.get(Vimeo.PARAMETER_VIDEO_VIEW);
 
-                    return null;
+            if (viewPrivacyValue != null) {
```
anthonycr: this null check is unnecessary, since the condition immediately following implicitly performs the same check

vimeo-networking/src/main/java/com/vimeo/networking/model/PrivacyType.java
```diff
@@ -0,0 +1,13 @@
+package com.vimeo.networking.model;
+
+/**
+ * The different types of capabilities for a video whose privacy values can be changed.
+ *
+ * Created by Mohit Sarveiya on 3/15/18.
+ */
+public enum PrivacyType {
```
anthonycr: it looks like this enum is not used anywhere else, can it be removed?

vimeo-networking/src/test/java/com/vimeo/networking/model/PrivacyTest.java
```diff
@@ -17,7 +17,18 @@ public void verifyTypeAdapterWasNotGenerated() throws Exception {
...
-        Utils.verifyTypeAdapterGeneration(Privacy.PrivacyValue.class);
+    public void verifyTypeAdapterWasGenerated_PrivacyCommentValue() throws Exception {
+        Utils.verifyTypeAdapterGeneration(Privacy.PrivacyCommentValue.class);
     }
-}
\ No newline at end of file
+
+    @Test
+    public void verifyTypeAdapterWasGenerated_PrivacyEmbedValue() throws Exception {
+        Utils.verifyTypeAdapterGeneration(Privacy.PrivacyEmbedValue.class);
```
anthonycr: 🔥

vimeo-networking/src/test/java/com/vimeo/networking/model/PrivacyTest.java
```diff
@@ -17,7 +17,18 @@ public void verifyTypeAdapterWasNotGenerated() throws Exception {
...
+    }
+
+    @Test
+    public void verifyTypeAdapterWasGenerated_PrivacyViewValue() throws Exception {
+        Utils.verifyTypeAdapterGeneration(Privacy.PrivacyViewValue.class);
```
anthonycr: 🔥

vimeo-networking/src/main/java/com/vimeo/networking/utils/PrivacySettingsParams.java
```diff
@@ -0,0 +1,91 @@
...
+        params.put(Vimeo.PARAMETER_VIDEO_VIEW, privacyViewType);
+        return this;
+    }
+
+    /**
+     * @return a map of all the privacy settings set.
+     */
+    @NotNull
+    public Map<String, Object> getParams() {
+        return new HashMap<>(params);
```
anthonycr: prevents mutation by consumers 👍🏻

vimeo-networking/src/main/java/com/vimeo/networking/model/Video.java
```diff
@@ -471,6 +471,17 @@ public Connection getWatchLaterConnection() {
...
     }
+
+    @Nullable
+    public Connection getUsersWithAccessConnection() {
+        if (mMetadata != null
+            && mMetadata.mConnections != null
+            && mMetadata.mConnections.getUsersWithAccess() != null) {
+            return mMetadata.mConnections.getUsersWithAccess();
+        }
+        return null;
```
anthonycr: This could be simplified with one less condition, as there is no need to null check `getUsersWithAccess` since the method is annotated `@Nullable`.
```java
if (mMetadata != null && mMetadata.mConnections != null) {
     return mMetadata.mConnections.getUsersWithAccess();
}
return null;
```

msya:
@anthonycr Thanks for the review. I made the code review changes. Back to you.

vimeo-networking/src/main/java/com/vimeo/networking/utils/PrivacySettingsParams.java
```diff
@@ -0,0 +1,90 @@
+package com.vimeo.networking.utils;
+
+import com.vimeo.networking.Vimeo;
+import com.vimeo.networking.model.Privacy.CommentValue;
+import com.vimeo.networking.model.Privacy.EmbedValue;
+import com.vimeo.networking.model.Privacy.ViewValue;
```
anthonycr: Import could just import `Privacy` for clarity, to force consumers to always use the qualified class name.

vimeo-networking/src/test/java/com/vimeo/networking/utils/PrivacySettingsParamsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.vimeo.networking.model.Privacy
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+/**
+ * Tests for {@link PrivacySettingsParams}.
+ *
+ * Created by Mohit Sarveiya on 10/30/15.
+ */
+class PrivacySettingsParamsTest {
```
anthonycr: nice tests

vimeo-networking/src/test/java/com/vimeo/networking/utils/PrivacySettingsParamsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+    @Test
+    fun `Check privacy settings are set`() {
+        privacySettingsParams
+                .comments(Privacy.CommentValue.ANYBODY)
+                .embed(Privacy.EmbedValue.PRIVATE)
+                .view(Privacy.ViewValue.ANYBODY)
+                .addToCollections(true)
+                .download(true)
+
+        privacySettingsParams.params.let {
```
anthonycr: I think this is a good use case for `with`

vimeo-networking/src/test/java/com/vimeo/networking/utils/PrivacySettingsParamsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Test
+
+/**
+ * Tests for {@link PrivacySettingsParams}.
+ *
+ * Created by Mohit Sarveiya on 10/30/15.
+ */
+class PrivacySettingsParamsTest {
+
+    private val privacySettingsParams = PrivacySettingsParams()
```
anthonycr: this property could just be a local variable in each test since there does not need to be shared state.

PR #305, opened 2018-06-15
Api category localization

harish-veeramani:
#### Issue
When requests are made, particularly with categories, we are supposed to be able to get a localized version back. On categories, this means a few titles will be in a different language. We need to tell the API that we want a different localization for a request.

#### Summary
Created a LanguageHeaderInterceptor to automatically add the `Accept-Language` header to HTTP requests made to the API. Done by passing in the user's locales to the Configuration Builder and then passing on the language codes to the interceptor. On devices that support multiple locales, a string of all the locales will be passed on and the API will select the first support locale. Defaults to English if none of the locales are supported.

#### How to Test
- Go to Explore tab and check the current language of the category titles.
- Close Vimeo app and change device language
- Open Vimeo app and go back to Explore tab and verify category titles have changed to new language.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -426,7 +426,6 @@ public class Configuration {
             return this;
         }
 
-
         /**
          * Takes a single locale and creates a single element list to add to Builder
          *
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -61,7 +61,7 @@ public class Configuration {
     protected String mScope;
 
     @NotNull
-    protected List<Locale> mLocales;
+    protected final List<Locale> mLocales;
 
     protected String mAccessToken;
 
@@ -341,8 +341,10 @@ public class Configuration {
         }
 
         /**
-         * If you used the basic Builder access token constructor but have the intent of
+         * This method is deprecated. Please use {@code Builder(String, String, String, AccountStore)}
+         * and provide the client id and secret there instead.
          */
+        @Deprecated
         public Builder setClientIdAndSecret(String clientId, String clientSecret) {
             this.mClientID = clientId;
             this.mClientSecret = clientSecret;
@@ -415,11 +417,11 @@ public class Configuration {
         }
 
         /**
-         * Sets a list of locales in the Builder
+         * Sets a list of locales in the Builder. Default language is English if nothing is set.
+         * Will replace any existing locales set in the Builder.
          *
-         * @param locales      The list of locales from the user
-         *
-         * @return             An instance of the current Builder
+         * @param locales The list of locales from the user
+         * @return An instance of the current Builder
          */
         public Builder setLocales(@NotNull List<Locale> locales) {
             mLocales = locales;
@@ -427,11 +429,11 @@ public class Configuration {
         }
 
         /**
-         * Takes a single locale and creates a single element list to add to Builder
+         * Takes a single locale and creates a single element list to add to Builder. Default
+         * language is English if nothing is set. Will replace any existing locales set in the Builder.
          *
-         * @param locale       A single locale
-         *
-         * @return             An instance of the current Builder
+         * @param locale A single locale
+         * @return An instance of the current Builder
          */
         public Builder setLocale(@NotNull Locale locale) {
             mLocales = new ArrayList<>();
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -32,6 +32,7 @@ import org.jetbrains.annotations.Nullable;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 import okhttp3.Cache;
 import okhttp3.Interceptor;
@@ -58,6 +59,9 @@ public class Configuration {
     protected String mClientID;
     protected String mClientSecret;
     protected String mScope;
+
+    @NotNull
+    protected List<Locale> mLocales;
 
     protected String mAccessToken;
 
@@ -98,6 +102,10 @@ public class Configuration {
 
     public String getScope() {
         return mScope;
+    }
+
+    public List<Locale> getLocales() {
+        return mLocales;
     }
 
     public String getAccessToken() {
@@ -215,6 +223,7 @@ public class Configuration {
         this.mClientID = builder.mClientID;
         this.mClientSecret = builder.mClientSecret;
         this.mScope = builder.mScope;
+        this.mLocales = builder.mLocales;
 
         this.mAccessToken = builder.mAccessToken;
 
@@ -259,6 +268,11 @@ public class Configuration {
         private String mClientID;
         private String mClientSecret;
         private String mScope;
+
+        @NotNull
+        private List<Locale> mLocales = new ArrayList<Locale>(){{
+            add(Locale.ENGLISH);
+        }};
 
         private String mAccessToken;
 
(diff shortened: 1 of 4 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -395,6 +411,17 @@ public Builder setLogLevel(LogLevel level) {
             return this;
         }
 
+        public Builder setLocales(List<Locale> locales) {
+            this.mLocales = locales;
```
anthonycr: `this.` is not needed due to use of hungarian notation.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -395,6 +411,17 @@ public Builder setLogLevel(LogLevel level) {
...
+            return this;
+        }
+
+        public Builder setLocale(Locale locale) {
+            this.mLocales = new ArrayList<>();
+            this.mLocales.add(locale);
```
anthonycr: `this.` is not needed due to use of hungarian notation.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
harish-veeramani: I added the `this.` since the rest of the methods also use `this.`

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
harish-veeramani: I added the `this.` since the rest of the methods also use `this.`

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -395,6 +411,17 @@ public Builder setLogLevel(LogLevel level) {
             return this;
         }
 
+        public Builder setLocales(List<Locale> locales) {
```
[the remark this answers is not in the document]
harish-veeramani: Good catch. To deal with nulls I have added 
```java
    private boolean isLocaleValid() {
        return !this.mLocales.isEmpty() && !this.mLocales.contains(null);
    }
```
in the `Configuration` class to throw an error if the Locales are invalid.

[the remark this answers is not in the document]
harish-veeramani: All of the other public methods in the class do not contain and docs. Should I still add docs?

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
anthonycr: no need to change all of them

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -143,7 +148,9 @@ public static void initialize(@NotNull Configuration configuration) {
 
     private VimeoClient(@NotNull Configuration configuration) {
         mConfiguration = configuration;
+        this.mLanguageHeaderInterceptor = new LanguageHeaderInterceptor(getLocaleCodes(mConfiguration));
```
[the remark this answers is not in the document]
anthonycr: I don't think that modifying the Locale's after initialization is something that we would want to support. I think we should move away from mutating the state of the `VimeoClient` after its creation. If the consumer wants to change the language, they can call `initialize` again.

vimeo-networking/src/main/java/com/vimeo/networking/interceptors/LanguageHeaderInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+/**
+ * Add custom {@code Accept-Language} header to all requests.
+ * Created by harishveeramani on 6/13/18.
+ */
+public class LanguageHeaderInterceptor implements Interceptor {
+    private static final String HEADER_SEPARATOR = ",";
+    private String mValidLocales;
```
anthonycr: minor: add an `@NotNull` annotation to this field to specify that it is not null. See info here: https://www.jetbrains.com/help/idea/2017.3/nullable-and-notnull-annotations.html

vimeo-networking/src/main/java/com/vimeo/networking/interceptors/LanguageHeaderInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+    public Response intercept(Interceptor.Chain chain) throws IOException {
+        return chain.proceed(
+                chain.request().newBuilder().header(
+                        Vimeo.HEADER_ACCEPT_LANGUAGE,
+                        mValidLocales
+                ).build()
+        );
+    }
+
+    private String parseLocales(List<Locale> locales) {
```
anthonycr: Add `@NotNull` annotation to the return type and to the `locales` parameter.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -395,6 +414,31 @@ public Builder setLogLevel(LogLevel level) {
...
         }
 
+        /**
+         * Sets a list of locales in the Builder
+         *
+         * @param locales      The list of locales from the user
+         *
+         * @return             An instance of the current Builder
+         */
+        public Builder setLocales(List<Locale> locales) {
```
anthonycr: add `@NotNull` annotation to `locales` parameter

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -395,6 +414,31 @@ public Builder setLogLevel(LogLevel level) {
...
+        }
+
+        /**
+         * Takes a single locale and creates a single element list to add to Builder
+         *
+         * @param locale       A single locale
+         *
+         * @return             An instance of the current Builder
+         */
+        public Builder setLocale(Locale locale) {
```
anthonycr: add `@NotNull` annotation to `locale` parameter.

harish-veeramani:
@anthonycr Added the annotations

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1402,7 +1409,8 @@ public void getCurrentUser(@NotNull VimeoCallback<User> callback) {
      * @param callback the callback to be invoked when the request finishes
      */
     public void getCurrentUser(@Nullable String filter, @NotNull VimeoCallback<User> callback) {
-        getContent(Vimeo.ENDPOINT_ME, CacheControl.FORCE_NETWORK, GetRequestCaller.USER, null, null, filter, callback);
+        getContent(Vimeo.ENDPOINT_ME, CacheControl.FORCE_NETWORK, GetRequestCaller.USER, null,
```
howardrigberg: A few things here:
1. Please format this file
2. Our current preferred style is to put each parameter on a new line if the method call will exceed the allowed line length
3. Please add @SuppressWarnings("WeakerAccess") to this method to silence the lint warning.

vimeo-networking/src/main/java/com/vimeo/networking/interceptors/LanguageHeaderInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+        mValidLocales = parseLocales(locales);
+    }
+
+    @Override
+    public Response intercept(Interceptor.Chain chain) throws IOException {
+        return chain.proceed(
+                chain.request().newBuilder().header(
+                        Vimeo.HEADER_ACCEPT_LANGUAGE,
+                        mValidLocales
+                ).build()
```
howardrigberg: Please run the formatter on this file.

vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
```diff
@@ -260,6 +272,11 @@ private boolean isValid() {
         private String mClientSecret;
         private String mScope;
 
+        @NotNull
+        private List<Locale> mLocales = new ArrayList<Locale>() {{
```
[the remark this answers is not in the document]
harish-veeramani: I'm not seeing any lint warnings related to those lines

vimeo-networking/src/main/java/com/vimeo/networking/interceptors/LanguageHeaderInterceptor.java
harish-veeramani: No lines were changed when formatting

PR #306, opened 2018-06-18
Updated Kotlin and Android Gradle Plugin

msya:
- Updated Android Gradle plugin to 3.1.3
- Updated Gradle to 4.4
- Updated Gradle Apt Plugin to 0.16


build.gradle
```diff
@@ -2,13 +2,13 @@
 
 buildscript {
     ext.supportLibraryVersion = '26.1.0'
-    ext.kotlin_version = '1.2.21'
+    ext.kotlin_version = '1.1.51'
```
anthonycr: Is downgrading the version unintentional?

build.gradle
```diff
@@ -2,13 +2,13 @@
...
     repositories {
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.3'
```
anthonycr: The latest build tools plugin will force our build tools to the minimum required version (`27.0.3`) even though we specify the version in our gradle files to be `26.0.2`. This means that the build failed because we did not accept the license for `27.0.3`. We can do that by updating the versions specified in the gradle files of the android and kotlin examples, and then updating L7 of the `.travis.yml` file to point to `27.0.3`.

build.gradle
```diff
@@ -2,14 +2,14 @@
...
-    ext.kotlin_version = '1.2.21'
+    ext.kotlin_version = '1.2.71'
     repositories {
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
-        classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7.3"
+        classpath 'com.android.tools.build:gradle:3.2.0'
```
anthonycr: build is failing because `3.2.0` requires build tools 28, but we are only including build tools 26 in the travis configs

msya:
@anthonycr The CI build succeeded. Back to you.

PR #307, opened 2018-06-19
Exposing the raw error code

anthonycr:
# Summary
We need to expose raw error codes as a part of `VimeoError` and `InvalidParameter` for debug reasons. We want to maintain the existing enum for use as well. This PR adds a property to each class that exposes the raw error code as a `String`. In order to maintain backwards compatibility, Gson is being used to support virtual properties of the `ErrorCode` type. These properties are no longer fields, but are computed at runtime based on the raw error code property, which is backed by a field.

Additional changes that were made:
- Wrote unit tests to verify (de)serialization of the `ErrorCode` type in both `VimeoError` and `InvalidParameter`.
- Removed `UNABLE_TO_LOGIN_PINCODE_EXPIRED` and `INVALID_INPUT_EMPTY_USER_URI` enum values of `ErrorCode` as they were duplicates. While this is technically a breaking change, it was a bug to have these in the code
- Converted existing tests to kotlin.

## How to Test
- Run the unit tests


vimeo-networking/src/main/java/com/vimeo/networking/model/error/InvalidParameter.java
```diff
@@ -54,16 +57,27 @@ public InvalidParameter() {}
...
-        return mField;
+    private void setErrorCode(@NotNull ErrorCode errorCode) {
+        final String json = VimeoNetworkUtil.getGson().toJson(errorCode);
+        mRawErrorCode = json.replaceAll("^\"|\"$", "");
     }
 
+    @NotNull
     public ErrorCode getErrorCode() {
-        return mErrorCode == null ? ErrorCode.DEFAULT : mErrorCode;
+        final ErrorCode errorCode = VimeoNetworkUtil.getGson().fromJson(mRawErrorCode, ErrorCode.class);
```
howardrigberg: There is obviously a performance cost related to this that wasn't there before. Is there a reason why you can't just save the error code as a member when this class is initialized with it in the constructor or with setErrorCode with and return it here?

anthonycr: It can be done, I will add in setters `setRawErrorCode` that deserialize the value and initialize the error code to reduce the performance code.

vimeo-networking/src/main/java/com/vimeo/networking/model/error/ErrorCode.java
```diff
@@ -175,8 +175,6 @@
     UNABLE_TO_LOGIN_TOKEN_CAN_NOT_DECRYPT,
     @SerializedName("2314")
     UNABLE_TO_LOGIN_TOKEN_TOO_LONG,
-    @SerializedName("2315")
-    UNABLE_TO_LOGIN_PINCODE_EXPIRED,
```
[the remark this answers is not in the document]
anthonycr: `INVALID_TOKEN` is the error that is sent from the API. Before this change, an expired pincode could technically result in two different `ErrorCode`s, `UNABLE_TO_LOGIN_PINCODE_EXPIRED` if we detect the invalidation at the client level, or `INVALID_TOKEN` if the invalidation is not detected by us, so this change brings the enum in line with the API's behavior. The numerical duplication means non determinism when it comes to serialization and deserialization, as when `UNABLE_TO_LOGIN_PINCODE_EXPIRED` was serialized to JSON and then deserialized, it would be deserialized as `UNABLE_TO_CREATE_USER_CAN_NOT_VALIDATE_TOKEN` instead.

Since this enum is meant to delineate error codes that can be returned by the API, putting in codes that can locally originate from our library means that the `ErrorCode` enum violates the single responsibility principle.

anthonycr:
@howardrigberg back to you

vimeo-networking/src/main/java/com/vimeo/networking/model/error/VimeoError.java
```diff
@@ -60,7 +61,13 @@
...
 
     @SerializedName("error_code")
-    protected ErrorCode mErrorCode;
+    private String mRawErrorCode;
+
+    @NotNull
+    private ErrorCode mErrorCode = ErrorCode.DEFAULT;
+
+    @NotNull
+    private LocalErrorCode mLocalErrorCode = LocalErrorCode.DEFAULT;
```
kvenn: Should this be nullable? We'll probably only want to log local error codes if we see they have a value, and a null check seems more intuitive to me than not logging if it shows up as "default", but logging otherwise.

The alternative would be adding documentation to say that a value of "DEFAULT" means there is no "LocalError".

kvenn:
cool 😎

vimeo-networking/src/main/java/com/vimeo/networking/model/error/VimeoError.java
anthonycr: The previous behavior was not assigning the `DEFAULT` value to the enum in the case where the value was unrecognized, instead it would assign it to null. I looked at usage of this property and it is being assumed to be not null and it is assumed that `DEFAULT` is being assigned if no code is present. For those reasons, I decided to change it to not null and assign it a default value. 

Alternately, I can remove the default and `null` can be used instead. However, making it non null and giving it a default value makes it easier for consumers to switch on the value.

kvenn: I think either is fine, but whatever you choose it should just be clearly documented.

anthonycr: I added a comment to the getter of `ErrorCode` and `LocalErrorCode` documenting that if no value was set, then the value will be set to `DEFAULT`.

anthonycr:
@howardrigberg back to you

vimeo-networking/src/main/java/com/vimeo/networking/model/error/InvalidParameter.java
```diff
@@ -54,16 +66,54 @@ public InvalidParameter() {}
...
+     * Returns the raw error code, may be null.
+     *
+     * @return the raw error code as a nullable string.
+     */
+    @Nullable
+    public String getRawErrorCode() {
+        return mRawErrorCode;
+    }
+
+    public String getField() {
```
howardrigberg: This needs a java doc

anthonycr: this method existed in the class prior to this PR, should I add javadocs to all methods in this class?

howardrigberg: Oh yeah, you just moved it... Obviously, it's better to have all of them if you have enough time to add them. I'll leave it up to you.

PR #330, opened 2018-09-17
Adding full constructor to Recommendation for copy use

anthonycr:
# Summary

`Recommendation` is the only model containing an `Entity` that does not allow those entities to be changed. We should add the ability to change these entity instances. Other models are mutable and have setters. Instead of making this class mutable, I have opted to add a constructor that can be used for copying from the previous `Recommendation` instance.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/Recommendation.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/Recommendation.java
@@ -55,13 +55,25 @@ public class Recommendation implements Serializable, Entity {
     Recommendation() { }
 
     public Recommendation(@Nullable Category category,
-                          @Nullable String recommendationType,
+                          @Nullable RecommendationType recommendationType,
                           @Nullable User user,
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/Recommendation.java
```diff
@@ -51,6 +51,35 @@
...
+
+    public Recommendation(@Nullable Category category,
+                          @Nullable RecommendationType recommendationType,
+                          @Nullable User user,
+                          @Nullable Channel channel,
+                          @Nullable String description,
+                          @Nullable String resourceKey) {
+        mCategory = category;
+        if (recommendationType != null) {
+            switch (recommendationType) {
```
howardrigberg: It would help to clean up the code if this switch and getRecommendationType were localized in the the RecommendationType enum class.

anthonycr:
@howardrigberg changes made, back to you for review.

PR #334, opened 2018-10-12
Moshi & Moshi-Kotlin for Serialization

msya:
The purpose of this PR is to illustrate the usage of Moshi for serialization and Moshi-Kotlin module.

**Implementation**
- All DTOs are annotated with `@JsonClass(generateAdapter = true)`. This will autogenerate an adapter for DTO.
- The DTOs have also been annotated with `@Json(name = "...")` to map to the specific fields. 
- There is an `example-moshi-android` module which illustrates using Moshi with Retrofit. It has an example request of getting all the albums. 
- It contains setup for Moshi to provide a custom date adapter.

There are several benefits and drawbacks of Moshi over GSON.

**Benefits**
- Small APK size. Gson is 227 KiB, Moshi+Okio together are 200 KiB.
- `moshi-code-gen` module generates all the adapters for each module. Refection is not used in this module.
- The library provides Kotlin support. The generated adapters are in Kotlin.
- The adapters are cached.
- Supports immutable DTOs.
- Supports nullable fields.
- Custom adapters are easy to write. Here is an example [Custom Adapter](https://github.com/square/moshi/blob/master/adapters/src/main/java/com/squareup/moshi/Rfc3339DateJsonAdapter.java).
- Qualifiers like @HexColor int permit multiple JSON representations for a single Java type.
- Built on top of Okio.
- Built by the same people who made GSON.
- Performance faster than GSON. I used this [benchmark repo](https://github.com/hzsweers/json-serialization-benchmarking) to analyze the performance. I ran it on my mac and found the Moshi-Kotlin version was faster than the Gson library.

**Drawbacks**
- Different versions of our API return different fields. For example, for API version 3.4.1, returns a collection of pictures whereas newer API versions return a single picture object. Moshi doesn't have `@Since` annotation like GSON to versioning.
- Less configurable - no field naming strategy. Moshi want you to name your field as `field_name` as the exact naming in the GSON.

**Resources**
- https://www.reddit.com/r/androiddev/comments/684flw/why_use_moshi_over_gson/
- https://medium.com/square-corner-blog/moshi-another-json-processor-624f8741f703


models/src/main/java/com/vimeo/networking2/Album.kt
```diff
@@ -8,112 +10,117 @@ import java.util.*
...
     /**
      * The time in ISO 8601 format that the album was created.
      */
+    @Json(name = "created_time")
     val createdTime: Date? = null,
 
-    /**
-     * Custom logo.
-     */
-    val customLogo: PictureCollection? = null,
```
anthonycr: a few properties were removed from this class, was that intentional?

models/src/main/java/com/vimeo/networking2/TextTrack.kt
```diff
@@ -1,58 +1,57 @@
...
     /**
      * The read-only URL of the text track file, intended for use with HLS playback.
      */
+    @Json(name = "hsl_link")
     val hlsLink: String? = null,
 
-    /**
-     * The time in ISO 8601 format when the read-only HLS playback text track file expires.
-     */
-    val hlsLinkExpiresTime: Date? = null,
```
anthonycr: looks like there was two expiration properties that have been removed, intentional?

models/src/main/java/com/vimeo/networking2/FacebookConfiguration.kt
```diff
@@ -1,20 +1,25 @@
 package com.vimeo.networking2
+import com.squareup.moshi.Json
```
anthonycr: formatting issue here, and a similar one exists in other files, might want to run the formatter over all files just to be in the clear.

models/src/main/java/com/vimeo/networking2/Album.kt
msya: I added it back in.

models/src/main/java/com/vimeo/networking2/TextTrack.kt
msya: I added back in.

models/src/main/java/com/vimeo/networking2/FacebookConfiguration.kt
msya: I ran the formatter on all the files in the models module.

msya:
@anthonycr These are the changes I made since last review:

- I added a test app to test request for Vimeo Networking 2.0 - [MainActivity](https://github.com/vimeo/vimeo-networking-java/pull/334/files#diff-4f9e1931f77087120b3c740853b326eb.)
- I added [Error Code](https://github.com/vimeo/vimeo-networking-java/pull/334/files#diff-74816d53e11d110461c6b02c6a8e4e91) and [ApiError](https://github.com/vimeo/vimeo-networking-java/pull/334/files#diff-4f278b15e3448578ddb07663b5d09ec5) classes.
- All enums have `@Json(name = "")` annotation.
- I created a custom date adapter to parse dates that maybe null. [CustomDateAdapter](https://github.com/vimeo/vimeo-networking-java/pull/334/files#diff-92d60a4e2a51d99b8247d2f7b74e78e1).
- I tested the DTOs for all get requests in the test app.

Back to you.

models/src/main/java/com/vimeo/networking2/Video.kt
```diff
@@ -190,20 +188,14 @@ data class Video(
...
      * An array of all tags assigned to this video.
      */
     @Json(name = "tags")
     val tags: List<Tag>? = null,
 
-    /**
-     * The transcode information for a video upload.
-     */
-    @Json(name = "transcode")
-    val transcode: TranscodeType? = TranscodeType.UNKNOWN,
```
anthonycr: Why was this property removed?

models/src/main/java/com/vimeo/networking2/adapters/CustomDateAdapter.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    @Synchronized
+    @Throws(IOException::class)
+    override fun fromJson(reader: JsonReader): Date? {
+        val token: JsonReader.Token = reader.peek()
+        return if (token == JsonReader.Token.NULL) {
+            reader.skipValue()
+            null
+        } else {
+            parseDate(reader)
```
anthonycr: I know there was concern about needing to write this class due to the null safety. I looked at the source of Moshi and it's possible that we won't need this class, assuming that the date formats supported by moshi are the ones our API supports.

Since this is the normal JSON Date adapter used by Moshi: https://github.com/square/moshi/blob/master/adapters/src/main/java/com/squareup/moshi/Rfc3339DateJsonAdapter.java

We can just do a `nullSafe` call on an instance of the adapter and we will be able to use it without having to write a custom adapter. See the function here: https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java#L130

How it will be consumed when setting up moshi:
```kotlin
.add(Date::class.java, Rfc3339DateJsonAdapter().nullSafe())
```
If Moshi is using a different date format then this won't work I don't think.

models/src/main/java/com/vimeo/networking2/enums/ErrorCode.kt
```diff
@@ -0,0 +1,464 @@
+package com.vimeo.networking2.enums
+
+import com.squareup.moshi.Json
+
+/**
+ * All the error codes that can come back from the Vimeo Api.
+ */
+enum class ErrorCode {
+    
+    @Json(name ="1000")
```
anthonycr: formatting problems in this file.

models/src/main/java/com/vimeo/networking2/enums/ErrorCode.kt
```diff
@@ -0,0 +1,464 @@
+package com.vimeo.networking2.enums
+
+import com.squareup.moshi.Json
+
+/**
+ * All the error codes that can come back from the Vimeo Api.
+ */
+enum class ErrorCode {
```
anthonycr: We should have a separate discussion about whether we want to support this `ErrorCode` enum as is because it has caused us problems in the previous design of the library. Examples of those problems:
- Not being able to handle an error until it was added to the library (no error code `Int` property to reference)
- It being an enum isn't all that helpful in practice because the number of codes that can be returned for a specific API call is very limited compared to the scope of the `ErrorCode`. 

I think it's fine to leave here as is for the time being, but we should definitely have a more serious discussion about to handle errors.

models/src/main/java/com/vimeo/networking2/Video.kt
msya: I had removed this property, because it didn't match what the API was returning. The API wraps the transcode status of a transcode object. I created a transcode object with the status.

models/src/main/java/com/vimeo/networking2/adapters/CustomDateAdapter.kt
msya: You're right. I updated the initialization to use `nullSafe`.

models/src/main/java/com/vimeo/networking2/enums/ErrorCode.kt
msya: Good point. I will take note of the issues around `ErrorCode` and leave this as it is for now. The upcoming phase of vimeo networking 2.0 is designing the API requests. It has error handling as part of the phase. Discussion around this issue and possible approach would fit the error handling task.

msya:
@anthonycr I made the changes you had requested. Back to you.

models/src/main/java/com/vimeo/networking2/Album.kt
```diff
@@ -8,112 +10,128 @@ import java.util.*
...
      * The privacy settings of the album.
      */
+    @Json(name = "privacy")
     val privacy: AlbumPrivacy? = null,
 
+    /**
+     * The album resource key.
+     */
+    @Json(name = "resource_key")
+    val resourceKey: String? = null,
```
anthonycr: Album should implement the `Entity` interface.

models/src/main/java/com/vimeo/networking2/Video.kt
```diff
@@ -226,7 +226,7 @@ data class Video(
     @Json(name = "width")
     val width: Int? = null
 
-) : Entity {
+): Entity {
```
anthonycr: Formatting problem.

PR #335, opened 2018-10-15
Add Lists DTOs

msya:
The purpose of this PR is to add all models for collections that could be returned by the API.

The lists can be pageable. So, a pageable interface has been created: 

```
interface Pageable<DATA_T> {

    val total: Int?

    val page: Int?

    val perPage: Int?

    val paging: Paging?

    val data: List<DATA_T>?
}
```

Any DTOs that are pageable must conform to this interface. The following pageable DTOs were created in this PR:

- CategoryList
- ChannelList
- CommentList
- FeedList
- NotificationList
- ProductsList
- CinemaList
- RecommendationList
- SearchResponse
- SeasonList
- TextTrackList
- TVoidItemList
- UserList
- VideoList


models/src/main/java/com/vimeo/networking2/Pageable.kt
```diff
@@ -0,0 +1,33 @@
+package com.vimeo.networking2
+
+/**
+ * All lists that can be paged should implement this interface to provide
+ * paging urls, total number of items returns for each page and the data for each page.
+ */
+interface Pageable<DATA_T> {
```
anthonycr: for consistency with other type parameters in the library this should be `Data_T` rather than in all caps

models/src/main/java/com/vimeo/networking2/enums/GroupForumsPrivacyValue.kt
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Type of group forums.
+ */
+enum class GroupForumsPrivacyValue {
+
+    /**
+     * Only moderators can comment on the group.
+     */
+    MODS
```
anthonycr: Should `UNKNOWN` types be added to all these enums that were added here? I don't see them mentioned in the PR description so I'm not sure what the plans are for them.

models/src/main/java/com/vimeo/networking2/Pageable.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Total number of items returned.
```
anthonycr: Since we have documentation on this interface, I wonder if it is necessary to have documentation on each of the subclasses.

models/src/main/java/com/vimeo/networking2/AlbumList.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Urls to the first, last page, next and previous pages.
+     */
+    override val paging: Paging? = null,
+
+    /**
+     * List of albums.
+     */
+    override val data: List<Album>?
```
anthonycr: There's a lot of `Pageable` implementations that have a default value of null here and a lot that don't. We should be consistent about whether there are default values for these properties.

msya:
@anthonycr Back to you.

PR #336, opened 2018-10-16
Add Entity Interface

msya:
Created entity interface in Kotlin to get a unique identifier for an object.

```
interface Entity {
    val identifier: String?
}
```
All models with resource keys should inherit this interface.  These are the models that were updated:

- Category
- Channel
- Comment
- Group
- Recommendation 
- Tag
- Season
- User
- Video

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/Comment.kt
+++ b/models/src/main/java/com/vimeo/networking2/Comment.kt
@@ -37,6 +37,8 @@ data class Comment(
     /**
      * The user who posted the comment.
      */
-    val user: User?
+    val user: User?,
 
-)
+    override val identifier: String? = resourceKey
+
+): Entity
--- a/models/src/main/java/com/vimeo/networking2/Comment.kt
+++ b/models/src/main/java/com/vimeo/networking2/Comment.kt
@@ -37,8 +37,10 @@ data class Comment(
     /**
      * The user who posted the comment.
      */
-    val user: User?,
+    val user: User? = null
+
+): Entity {
 
     override val identifier: String? = resourceKey
 
-): Entity
+}
--- a/models/src/main/java/com/vimeo/networking2/Entity.kt
+++ b/models/src/main/java/com/vimeo/networking2/Entity.kt
@@ -1,0 +1,17 @@
+package com.vimeo.networking2
+
+/**
+ * Represents data that can be uniquely identified by an [Entity.identifier].
+ *
+ * Implementors of this interface must provide a unique identifier.
+ */
+interface Entity {
+
+    /**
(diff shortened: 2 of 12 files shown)


models/src/main/java/com/vimeo/networking2/Comment.kt
```diff
@@ -37,6 +37,8 @@ data class Comment(
     /**
      * The user who posted the comment.
      */
-    val user: User?
+    val user: User?,
```
anthonycr: this model is missing a default value

models/src/main/java/com/vimeo/networking2/Entity.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Implementors of this interface must provide a unique identifier.
+ */
+interface Entity {
+
+    /**
+     * The identifier, as a [String], that uniquely identifies this entity.
+     *
+     * @return a [String], or null if the entity does not have an identifier.
+     */
+    val identifier: String?
```
anthonycr: After seeing how this will be implemented in the new DTOs, I think that we should rename this property to `resourceKey` (or rename `resourceKey` to `identifier` if you think that's more semantically correct). The reason for this is that the way the `Entity` interface is implemented in each DTO means that the identifier can be changed independently of the resource key. The below code for instantiating a `Category` is valid code:
```kotlin
val category1 = Category(resourceKey = "1", identifier = "2") // identifier == 2
val category2 = Category(resourceKey = "1') // identifier == 1
```
If you think that `identifier` should be separate from the properties of the DTO, then another alternative is to implement the interface as:
```kotlin
data class Category(val resourceKey: String? = null) : Entity {
  override val identifier: String? = resourceKey
}
```

models/src/main/java/com/vimeo/networking2/Comment.kt
anthonycr: Since we keep finding missing default values, an easy way to guarantee that all DTOs have default values and that new properties added also have defaults would be to create a test that just instantiated each DTO without any constructor parameters.

msya: I have an upcoming tasks on writing tests for models. I'll add tests for it in that task.

models/src/main/java/com/vimeo/networking2/Entity.kt
msya: I like the second option of having it separate from the properties of the DTO. . The benefit is that the DTO will represent what is returned by the API. Another benefit is that it doesn’t tie down to the resource key as always being the unique identifier.

msya:
@anthonycr Back to you.

PR #339, opened 2018-10-29
VA-3570 Adding support for albums

howardrigberg:
#### Issue
https://github.com/vimeo//vimeo-networking-java/issues/VA-3570

#### Summary
This ticket is about adding albums support to the library. The `Album` data type will be added along with the related data classes and request support necessary to get `Album` information from the API.

#### How to Test
No testing necessary at this point.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/InteractionCollection.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/InteractionCollection.java
@@ -23,11 +23,16 @@
 package com.vimeo.networking.model;
 
 import com.google.gson.annotations.SerializedName;
+import com.vimeo.networking.VimeoClient;
+import com.vimeo.networking.callbacks.IgnoreResponseVimeoCallback;
 import com.vimeo.stag.UseStag;
 
 import org.jetbrains.annotations.Nullable;
 
 import java.io.Serializable;
+import java.util.ArrayList;
+
+import okhttp3.CacheControl;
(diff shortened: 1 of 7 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/InteractionCollection.java
```diff
@@ -83,6 +83,14 @@
     @SerializedName("moderate_videos")
     private Interaction mModerateVideos;
 
+    @Nullable
+    @SerializedName("add_logos")
+    private Interaction mAddLogos;
+
+    @Nullable
+    @SerializedName("add_videos")
+    private Interaction mAddVideos;
```
msya: The uri returned from the API for adding videos to an album is `/users/{user_id}/albums/{album_id}/videos/{video_id}`. The API requires appending the video id to the uri. Providing video id as a query param or part of the request body will not work. There are two options that I think that could be taken:

1. We could add documentation to the `getAddVideos` method to indicate you have to mutate the uri by adding the video id to it.

2. An` addVideosToAlbum` method could be created in the SDK that mutates the uri by appending the video id for the client.

Option 2 is optional. But, I would recommend adding a doc to the `getAddVideos` method.

howardrigberg: Great suggestion, @msya. I chose option #1. Back to you.

PR #340, opened 2018-10-29
Add Feed Models

msya:
## Summary
- Added feed list model
- Added Moshi Annotations for all Pageable models.

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/AlbumList.kt
+++ b/models/src/main/java/com/vimeo/networking2/AlbumList.kt
@@ -11,18 +11,18 @@ import com.vimeo.networking2.common.Pageable
 data class AlbumList(
 
     @Json(name = "total")
-    override val total: Int?,
+    override val total: Int? = null,
 
     @Json(name = "page")
-    override val page: Int?,
+    override val page: Int? = null,
 
     @Json(name = "per_page")
-    override val perPage: Int?,
+    override val perPage: Int? = null,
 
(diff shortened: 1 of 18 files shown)


models/src/main/java/com/vimeo/networking2/FeedItem.kt
```diff
@@ -0,0 +1,79 @@
...
+ * An item in the user's feed.
+ */
+@JsonClass(generateAdapter = true)
+data class FeedItem(
+
+    /**
+     * The category that this event occurred for. This will be present for only
+     * [AttributionType.CATEGORY] feed item type.
+     */
+    @Json(name ="category")
```
anthonycr: formatting problem here.

models/src/main/java/com/vimeo/networking2/AlbumList.kt
```diff
@@ -1,14 +1,28 @@
...
+import com.squareup.moshi.JsonClass
 import com.vimeo.networking2.common.Pageable
 
 /**
  * List of albums that could be paged.
  */
+@JsonClass(generateAdapter = true)
 data class AlbumList(
+
+    @Json(name = "total")
```
anthonycr: I can't think of a way to de-duplicate these `@Json` annotations without making an abstract class and then subclassing it with the type, e.g. `AlbumList : AbstractList<Album>()`. I guess it might not be a big deal though.

PR #341, opened 2018-10-31
Update VimeoClient for using Moshi

msya:
## Summary

The purpose of this PR is to add support for the new models in the current implementation of `VimeoClient`.  Using the new models will require using Moshi for serialization and deserialization. The changes made to `VimeoClient` should be backward compatible. The client should be able to continue to use the old models. 

In order to achieve these requirements, these are the changes that were made:

1. [AdapterConverterFactory](https://github.com/vimeo/vimeo-networking-java/compare/v2.0.0...staff-picks-test?expand=1#diff-6ee966a68e85fa09ead8b7bf1389c920) was added to be able to use the Moshi or Gson converter with Retrofit. The converter that you wish to use should be specified as an annotation on `VimeoService`. Here is an example [Get users request using the new models](https://github.com/vimeo/vimeo-networking-java/compare/v2.0.0...staff-picks-test?expand=1#diff-ad54125240d6234452e1b230ba362e18R317). If you do not specify an annotation, the Gson converter will be used by default.

2. Caller classes were created for the new models. [Request Callers for New Models](https://github.com/vimeo/vimeo-networking-java/compare/v2.0.0...staff-picks-test?expand=1#diff-866fb795429064a803d34426b80b44e3R49). If a request needs to be made using the new models, use one of the callers in this class and specify the new models in the request.

Ex:

```
mApiClient.getContent(STAFF_PICKS_VIDEO_URI, 
                       CacheControl.FORCE_NETWORK, 
                       MoshiGetRequestCaller.VIDEO_LIST, null, null, null, 
                       new VimeoCallback<com.vimeo.networking2.VideoList>() {
            
                          @Override
                          public void success(com.vimeo.networking2.VideoList videoList) { ... }

                           @Override
                           public void failure(VimeoError error) { ... }

});
```

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
@@ -25,6 +25,7 @@ public final class AnnotatedConverterFactory extends Converter.Factory {
     @Retention(RetentionPolicy.RUNTIME)
     @interface Serializer {
 
+        @NotNull
         ConverterType converter();
 
     }
@@ -62,8 +63,9 @@ public final class AnnotatedConverterFactory extends Converter.Factory {
     }
 
     @Override
-    public Converter<ResponseBody, ?> responseBodyConverter(final Type type, final Annotation[] annotations,
-                                                            final Retrofit retrofit) {
+    public Converter<ResponseBody, ?> responseBodyConverter(@NotNull final Type type,
+                                                            @NotNull final Annotation[] annotations,
+                                                            @NotNull final Retrofit retrofit) {
         Converter.Factory converterFactory = gsonFactory;
 
         for (final Annotation annotation : annotations) {
--- a/vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
@@ -1,91 +1,81 @@
 package com.vimeo.networking;
 
+import org.jetbrains.annotations.NotNull;
+
 import java.lang.annotation.Annotation;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.reflect.Type;
-import java.util.LinkedHashMap;
-import java.util.Map;
 
 import okhttp3.ResponseBody;
 import retrofit2.Converter;
-import retrofit2.Converter.Factory;
 import retrofit2.Retrofit;
 
 /**
  * This is a converter factory that allows you to specify different converters to use for each request.
  * This was taken from https://stackoverflow.com/questions/40824122/android-retrofit-2-multiple-converters-gson-simplexml-error.
  * It provides two annotations - Gson and Moshi that should be used on a Retrofit request service to specify which
- * serialization framework to use.
- * <p>
- * Ex:
- *
- * @<code>
- *
- *  @GET
- *  @Moshi Call<com.vimeo.networking2.VideoList> getVideoList(
- *  @Header("Authorization") String authHeader,
- *  @Url String uri,
- *  @QueryMap Map<String, String> options,
- *  @Header("Cache-Control") String cacheHeaderValue);
- *
- * </code>
- *
- * If a annotation is not specified, Gson will be used.
+ * serialization framework to use. If a annotation is not specified, Gson will be used.
  */
 public final class AnnotatedConverterFactory extends Converter.Factory {
 
+    /**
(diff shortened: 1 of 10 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
```diff
@@ -0,0 +1,91 @@
...
+            return gsonFactory.responseBodyConverter(type, annotations, retrofit);
+        }
+        return null;
+    }
+
+    static class Builder {
+
+        Map<Class<?>, Factory> mFactoryMap = new LinkedHashMap<>();
+
+        Builder add(final Class<? extends Annotation> factoryType, final Converter.Factory factory) {
```
anthonycr: params should be annotated non null

vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private AnnotatedConverterFactory(final Map<Class<?>, Converter.Factory> factoryMap) {
+        mFactoryMap = factoryMap;
+    }
+
+    @Override
+    public Converter<ResponseBody, ?> responseBodyConverter(final Type type,
+                                                            final Annotation[] annotations,
+                                                            final Retrofit retrofit) {
+        for (final Annotation annotation : annotations) {
```
anthonycr: Is it possible that we could simplify this factory? According to the documentation of the converter interface, returning null signals that the converter does not support the response. What could be done instead is that instead of holding a map, the `AnnotatedConverterFactory` could just take an annotation class and a `Converter.Factory` as a parameter. Then the `responseBodyConverter` method would just check if the annotation matches, and if so it would just convert the response using the injected factory. The benefit I see in doing this is that it reduces the amount of logic necessary in this class, and would eliminate the need for the builder. In order to prevent a scenario where multiple annotations are provided and potentially undefined behavior occurs, a single annotation could be used
```
annotation class Serializer(Converter.GSON/MOSHI)
```
Happy to talk about this further.

msya:
@anthonycr Back to you.

vimeo-networking/src/main/java/com/vimeo/networking/RetrofitSetup.java
```diff
@@ -89,15 +93,23 @@
     public Retrofit createRetrofit() {
         return new Retrofit.Builder().baseUrl(mConfiguration.getBaseUrl())
                 .client(createOkHttpClient())
-                .addConverterFactory(GsonConverterFactory.create(mGson))
+                .addConverterFactory(new AnnotatedConverterFactory(GsonConverterFactory.create(mGson),
+                                                                   MoshiConverterFactory.create(
+                                                                           createMoshi())))
```
anthonycr: formatting here is quite awkward.

vimeo-networking/src/main/java/com/vimeo/networking/RetrofitSetup.java
```diff
@@ -89,15 +93,23 @@
...
                 .build();
     }
 
+    /**
+     * Create an instance of Moshi with a date adapter.
+     */
+    private Moshi createMoshi() {
```
anthonycr: annotate non null. there are also formatting problems in this file.

vimeo-networking/src/main/java/com/vimeo/networking/RetrofitSetup.java
```diff
@@ -131,11 +143,11 @@ private void setupCertPinning(@NotNull RetrofitClientBuilder retrofitClientBuild
     /**
      * @return value used for {@code User-Agent} HTTP header value.
      */
-    @NotNull
-    String createUserAgent() {
+    @NotNull String createUserAgent() {
         final String userProvidedAgent = mConfiguration.getUserAgentString();
 
-        if (userProvidedAgent != null && !userProvidedAgent.isEmpty() && isValidUserAgent(userProvidedAgent)) {
+        if (userProvidedAgent != null && !userProvidedAgent.isEmpty() &&
```
anthonycr: ampersand should go on next line per codestyle.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -312,6 +314,13 @@
                                @QueryMap Map<String, String> options,
                                @Header("Cache-Control") String cacheHeaderValue);
 
+    @GET
+    @Serializer(converter =  ConverterType.MOSHI)
```
anthonycr: formatting problems in this file.

PR #343, opened 2018-11-02
Add staff picks example with Moshi

msya:
The purpose of this PR is to add an example in the Java app illustrating how to make request with Moshi. The request gets staff picks and records the time it took to get the data for both Gson and Moshi.


msya:
@kvenn Thanks for the review. I updated the Kotlin example to calculate how long the request take with Coroutines. I modified the Coroutines logic to specify UI scope, cancellation and used deferred to get the result. The result of how long the request takes is now shown in the UI for the both the Java and Kotlin samples. When running the sample, I get similar results as you had above. Back to you.

example-java-android/src/main/java/com/vimeo/android/networking/example/TestApp.java
```diff
@@ -21,7 +21,7 @@
 
     private static final boolean IS_DEBUG_BUILD = false;
     // Switch to true to see how access token auth works.
-    private static final boolean ACCESS_TOKEN_PROVIDED = false;
+    private static final boolean ACCESS_TOKEN_PROVIDED = true;
```
kvenn: Is this an intentional change?

PR #344, opened 2018-11-02
Write DTOs Test

msya:
Tests were added for all models and enums. 

Model Test Cases
- Verify each field in the constructor is nullable.
- Verify the model is a data class.

Enum Test Cases
- Verify an UNKNOWN type was defined in the enum.


kvenn:
📆I will get to this tomorrow (apologies for the delay).

msya:
@kvenn I also looked into the Reflections library. I see the advantage the library gives to get all classes in a package via reflection. The tradeoff is readability. The test I have thus far are very simple. I prefer the option of listing all the classes and enums for now. The list could be extracted into a constant so that several tests can use it. The downside is syncing the list with the DTOs. It would be good to explore libraries as we look into validating against the Open API spec.

PR #346, opened 2018-11-06
[WIP] Extensions extraction for DTO enum fields

anthonycr:
# Summary
The nature of our API is such that enum values may be added that are publicly available and are also backwards compatible. An example is if a new account type is created, that account type must be accessible to old consumers of the API. As such, we cannot guarantee that the list of enums shipped with any specific version of the library will be comprehensive, and so we must make a fallback available.

## Implementation
I created an inline extension function on `String` that can convert the value to an `enum`. The enum must also implement a `StringValue` interface. Using this `asEnum` function, extensions can be created on a DTO that allow an enum value to be extracted from a `String` property on the DTO. It does this by grabbing all the enum values from the enum and checking the `StringValue.value` property that each enum is required to provide against the `String` provided. So for instance, on the `User` class, an extension property was created called `accountType: AccountType` and it converts the `account` property to its enum representation or defaults to `AccountType.UNKNOWN` if the type is unknown. This extension function can be consumed from Java or Kotlin:

```java
final User user = ...
final AccountType accountType = UserExtensions.getAccountType(user);
```

```kotlin
val user = ...
val accountType = user.accountType
```

As this is a WIP, here are the enums I converted so far are:
- AccountType
- AlbumLayoutType
- AlbumPrivacyViewValue
- AlbumThemeType
- ApiOptionsType

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/Album.kt
+++ b/models/src/main/java/com/vimeo/networking2/Album.kt
@@ -1,3 +1,5 @@
+@file:JvmName("AlbumUtils")
+
 package com.vimeo.networking2
 
 import com.squareup.moshi.Json
--- a/models/src/main/java/com/vimeo/networking2/Album.kt
+++ b/models/src/main/java/com/vimeo/networking2/Album.kt
@@ -5,6 +5,7 @@ import com.squareup.moshi.JsonClass
 import com.vimeo.networking2.enums.AlbumLayoutType
 import com.vimeo.networking2.enums.AlbumThemeType
 import com.vimeo.networking2.enums.SortType
+import com.vimeo.networking2.enums.asEnum
 import java.util.*
 
 /**
@@ -60,7 +61,7 @@ data class Album(
      * The album's layout preference.
      */
     @Json(name = "layout")
-    val layout: AlbumLayoutType? = null,
+    val layout: String? = null,
 
     /**
      * The URL to access the album.
@@ -120,7 +121,7 @@ data class Album(
      * The album's color theme preference.
      */
     @Json(name = "theme")
-    val theme: AlbumThemeType? = null,
+    val theme: String? = null,
 
     /**
      * The album's URI.
@@ -139,3 +140,15 @@ data class Album(
     override val identifier: String? = resourceKey
 
 }
+
+/**
+ * @see Album.layout
+ */
+val Album.albumLayoutType: AlbumLayoutType
+    get() = layout?.asEnum() ?: AlbumLayoutType.UNKNOWN
+
+/**
+ * @see Album.theme
+ */
+val Album.themeType: AlbumThemeType
+    get() = theme?.asEnum() ?: AlbumThemeType.UNKNOWN
--- a/models/src/main/java/com/vimeo/networking2/PurchaseInteraction.kt
+++ b/models/src/main/java/com/vimeo/networking2/PurchaseInteraction.kt
@@ -1,9 +1,8 @@
 package com.vimeo.networking2
 
-import com.vimeo.networking2.common.Interaction
-import com.vimeo.networking2.enums.ApiOptionsType
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
+import com.vimeo.networking2.common.Interaction
 import com.vimeo.networking2.enums.PurchaseStatusType
 
 /**
@@ -13,7 +12,7 @@ import com.vimeo.networking2.enums.PurchaseStatusType
 data class PurchaseInteraction(
 
     @Json(name = "options")
-    override val options: List<ApiOptionsType>? = null,
+    override val options: List<String>? = null,
 
     @Json(name = "uri")
     override val uri: String? = null,
--- a/models/src/main/java/com/vimeo/networking2/AlbumPrivacy.kt
+++ b/models/src/main/java/com/vimeo/networking2/AlbumPrivacy.kt
@@ -3,6 +3,7 @@ package com.vimeo.networking2
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 import com.vimeo.networking2.enums.AlbumPrivacyViewValue
+import com.vimeo.networking2.enums.asEnum
 
 /**
  * The privacy set for an album.
@@ -20,5 +21,11 @@ data class AlbumPrivacy(
      * Who can view the album.
      */
     @Json(name = "view")
-    val viewingPermissions: AlbumPrivacyViewValue? = null
+    val viewingPermissions: String? = null
 )
+
+/**
+ * @see AlbumPrivacy.viewingPermissions
+ */
+val AlbumPrivacy.viewingPermissionsType: AlbumPrivacyViewValue
+    get() = viewingPermissions?.asEnum() ?: AlbumPrivacyViewValue.UNKNOWN
--- a/models/src/main/java/com/vimeo/networking2/AlbumPrivacy.kt
+++ b/models/src/main/java/com/vimeo/networking2/AlbumPrivacy.kt
@@ -1,3 +1,5 @@
+@file:JvmName("AlbumPrivacyUtils")
+
 package com.vimeo.networking2
 
 import com.squareup.moshi.Json
--- a/models/src/main/java/com/vimeo/networking2/BuyInteraction.kt
+++ b/models/src/main/java/com/vimeo/networking2/BuyInteraction.kt
@@ -3,7 +3,6 @@ package com.vimeo.networking2
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 import com.vimeo.networking2.common.Interaction
-import com.vimeo.networking2.enums.ApiOptionsType
 import com.vimeo.networking2.enums.DownloadType
 import com.vimeo.networking2.enums.StreamType
 import java.util.*
@@ -63,7 +62,7 @@ data class BuyInteraction(
     val streamType: StreamType? = null,
 
     @Json(name = "options")
-    override val options: List<ApiOptionsType>? = null,
+    override val options: List<String>? = null,
 
     @Json(name = "uri")
     override val uri: String? = null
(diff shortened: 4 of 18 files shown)


models/src/main/java/com/vimeo/networking2/enums/TranscodeType.kt
```diff
@@ -1,17 +1,12 @@
 package com.vimeo.networking2.enums
 
-import com.squareup.moshi.Json
+enum class TranscodeType(override val value: String?) : StringValue {
```
howardrigberg: This class is missing kdocs

models/src/main/java/com/vimeo/networking2/enums/UploadSpaceType.kt
```diff
@@ -1,14 +1,10 @@
 package com.vimeo.networking2.enums
 
-import com.squareup.moshi.Json
+enum class UploadSpaceType(override val value: String?) : StringValue {
```
howardrigberg: This class is missing kdocs

models/src/main/java/com/vimeo/networking2/enums/UploadStatusType.kt
```diff
@@ -1,17 +1,12 @@
 package com.vimeo.networking2.enums
 
-import com.squareup.moshi.Json
+enum class UploadStatusType(override val value: String?) : StringValue {
```
howardrigberg: This class is missing kdocs

models/src/main/java/com/vimeo/networking2/enums/UserActivityType.kt
```diff
@@ -1,80 +1,55 @@
...
-
 /**
  * All type of actions that can be taken by the user.
  */
-enum class UserActivityType {
+enum class UserActivityType(override val value: String?) : StringValue {
 
-    @Json(name = "add_comment")
-    ADD_COMMENT,
+    ADD_COMMENT("add_comment"),
```
howardrigberg: This class is missing kdocs on the types

models/src/main/java/com/vimeo/networking2/enums/VideoActionType.kt
```diff
@@ -1,23 +1,17 @@
...
-
 /**
  * All actions that can be taken on a video.
  */
-enum class VideoActionType {
+enum class VideoActionType(override val value: String?) : StringValue {
 
-    @Json(name = "Added to")
-    ADDED_TO,
+    ADDED_TO("Added to"),
```
howardrigberg: This class is missing kdocs on the types

models/src/main/java/com/vimeo/networking2/enums/VideoBadgeType.kt
```diff
@@ -1,29 +1,21 @@
...
-
 /**
  * Different types of badges for videos.
  */
-enum class VideoBadgeType {
+enum class VideoBadgeType(override val value: String?) : StringValue {
 
-    @Json(name = "cameo")
-    CAMEO,
+    CAMEO("cameo"),
```
howardrigberg: This class is missing kdocs on the types

models/src/main/java/com/vimeo/networking2/enums/VideoQualityType.kt
```diff
@@ -1,23 +1,17 @@
...
-
 /**
  * Video qualities.
  */
-enum class VideoQualityType {
+enum class VideoQualityType(override val value: String?) : StringValue {
 
-    @Json(name = "hd")
-    HD,
+    HD("hd"),
```
howardrigberg: This class is missing kdocs on the types

models/src/main/java/com/vimeo/networking2/enums/VideoSourceType.kt
```diff
@@ -1,20 +1,14 @@
 package com.vimeo.networking2.enums
 
-import com.squareup.moshi.Json
+enum class VideoSourceType(override val value: String?) : StringValue {
```
howardrigberg: This class is missing kdocs

models/src/main/java/com/vimeo/networking2/enums/VideoStatusType.kt
```diff
@@ -1,38 +1,27 @@
...
-
 /**
  * Video statuses.
  */
-enum class VideoStatusType {
+enum class VideoStatusType(override val value: String?) : StringValue {
 
-    @Json(name = "available")
-    AVAILABLE,
+    AVAILABLE("available"),
```
howardrigberg: This class is missing kdocs on the types

models/src/main/java/com/vimeo/networking2/Video.kt
```diff
@@ -185,10 +189,11 @@ data class Video(
     /**
      * The status code for the availability of the video. This field is deprecated in favor
      * of [upload] and [transcode].
+     * @see Video.statusType
      */
     @Json(name = "status")
-    @Deprecated("This field is deprecated in favor of upload and transcode.")
-    val status: VideoStatusType? = null,
+    @Deprecated("This property is deprecated in favor of upload and transcode.")
```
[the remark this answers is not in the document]
anthonycr: My view is that if a property is still available via the API, we should support it, even if it is deprecated. An additional reason to keep it is that we still heavily use the property in our Android application. 

Maybe @msya can weigh in on what he thinks.

models/src/main/java/com/vimeo/networking2/Album.kt
```diff
@@ -139,3 +145,24 @@ data class Album(
     override val identifier: String? = resourceKey
 
 }
+
+/**
+ * @see Album.layout
+ * @see AlbumLayoutType
+ */
+val Album.albumLayoutType: AlbumLayoutType
```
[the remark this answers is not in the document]
anthonycr: I don't have a strong preference either way. The code is definitely boilerplate and annoying to write, but it's not worryingly error prone or difficult to maintain as the `get() =` is straight forward and the underlying logic is shared between every extension. Discoverability _might_ suffer with the usage of annotation processing for those reading source code. However, the only concern I have is in relation to the reference in the property documentation to the extension property. I'm not sure how well KDoc would do referencing a generated property. Regarding the feasibility of creating such an annotation processor, I would say that it's very easy. The design and implementation would look very similar to this simple example annotation processor I created for a talk: https://github.com/anthonycr/Tagger

@msya do you think it would be a good idea to generate these using an annotation processor, or would you prefer to keep them in the source of the file? If you prefer to generate them, I can ticket the work.

models/src/main/java/com/vimeo/networking2/Video.kt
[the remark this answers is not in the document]
msya: For v1 of models, the goal is to have parity with our current models. I prefer keeping the status field for now as our app uses it. The API spec will in the future specify a deprecated property for fields and it will specify the version in which it was deprecated. This will help in annotating which fields should be deprecated. There is a ticket in the queue to handle API versioning and this problem would be addressed as part of that ticket.

models/src/main/java/com/vimeo/networking2/Cinema.kt
```diff
@@ -42,9 +45,10 @@ data class Cinema(
 
     /**
      * The type of programmed cinema item.
+     * @see Cinema.cinemaType
      */
     @Json(name = "type")
-    val type: CinemaType? = null,
+    val type: String? = null,
```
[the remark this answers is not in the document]
anthonycr: used "programmedContentItem" instead due to recent renaming by @msya on another PR

models/src/main/java/com/vimeo/networking2/Album.kt
```diff
@@ -58,9 +61,10 @@ data class Album(
 
     /**
      * The album's layout preference.
+     * @see Album.albumLayoutType
      */
     @Json(name = "layout")
-    val layout: AlbumLayoutType? = null,
+    val layout: String? = null,
```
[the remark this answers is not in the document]
anthonycr: Naming patterns are now as follows for enums (it's slightly relaxed but basically follows this):
```kotlin
class ClassName {
  @Json(name = "enum_name")
  val enumName: String?
}

val ClassName.enumNameType: EnumNameType
```

For enums that represent the type of the class (e.g. `NotificationType`, `VideoBadgeType`), the naming pattern is strictly as follows:
```kotlin
class ClassName {
  @Json(name = "type")
  val rawType: String?
}

val ClassName.type: ClassNameType
```

models/src/main/java/com/vimeo/networking2/VideoContext.kt
```diff
@@ -12,14 +15,22 @@ data class VideoContext(
...
      */
     @Json(name = "action")
-    val action: VideoActionType? = null,
+    val action: String? = null,
 
     /**
      * The contextual resource type.
      */
-    @Json(name = "resource_key")
+    @Json(name = "resource_type")
```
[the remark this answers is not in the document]
anthonycr: i'm not sure

anthonycr:
@howardrigberg I created a ticket for documenting the enums as that work is quite involved, see here: https://vimean.atlassian.net/browse/VA-3704 I think I got every enum that you commented on.

I made several changes
- I fixed the naming of enums so that all enums now have `-Type` appended.
- I enabled the CI build on the `models` module.
- I renamed all the fields that have an `@Json(name = "type")` annotation to `rawType: String` and the extension to `type: Enum`. Other enums have normal naming scheme of `propertyName: String` and `propertyNameType: Enum`
- Created `SeasonType` enum as it was missing.
- Renamed `Tvod` classes to be consistently named... one was named `Tvoid`, some were named `TVod`.
- Added missing `@Json` annotations
- Added missing `@JsonClass` annotations
- Fixed formatting on a number of files.

models/src/main/java/com/vimeo/networking2/PurchaseInteraction.kt
```diff
@@ -13,15 +15,23 @@ import com.vimeo.networking2.enums.PurchaseStatusType
 data class PurchaseInteraction(
 
     @Json(name = "options")
-    override val options: List<ApiOptionsType>? = null,
+    override val options: List<String>? = null,
```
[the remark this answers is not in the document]
anthonycr: This will hide the documentation it inherits from the `Interaction` interface, so i'm going to leave it as is.

PR #347, opened 2018-11-08
Configuration request for new DTOs

anthonycr:
# Summary
Created new function in `VimeoService` and new `GetRequestCaller` field that allows for requesting an `AppConfiguration` object from the `/configs` endpoint. Additionally, a bug was fixed in the `AnnotatedConverterFactory` where the `requestBodyConverter` was never being created. I also did some refactoring in the factory to pull out the code searching for an annotation into a separate function.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
@@ -90,6 +90,27 @@ public final class GetRequestCaller {
      * Used in association with
      * {@link VimeoClient#getContent(String, CacheControl, Caller, String, Map, String, VimeoCallback)} or
      * {@link VimeoClient#getContentSync(String, CacheControl, String, Map, String, Caller)}
+     * to get an {@link com.vimeo.networking2.AppConfiguration} response from an API endpoint. This caller will call
+     * through to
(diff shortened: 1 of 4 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/callers/GetRequestCaller.java
```diff
@@ -86,6 +86,27 @@
...
 
+    /**
+     * Used in association with
+     * {@link VimeoClient#getContent(String, CacheControl, Caller, String, Map, String, VimeoCallback)} or
+     * {@link VimeoClient#getContentSync(String, CacheControl, String, Map, String, Caller)}
+     * to get an {@link com.vimeo.networking2.AppConfiguration} response from an API endpoint. This caller will call
+     * through to
+     * {@link VimeoService#getAppConfigurationMoshi(String, String, Map, String)} (String, String, Map, String)}
+     */
+    public static final Caller<com.vimeo.networking2.AppConfiguration> APP_CONFIGURATION_MOSHI =
```
msya: I had created the class `MoshiGetRequestCaller` to hold all Moshi related callers. Should this be moved into that class? Does it cause any complications in the integration?

anthonycr:
@msya back to you. Good catch, I was not aware of `MoshiGetRequestCaller`, and have moved the `APP_CONFIGURATION_MOSHI` caller to the correct class.

PR #348, opened 2018-11-09
Create internal annotation for models

msya:
- Created `Internal` annotation for models to specify internal models and fields.
- Added internal annotation at specific fields and models.
- Created models for search.
- Removed unused models.


models/src/main/java/com/vimeo/networking2/Subscriptions.kt
```diff
@@ -0,0 +1,56 @@
+package com.vimeo.networking2
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * A model that holds the type of push subscriptions a user has. This class
+ * has both getters and setters since a user can update their subscriptions
```
anthonycr: docs on this class are incorrect as they references setters which this data class does not have.

models/src/main/java/com/vimeo/networking2/Subscriptions.kt
```diff
@@ -0,0 +1,56 @@
...
+ * using a Patch.
+ */
+@JsonClass(generateAdapter = true)
+data class Subscriptions(
```
anthonycr: This class seems to be a duplicate of `NotificationTypeCount`. Is it different?

models/src/main/java/com/vimeo/networking2/Transcode.kt
```diff
@@ -1,15 +0,0 @@
-package com.vimeo.networking2
-
-import com.vimeo.networking2.enums.TranscodeType
-
-/**
- * The transcode information for a video upload.
- */
-data class Transcode(
```
anthonycr: What is the reason for getting rid of this class? It seems like it was created in part to replace the deprecated `status` property. The replacement is the `upload` property, which does not contain any information about transcoding, which is why the `transcode` property exists on videos as far as I can tell.

models/src/main/java/com/vimeo/networking2/VideoFile.kt
```diff
@@ -27,6 +26,7 @@ data class VideoFile(
     /**
      * The URL for logging events.
      */
+    @Internal
```
anthonycr: are the link and linkExpirationTime accessible? the previous comment was at the class level while this annotation is applied to the property level.

models/src/main/java/com/vimeo/networking2/annotations/Internal.kt
```diff
@@ -0,0 +1,9 @@
+package com.vimeo.networking2.annotations
+
+/**
+ * Annotation to describe certain models as internal.
+ */
+@Target(AnnotationTarget.CLASS, AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.SOURCE)
```
anthonycr: The retention should be `RUNTIME` instead as `SOURCE` results in the annotation not being removed at compile time. You can see this in action if you decompile one of the annotated kotlin classes to java. When using `SOURCE`, the annotation is no longer visible in the class, while when using `RUNTIME` it becomes visible. The main purpose of this annotation is documentation and it therefore should be outputted to the class file.

models/src/main/java/com/vimeo/networking2/Subscriptions.kt
msya: `NotificationTypeCount` has fields which are `int`. Each field represents the number of notifications you have by type. It is part of the `Notification` model. `Subscriptions` has fields which are booleans. Each fields represents whether you have the notification turned on or off. They are distinct models.

models/src/main/java/com/vimeo/networking2/Transcode.kt
msya: I had removed it, because it didn't exists in the current `Video` model. I added it back in.

models/src/main/java/com/vimeo/networking2/VideoFile.kt
msya: There was an issue in the OpenAPI spec. The `VideoFile` object requires a capability. I will annotate the class as internal.

msya:
@anthonycr I made the changes. Back to you.

models/src/main/java/com/vimeo/networking2/Video.kt
```diff
@@ -200,7 +192,6 @@ data class Video(
     /**
      * The transcode information for a video upload.
      */
-    @Json(name = "transcode")
```
anthonycr: The annotation should be added back as well.

models/src/main/java/com/vimeo/networking2/Live.kt
```diff
@@ -69,9 +69,8 @@ data class Live(
 
     /**
      * If [status] is [LiveStatus.STREAMING_ERROR], this is the reason for that error.
-     *
-     * Requires [CapabilitiesType.CAPABILITY_LIVE_EVENT].
      */
+    @Internal
```
anthonycr: if the class itself is annotated `@Internal` should we also be annotating the properties of the class?

models/src/main/java/com/vimeo/networking2/SearchResultList.kt
```diff
@@ -10,6 +10,12 @@ import com.vimeo.networking2.common.Pageable
 @JsonClass(generateAdapter = true)
 data class SearchResultList(
 
+    @Json(name = "facets")
+    val facetCollection: SearchFacetCollection? = null,
```
anthonycr: should there be docs on these properties

models/src/main/java/com/vimeo/networking2/annotations/Internal.kt
```diff
@@ -0,0 +1,9 @@
+package com.vimeo.networking2.annotations
+
+/**
+ * Annotation to describe certain models as internal.
```
anthonycr: "models or properties."

Also, might be helpful to elaborate on what internal means... internal to the library or internal to vimeo? Obviously we know what they mean, but consumer might not understand. I'd suggest using the phrasing "...as being for internal Vimeo use only."

models/src/main/java/com/vimeo/networking2/Live.kt
msya: Yeah. I update the PR to mark the fields internal as well.

msya:
@anthonycr I made the changes. Back to you.

PR #350, opened 2018-11-14
Kotlin Models with Moshi

msya:
This PR merges into develop the `models` module where all the DTOs are data classes in Kotlin. 

These are the following new features for models:

1. Moshi is used for serialization for the new models.
2. All models are immutable and have nullable fields.
3. Models and fields specific to Vimeo are specified with an `Internal`  annotation.
4. All connections and interactions have specific types. Only connections and interactions specific to an entity is exposed.


anthonycr (approved):
Code has already gone through extensive review when merged into branch v2.0.0, so giving the 👍

anthonycr:
We should wait to merge before https://github.com/vimeo/vimeo-networking-java/pull/346 is merged as it contains a number of fixes.

PR #358, opened 2018-12-04
Support new API version 3.4.2

brentwatson:
#### Issue
VA-3735

#### Summary
Bump API version from `3.4.1` -> `3.4.2`

Per API Changelog:

## 3.4.2 (2018-06-02)
* `account` in the `User` representation has been deprecated for `membership.type`.
* `badge` in the `User` representation has been deprecated for `membership.badge`.

#### Added

- The `User` representation has added the following fields:
    - `membership`
    - `membership.badge`
    - `membership.badge.alt_text`
    - `membership.badge.text`
    - `membership.badge.type`
    - `membership.badge.url`
    - `membership.display`
    - `membership.subscription`
    - `membership.subscription.renewal`
    - `membership.subscription.renewal.display_date`
    - `membership.subscription.trial`
    - `membership.subscription.trial.status`
    - `membership.type`

#### Removed

- The `User` representation has removed the following fields:
    - `account`
    - `badge`
    - `badge.alt_text`
    - `badge.text`
    - `badge.type`
    - `badge.url`
    

#### How to Test
Test apps should compile and run.
Vimeo App should correctly show badges and membership information after this change.


models/src/main/java/com/vimeo/networking2/Membership.kt
```diff
@@ -0,0 +1,43 @@
...
+     * Information about the user's subscription.
+     */
+    @Json(name = "subscription")
+    val subscription: Subscription? = null
+)
+
+/**
+ * @see AccountType
+ */
+val Membership.accountType: AccountType
```
anthonycr: Glad to see the extension pattern being put to use. In order to follow the "type" pattern we've used elsewhere, a few minor changes need to be made:
- `type` should be renamed to `rawType`
- `accountType` should be renamed to `type`
- `AccountType` should be renamed to `MembershipType`
- this extension function should also have a `@see` pointing back to the raw string property
- the raw string property should have an `@see pointing to the extension property

models/src/main/java/com/vimeo/networking2/Subscription.kt
```diff
@@ -0,0 +1,17 @@
...
+import com.squareup.moshi.JsonClass
+
+/**
+ * Information about the user's subscription.
+ */
+@JsonClass(generateAdapter = true)
+data class Subscription(
+
+    @Json(name = "renewal")
+    val renewal: SubscriptionRenewal,
```
anthonycr: properties could have Kdocs (or they could be written as param docs)

anthonycr: should have a default value of `null`

models/src/main/java/com/vimeo/networking2/SubscriptionRenewal.kt
```diff
@@ -0,0 +1,19 @@
...
+ */
+@JsonClass(generateAdapter = true)
+data class SubscriptionRenewal(
+
+    /**
+     * The date in YYYY-MM-DD format when the user's membership renews (or expires, if they have
+     * disabled autorenew). For display only.
+     */
+    @Json(name = "display_date")
+    val displayDate: String
```
anthonycr: should have a default value of null

models/src/main/java/com/vimeo/networking2/SubscriptionTrial.kt
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * The status of the user's trial.
+     * If the value is "free_trial" the user is currently in a free trial.
+     */
+    @Json(name = "status")
+    var status: String
+
+)
+
+val SubscriptionTrial.trialStatus: TrialType
```
anthonycr: Comments on the `Membership.accountType` extension also apply here
- `status` -> `trialStatus`
- `SubscriptionTrial.trialStatus` -> `SubscriptionTrial.trialStatusType`
- `TrialType` -> `TrialStatusType`

vimeo-networking/src/main/java/com/vimeo/networking/model/Membership.java
```diff
@@ -0,0 +1,95 @@
...
+@SuppressWarnings("unused")
+public class Membership implements Serializable {
+
+    private static final long serialVersionUID = 2005778103318600521L;
+
+    /**
+     * The user's membership level
+     */
+    @SerializedName("display")
+    private String mDisplay;
```
anthonycr: missing nullability annotation

vimeo-networking/src/main/java/com/vimeo/networking/model/Membership.java
```diff
@@ -0,0 +1,95 @@
...
+import com.google.gson.annotations.SerializedName;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.io.Serializable;
+
+/**
+ * This object contains user membership information.
+ * <p>
+ * Created by brentwatson on 12/3/18.
```
anthonycr: unnecessary created by, also applies to other files.

models/src/main/java/com/vimeo/networking2/Membership.kt
brentwatson: ✔️ I did 1,2,4,5
Regarding:
> AccountType should be renamed to MembershipType

I think "account type" is used most places instead of "membership type", so even though this is within a `membership.type` field I think it makes sense to stick with the naming used by the API, in our docs, etc.

Even the official API docs for the field is:
> description: "The user's account type:... "

We could also choose to keep the extension property as `accountType` based on this. Let me know what you think.

models/src/main/java/com/vimeo/networking2/enums/TrialType.kt
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Status types for a user's trial.
+ */
+enum class TrialType(override val value: String?) : StringValue {
+
+    /**
+     * Unknown trial type.
+     */
+    UNKNOWN(null),
```
msya: This should be the list last in the enum to consistency follow the setup in other DTOs.

models/src/main/java/com/vimeo/networking2/Subscription.kt
brentwatson: - Good point on using param docs. The other Kotlin DTOs use inline kdocs but these could all be changed to `@param` docs. We should probably stay consistent. I'll add as inline docs for now and will cc @msya on this one to consider future direction of these docs.

- All the fields that don't have a default of null are not nullable (per the API specs). Are we treating **everything** as nullable regardless?

vimeo-networking/src/main/java/com/vimeo/networking/model/Membership.java
brentwatson: This field is not-nullable (per API docs). So I was following convention here. Seems like Nullable fields have `@Nullable` and other fields have no annotations.  If you add `@NotNull` then you need a constructor to enforce that fact otherwise you get IDE lint errors.

vimeo-networking/src/main/java/com/vimeo/networking/model/Membership.java
brentwatson: Again, I was trying to follow convention. All Java DTOs have this but Kotlin DTOs do not, so I did the same.

models/src/main/java/com/vimeo/networking2/Membership.kt
```diff
@@ -0,0 +1,43 @@
...
+     * The user's account type.
+     */
+    @Json(name = "type")
+    val type: String,
+
+    /**
+     * Information about the user's badge.
+     */
+    @Json(name = "badge")
+    val badge: UserBadge? = null,
```
msya: As the field requires a capability, it should be annotated as internal.

models/src/main/java/com/vimeo/networking2/SubscriptionRenewal.kt
msya: This should be a date type.

models/src/main/java/com/vimeo/networking2/Membership.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Information about the user's subscription.
+     */
+    @Json(name = "subscription")
+    val subscription: Subscription? = null
```
msya: Based on the spec, this field is required. Should it be non null?

models/src/main/java/com/vimeo/networking2/SubscriptionRenewal.kt
brentwatson: Based on the name and description of this field I'm thinking this should maybe stay a string...? It is apparently "for display only." Might also be an issue to actually use this field because it is formatted into a String by the backend PHP code so we lose timezone information meaning the data in the field could actually be incorrect.

models/src/main/java/com/vimeo/networking2/Membership.kt
brentwatson: I went with nullable here because the spec is:
> `* @api-data membership.subscription (object, optional) - Information about the user's subscription.`

models/src/main/java/com/vimeo/networking2/Subscription.kt
anthonycr: Elsewhere in the library, we have always defaulted everything to nullable. The reason is that providing a field filter that excludes a property marked as non null will result in a parsing error. Considering that field filters can be injected by external consumers of the library, properties will need to be nullable. If we control the field filters internally in the library for a certain DTO then we could theoretically make the properties non null. I'd like @msya to weigh in on his thoughts as designer of the DTOs.

msya:
@brentwatson One more thing, the test should also be updated for the new models. The list of models in the tests should be updated to include Membership, SubscriptionRenwal, etc...

PR #360, opened 2018-12-07
Trial Eligibility DTOs

brentwatson:
#### Issue
VA-3736

#### Summary
New DTOs used for Trial Eligibility requests (internal).

Also added missing copyright notices from #358 and Unit tests.

#### How to Test
Run unit tests.

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/TrialEligibility.kt
+++ b/models/src/main/java/com/vimeo/networking2/TrialEligibility.kt
@@ -2,16 +2,19 @@ package com.vimeo.networking2
 
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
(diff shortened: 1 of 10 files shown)


models/src/main/java/com/vimeo/networking2/TrialEligibility.kt
```diff
@@ -0,0 +1,17 @@
+package com.vimeo.networking2
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Trial account eligibility.
+ */
+@JsonClass(generateAdapter = true)
+data class TrialEligibility(
```
msya: The class and the field `eligible` should be marked `@Internal`.

PR #362, opened 2018-12-11
Create Auth & Api Modules

msya:
This task creates the skeleton for the authentication work. The auth and api module have been created. The auth module depends on models and the api module. A `ServerConfig` class was also created for the necessary config needed to implement authentication. There are more configurations in the original config class. However, they will be added as in the future. A temporary `Authenticator` class was created in the auth module. All auth request will be implemented in it in further PRs.


api/src/main/java/com/vimeo/networking2/ServerConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Configuration for authentication and making requests to the Vimeo API.
+ *
+ * @param clientId              Your client id provided from https://developer.vimeo.com/apps/.
+ * @param clientSecret          Your client secret for authentication provided from https://developer.vimeo.com/apps/.
+ * @param baseUrl               The base url for all requests to the Vimeo API. This is can be overridden to test against a staging sever.
+ * @param scopes                A list of your scopes. See https://developer.vimeo.com/api/authentication#scopes.
+ * @param certPinningEnabled    Enable certificate pining. It is disabled by default.
+ * @param timeout               Read and the connection timeout for a request. The default time is 60 seconds.
+ * @param interceptors          Interceptors to add to for requests.
```
brentwatson: Might be a bit more dev-friendly to link to `[Interceptor]` in these docs.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -0,0 +1,7 @@
+package com.vimeo.networking2
+
+/**
+ * Authentication with email, google, facebook or pincode.
+ * This is a temporary class that will be updated as requests are added.
+ */
+class Authenticator(val serverConfig: ServerConfig)
```
brentwatson: Include `@param` doc.

models/src/main/java/com/vimeo/networking2/enums/ScopeType.kt
```diff
@@ -0,0 +1,20 @@
+package com.vimeo.networking2.enums
+
+/**
+ * All scopes that can be passed to the API. They determine what information you
+ * could get and your permissions in the Vimeo API.
```
brentwatson: Might be helpful to point to `https://developer.vimeo.com/api/authentication#scopes` in these docs so devs can go do more reading around "Enables your app to... "

msya:
@brentwatson I made the code review changes. I extracted the publishing to a separate gradle file. Here is the log from JitPack where it uses `piublishToMavenLocal` https://jitpack.io/com/github/vimeo/vimeo-networking-java/4259d763c5/build.log. It created artifacts for models, api and auth.

msya:
@brentwatson Back to you.

PR #366, opened 2019-01-04
Client Credentials Request

msya:
This PR implements the auth request `/oauth/authorize/client`. 

Here is the implementation:

#### Authenticator

Authenticator is an interface the client will use to perform any type of authentication. Basic authentication with a client id and client secret can be done by using the `clientCredentials` method  in the interface.

```
interface Authenticator {

    fun clientCredentials(authCallback: AuthCallback)

}
```

#### Creating Authenticator

```
interface Authenticator {

    fun clientCredentials(authCallback: AuthCallback)

    companion Factory {
 
         fun create(serverConfig: ServerConfig)

    }
}
```

The `Authenticator` interface contains a factory to create an instance of it by supplying a `ServerConfig` object.

#### VimeoCallAdapter

If an error occurs, the API will send an error code and developer message. The information needs to be parsed into an `ApiError` object. This call adapter was created in order to do this.

#### Usage

This is how it looks from a client's perspective.

```
            val serverConfig = ServerConfig(clientId, clientSecret, scopes)
            val authenticator = Authenticator.Factory.create(serverConfig)

           authenticator.clientCredentials(object : AuthCallback {

                override fun onSuccess(authResponse: ApiResponse.Success<String>) {
                    
                }

                override fun onGenericError(genericFailure: ApiResponse.Failure.GenericFailure) {
                    
                }

                override fun onApiError(apiFailure: ApiResponse.Failure.ApiFailure) {
                    
                }

                override fun onExceptionError(exceptionFailure: ApiResponse.Failure.ExceptionFailure) {
                    
                }
            })

```

`VimeoAccount` is not provided to the user only the access token is provided. This is because that's only important needed to make subsequent requests. Other info on the `VimeoAcccount` could be retrieved from the API such as the user object.


api/src/main/java/com/vimeo/networking2/requests/ApiRequestRetrofitCallback.kt
```diff
@@ -0,0 +1,34 @@
...
+class ApiRequestRetrofitCallback<ResponseType_T>(
+    private val apiResponseErrorConverter: ApiResponseErrorConverter,
+    private val apiResponseCallback: ApiResponseCallback<ResponseType_T>
+) : Callback<ResponseType_T> {
+
+    override fun onResponse(call: Call<ResponseType_T>, response: Response<ResponseType_T>) {
+        if (response.isSuccessful) {
+            response.body()?.let {
+                apiResponseCallback.onSuccess(it)
+            } ?: apiResponseCallback.onError(ApiError(errorMessage = "Failed to parse data."))
```
brentwatson: Should this error message be that body is null instead of assuming that parsing failed?

api/src/main/java/com/vimeo/networking2/requests/ApiResponseCallback.kt
```diff
@@ -0,0 +1,15 @@
+package com.vimeo.networking2.requests
+
+import com.vimeo.networking2.ApiError
+
+/**
+ * Informs you of the result of an API request. Provides the response or
```
brentwatson: minor / optional: docs don't need to wrap.

api/src/main/java/com/vimeo/networking2/utils/ApiResponseErrorConverter.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.vimeo.networking2.ApiError
+import retrofit2.Response
+import retrofit2.Retrofit
+
+/**
+ * Converter for parsing an error from the API into a [ApiError] object.
+ *
+ * @param retrofit Required for using the response body converter to do the conversion.
+ */
+class ApiResponseErrorConverter(private val retrofit: Retrofit) {
```
brentwatson: I'd suggest you extract an interface for this class and change `ApiRequestRetrofitCallback` and `Authenticator` to take the interface. That allows those classes to be more easily tested.

api/src/main/java/com/vimeo/networking2/utils/ApiResponseErrorConverter.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    fun <T> getErrorFromResponse(response: Response<T>): ApiError {
+        return try {
+            val errorConverter = retrofit.responseBodyConverter<ApiError>(
+                ApiError::class.java,
+                arrayOfNulls<Annotation>(0)
+            )
+            response.errorBody()?.let { errorConverter.convert(it) } ?: ApiError()
+        } catch (e: Exception) {
+            ApiError()
```
brentwatson: Should we do anything else here? Maybe log the exception or put something from the exception into `ApiError` so that it is surfaced?

api/src/main/java/com/vimeo/networking2/utils/ApiResponseErrorConverter.kt
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @param retrofit Required for using the response body converter to do the conversion.
+ */
+class ApiResponseErrorConverter(private val retrofit: Retrofit) {
+
+    fun <T> getErrorFromResponse(response: Response<T>): ApiError {
+        return try {
+            val errorConverter = retrofit.responseBodyConverter<ApiError>(
+                ApiError::class.java,
+                arrayOfNulls<Annotation>(0)
```
brentwatson: optional: this can just be `emptyArray()`.

api/src/main/java/com/vimeo/networking2/utils/ApiResponseErrorConverter.kt
```diff
@@ -0,0 +1,26 @@
...
+            )
+            response.errorBody()?.let { errorConverter.convert(it) } ?: ApiError()
```
brentwatson: Do we want to include that the fact that the error body is blank somewhere in `ApiError`?

Both of the empty ApiError instances should probably at least include the http error code.

brentwatson: Optional: you could use the new `ifEmpty { }` lambda here to provide a default value instead of using the elvis operator.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -31,31 +29,23 @@ class Authenticator(private val serverConfig: ServerConfig) {
      */
     private var retrofitServicesCache: RetrofitServicesCache
 
+    private var apiErrorResponseErrorConverter: ApiResponseErrorConverter
```
brentwatson: Other 2 member vars have kdocs. Maybe add the for this var too.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -31,31 +29,23 @@ class Authenticator(private val serverConfig: ServerConfig) {
...
+
     init {
         retrofit = RetrofitSetupComponent.retrofit(serverConfig)
         retrofitServicesCache = RetrofitSetupComponent.retrofitCache(retrofit)
+        apiErrorResponseErrorConverter = RetrofitSetupComponent.apiResponseErrorConverter(retrofit)
```
brentwatson: I don't think the `init` block is actually needed. I think you could simply init this in-line.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -31,31 +29,23 @@ class Authenticator(private val serverConfig: ServerConfig) {
...
-     * This is a temporary request that is added in to show how the access token will be set.
-     * This will be removed in the next PR which is to design and implement all the auth requests.
-     */
-    fun authenticateWithClientCredentials() {
-        val apiService = retrofitServicesCache.getService(AuthService::class.java)
-
-        val call = apiService.authorizeWithClientCredentialsGrant(
+    fun clientCredentials(): ClientCredentialsAuthenticator {
+        return ClientCredentialsInteractor(
+            apiErrorResponseErrorConverter,
```
brentwatson: Optional: I suggest added named args here. Makes it a bit nicer to read.

auth/src/main/java/com/vimeo/networking2/requests/clientcredentials/ClientCredentialsInteractor.kt
```diff
@@ -0,0 +1,52 @@
...
+ * @param scopes                    Comma separated list of scopes.
+ * @param setAuthTokenCallback      Callback to inform you that the authentication was successful
+ *                                  and the access token can be set for every request.
+ */
+class ClientCredentialsInteractor(
+    private val apiResponseErrorConverter: ApiResponseErrorConverter,
+    private val authService: AuthService,
+    private val authHeaders: String,
+    private val scopes: String,
+    private val setAuthTokenCallback: (String) -> Unit
```
brentwatson: This val should probably just be named `authTokenCallback` not `setAuthTokenCallback`.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -31,31 +29,23 @@ class Authenticator(private val serverConfig: ServerConfig) {
...
-
-            override fun onResponse(call: Call<VimeoAccount>, response: Response<VimeoAccount>) {
-                response.body()?.accessToken?.let { setAccessToken(it) }
-            }
-
-            override fun onFailure(call: Call<VimeoAccount>, t: Throwable) {}
-        })
+            serverConfig.scopes.joinToString { it.value }
+        ) {
+            setAccessToken(it)
```
brentwatson: Optional (and maybe a personal pref) but I find prefer using last-arg lambads like this when something is being executed and we're expecting a response.

eg:
```kotlin
makeNetworkRequest(scope) { response ->
 ...
}
```

But for constructors it's weird because in a case like this where the lambda executed after a call to `authenticate`, and only on success, this makes it so that it's not obvious what the lambda is doing or when it's going to be called, so a dev needs to dig into the implementation to find out.

simple eg:
```kotlin
NetworkRequestor(scope, url, blah, blah) {
  8080
}
```
here you would need to go into the impl to find that this is some callback to get a port number, why it's needed, and when it's used.

If you agree with this, I'd suggest adding this lambda to the params instead of hanging it off the end.

i.e.:
```kotlin
return ClientCredentialsInteractor(
            apiResponseErrorConverter = apiErrorResponseErrorConverter,
            authService = retrofitServicesCache.getService(AuthService::class.java),
            authHeaders = RetrofitSetupComponent.authHeaders(serverConfig.clientId, serverConfig.clientSecret),
            scopes = serverConfig.scopes.joinToString { it.value },
            authTokenCallback = { setAccessToken(it) }
        )
```

... to me this makes things more clear when & why it will be used.

auth/src/main/java/com/vimeo/networking2/requests/AuthCallback.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.vimeo.networking2.ApiError
+
+/**
+ * Inform you of the result of an authentication request. The access token is provided upon
+ * success and errors are propagated to you through the callback.
+ */
+interface AuthCallback {
+
+    /**
+     * The authentication was successful and the access token
```
brentwatson: nit: can be 1 line.

auth/src/main/java/com/vimeo/networking2/requests/clientcredentials/ClientCredentialsInteractor.kt
```diff
@@ -0,0 +1,52 @@
...
+): ClientCredentialsAuthenticator {
+
+    override fun authenticate(authCallback: AuthCallback) {
+        val call = authService.authorizeWithClientCredentialsGrant(
+            authHeaders,
+            GrantType.CLIENT_CREDENTIALS.value,
+            scopes)
```
brentwatson: If I run our formatter on this file I get 4 changes.

auth/src/main/java/com/vimeo/networking2/requests/clientcredentials/ClientCredentialsInteractor.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        val apiResponseCallback = object : ApiResponseCallback<VimeoAccount> {
+
+            override fun onSuccess(response: VimeoAccount) {
+                response.accessToken?.let {
```
brentwatson: Do we want to call `onError` if the `accessToken` is null? Currently this would no-op.

api/src/main/java/com/vimeo/networking2/requests/ApiRequestRetrofitCallback.kt
msya: I updated it to "Response body is null" to be more specific.

api/src/main/java/com/vimeo/networking2/utils/ApiResponseErrorConverter.kt
msya: I added the exception message to `ApiError`.

msya:
@brentwatson @anthonycr I made the code review changes. I implemented a pattern to handle cancelling and making requests. The error response is now converted to a model in a different thread. Back to you.

api/src/main/java/com/vimeo/networking2/utils/CallExtensions.kt
```diff
@@ -0,0 +1,29 @@
...
+import retrofit2.Response
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+/**
+ * This is an extensions on a call to make an API in a suspending
+ * function. When the request comes back, the method will resume
+ * the suspending method.
+ */
+suspend fun <T : Any> Call<T>.awaitResponse(): Response<T> =
```
anthonycr: since this requires coroutines, should it be in an extension module (we can come up with a name for the module) rather than the main api module

auth/src/main/java/com/vimeo/networking2/requests/clientcredentials/ClientCredentialsInteractor.kt
```diff
@@ -0,0 +1,90 @@
...
+ * @param apiScopes                 List of [ScopeType]s.
+ * @param authTokenCallback         Callback to inform you that the authentication was successful
+ *                                  and the access token can be set for every request.
+ */
+class ClientCredentialsInteractor(
+    private val errorResponseConverter: ErrorResponseConverter,
+    private val authService: AuthService,
+    private val authHeaders: String,
+    private val apiScopes: List<ScopeType>,
+    private val coroutineScope: CoroutineScope,
```
anthonycr: Will java consumers be able to use this interactor? Perhaps we should have a separate `auth-coroutines` module

auth/src/main/java/com/vimeo/networking2/requests/clientcredentials/ClientCredentialsAuthenticator.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Authenticate client id and client secret.
+ */
+interface ClientCredentialsAuthenticator {
+
+    /**
+     * Authenticate client id and client secret.
+     *
+     * @param authCallback informs you of the result of the response.
+     */
+    fun authenticate(authCallback: AuthCallback)
```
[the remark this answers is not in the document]
anthonycr: Should the `authenticate` function return an `ApiRequest` rather than the whole authenticator itself implementing it?

api/src/main/java/com/vimeo/networking2/requests/ApiRequestRetrofitCallback.kt
```diff
@@ -0,0 +1,39 @@
...
+import retrofit2.Callback
+import retrofit2.Response
+
+/**
+ * Custom Retrofit [Callback] to listen for requests and parse any errors.
+ *
+ * @param errorResponseConverter Used to convert an error in the response to [ApiError].
+ * @param apiResponseCallback       Used to notify you of the response or error.
+ */
+class ApiRequestRetrofitCallback<ResponseType_T>(
```
anthonycr: this class is not used anywhere, will it be used in a later implementation?

api/src/main/java/com/vimeo/networking2/requests/ApiResponse.kt
```diff
@@ -0,0 +1,40 @@
...
+ * [Failure] class will inform you of an api error.
+ */
+sealed class ApiResponse<out T> {
+
+    /**
+     * A successful response.
+     *
+     * @param accessToken The accessToken parsed by this successful response.
+     */
+    data class Success<out T>(val accessToken: T) : ApiResponse<T>()
```
anthonycr: this looks like a generic response enumeration, but the property is called `accessToken`, should it be called `data` or something similar?

api/src/main/java/com/vimeo/networking2/requests/ApiResponseCallback.kt
```diff
@@ -0,0 +1,14 @@
+package com.vimeo.networking2.requests
+
+import com.vimeo.networking2.ApiError
+
+/**
+ * Informs you of the result of an API request. Provides the response or any errors.
+ */
+interface ApiResponseCallback<ResponseType_T> {
```
anthonycr: Should this be using the enumerated `ApiResponse` instead of having singular success and error methods?

msya:
@anthonycr I made the changes as we did in person. Back to you.

api/src/main/java/com/vimeo/networking2/adapters/ErrorHandlingCallAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+import retrofit2.CallAdapter
+import retrofit2.Converter
+import java.lang.reflect.Type
+import java.util.concurrent.Executor
+
+/**
+ * Custom call adapter to handle errors.
+ *
+ * @param responseType              Type of the response.
+ * @param callbackExecutor          The executor upon which the request is made.
```
anthonycr: Is this documentation correct? Seems like it should be the executor on which the caller is notified.

api/src/main/java/com/vimeo/networking2/adapters/ErrorHandlingCallAdapterFactory.kt
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Factory for creating a custom [ErrorHandlingCallAdapter].
+ */
+class ErrorHandlingCallAdapterFactory: CallAdapter.Factory() {
+
+    /**
+     * Creates and returns a [ErrorHandlingCallAdapter].
+     */
+    override fun get(returnType: Type,
```
anthonycr: minor formatting: the first parameter should be moved to a new line

api/src/main/java/com/vimeo/networking2/adapters/ErrorHandlingCallAdapterFactory.kt
```diff
@@ -0,0 +1,35 @@
...
+        if (returnType !is ParameterizedType) {
+            throw IllegalStateException("VimeoCall must have generic type (e.g., VimeoCall<ResponseBody>)")
+        }
+        val responseType = CallAdapter.Factory.getParameterUpperBound(0, returnType)
+        val callbackExecutor = retrofit.callbackExecutor()
+        val errorResponseConverter = retrofit.responseBodyConverter<ApiError>(
+            ApiError::class.java,
+            emptyArray()
+        )
+        return ErrorHandlingCallAdapter<Any>(responseType, callbackExecutor!!, errorResponseConverter)
```
anthonycr: I can't remember if we discussed this and maybe it's guaranteed non null elsewhere in this PR, but could it be null?

api/src/main/java/com/vimeo/networking2/adapters/VimeoCallAdapter.kt
```diff
@@ -0,0 +1,76 @@
...
+                sendResponse { callback.onExceptionError(t) }
+            }
+        })
+    }
+
+    /**
+     * Send response on [callbackExecutor].
+     */
+    fun sendResponse(action: () -> Unit) {
+        callbackExecutor.execute { action() }
```
anthonycr: `callbackExecutor.execute(action)` results in cleaner bytecode as it allows the `action` to be wrapped in a static `Runnable` instead of the anonymous instance that results from the current code. Going further, inlining `sendResponse eliminates the extra class creation all together.

api/src/main/java/com/vimeo/networking2/adapters/VimeoCallAdapter.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Determine is the response has a body.
+     */
+    fun Response<T>.hasBody() = isSuccessful && body() != null
```
anthonycr: can be private

api/src/main/java/com/vimeo/networking2/adapters/VimeoCallAdapter.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Parse the error body into a [ApiError] object.
+     */
+    fun Response<T>.parseApiError() = errorBody()?.let { responseBodyConverter.convert(it) }
```
anthonycr: can be private

api/src/main/java/com/vimeo/networking2/adapters/VimeoCallback.kt
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * An error occurred on the API. [apiError] contains information about the error that occurred.
+     */
+    fun onApiError(apiError: ApiError)
+
+    /**
+     * Generic error occurred. This occurred was not caused by the API rather it was due to
+     * parsing the data or no response body was null.
+     */
+    fun onGenericError(requestCode: Int)
```
anthonycr: Should the parameter name be responseCode?

auth/src/main/java/com/vimeo/networking2/requests/AuthService.kt
```diff
@@ -17,7 +17,7 @@ interface AuthService {
      * and scope types.
      *
      * @param authHeader It is created from the client id and client secret.
-     * @param grantType  Determines whether you have access to public or private data.
+     * @param grantType  Determines whether you have access to public or private accessToken.
```
anthonycr: This change looks to be the result of a broken rename and should be reverted.

api/src/main/java/com/vimeo/networking2/requests/ApiResponse.kt
anthonycr: bump

auth/src/main/java/com/vimeo/networking2/utils/AuthExtensions.kt
```diff
@@ -0,0 +1,47 @@
+package com.vimeo.networking2.utils
```
anthonycr: Should this file utilize `@file:JvmName` for better java interop or are we okay with the name being `AuthExtensions`?

msya: This extension will only be used internally by the SDK. Therefore, I didn't add `@file:JvmName`. If public extensions are added to this class later on, then I could create a public extensions file and add `@file:JvmName`.

msya:
@anthonycr Thanks for the review. Back to you.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -1,73 +1,68 @@
...
+ * - Facebook
+ * - Email login
+ * - Email join
+ * - Logout
+ *
+ * Create an instance of the Authenticator by using its factory and make requests as follows.
+ *
+ * Ex:
+ *
+ * val authenticator = Authenticator.create(serverConfig)
```
anthonycr: could surround this codeblock with ` ``` ` to get it formatted as code

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -1,73 +1,68 @@
...
      */
-    private var retrofit: Retrofit
+    fun clientCredentials(authCallback: AuthCallback): VimeoRequest
 
     /**
-     * Used to store created retrofit services.
+     * Factory to create an instance of [Authenticator].
      */
-    private var retrofitServicesCache: RetrofitServicesCache
+    companion object Factory {
```
anthonycr: elegant use of the companion object 💃

api/src/main/java/com/vimeo/networking2/adapters/VimeoCallback.kt
```diff
@@ -0,0 +1,40 @@
+package com.vimeo.networking2.adapters
```
anthonycr: is the `adapters` package the correct package for this? could it be moved into the parent package as that will be more accessible for consumers. perhaps we could follow what some other libraries do where they have an `internal` package where we could put the internal classes in the adapter. happy to talk offline about the package structure.

auth/src/main/java/com/vimeo/networking2/requests/interactors/ClientCredentialsInteractor.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.vimeo.networking2.utils.enqueueAuthRequest
+
+/**
+ * Interactor which performs auth request with client credentials.
+ *
+ * @param authService   Retrofit API service for auth.
+ * @param authHeaders   Client id and client secret headers.
+ * @param scopes        API scopes to support.
+ */
+internal class ClientCredentialsInteractor(
```
anthonycr: this class is so small, and it's also internal - could the `AuthenticatorImpl` just do what this class is doing? There's almost as much setup of the `ClientCredentialsInteractor` required as there is just to talk to the `AuthService` directly.

PR #367, opened 2019-01-04
Updating retrofit to version 2.5.0

anthonycr:
# Summary
We want to update retrofit to the latest version, as the current version being used is almost 3 years old and is lacking features that we want to use downstream of this library. We cannot just override the version used here downstream, as there are some API changes that seemed to cause build failures, so we are upgrading the library to 2.5.0.

Nothing in the library needed to be changed except for unit tests. Those tests:
- `RetrofitSetupTest`
- `BaseUrlInterceptorTest`

See the changelog for retrofit: https://github.com/square/retrofit/blob/master/CHANGELOG.md

#### How to Test
- Run tests
- Verify that the library works correctly with the new version.

### DIFF
--- a/vimeo-networking/build.gradle
+++ b/vimeo-networking/build.gradle
@@ -38,7 +38,7 @@ dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
     compile project(':models')
 
-    def retrofitVersion = '2.1.0'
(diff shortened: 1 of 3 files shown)


anthonycr:
@msya adding you as a secondary reviewer

PR #371, opened 2019-01-29
Facebook and Google Login

msya:
In this PR, Google and Facebook login are implemented. 

### Implementation
- Created a `SignInInteractor` to handle making the requests.
- Created a `SignInAuthParams` data class. It has an extension to convert the data class to a map.
- Updated `AuthService` to include info on performing a POST request to `/users`.
- Created the interface `SignInAuthenticator` for the consumer to provide `SignInAuthParams` and a callback to get a response back.


api/src/main/java/com/vimeo/networking2/internal/VimeoCallAdapter.kt
```diff
@@ -26,38 +25,40 @@ internal class VimeoCallAdapter<T>(
...
      * the response synchronously on the background thread.
      */
-    private fun sendResponse(action: () -> Unit) {
-        if (callbackExecutor != null) {
-            callbackExecutor.execute(action)
-        } else {
-            action()
-        }
+    private fun Executor?.sendResponse(action: () -> Unit) {
+        this?.let { execute(action) } ?: action()
```
anthonycr: can be simplified to `this?.execute(action) ?: action()`

api/src/main/java/com/vimeo/networking2/VimeoRequest.kt
```diff
@@ -11,3 +14,24 @@ interface VimeoRequest {
     fun cancel()
 
 }
+
+/**
+ * No-op API request. Used when an invalid param was given by the client. In this case,
+ * the actual API request is not made. So, there is nothing to cancel.
+ */
+class NoOpVimeoRequest: VimeoRequest {
```
anthonycr: If this is made to be an `object` then we don't need to allocate a new one every time it is used.

auth/src/main/java/com/vimeo/networking2/internal/AuthUtils.kt
```diff
@@ -36,11 +32,34 @@ internal fun VimeoCall<VimeoAccount>.enqueueAuthRequest(authCallback: AuthCallba
...
+
+/**
+ * Validates any authentication params given by the client.
+ *
+ * @return a list of invalid parameters.
+ */
+internal fun Map<AuthParam, String>.validate(): List<InvalidParameter> =
+        filter { it.value.isEmpty() }
+        .map { InvalidParameter(it.key.name, it.key.errorCode?.value, it.key.developerMessage) }
+        .toList()
```
anthonycr: `toList()` is unnecessary here.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -27,4 +25,42 @@ internal class AuthenticatorImpl(
...
+            AuthParam.FIELD_EMAIL to email,
+            AuthParam.FIELD_MARKETING_OPT_IN to marketingOptIn.toString()
+        )
+        val call = authService.join(authHeaders, params)
+
+        val invalidAuthParams = params.validate()
+
+        return if (invalidAuthParams.isNotEmpty()) {
+            val apiError = ApiError(
+                "Google authentication error.",
```
anthonycr: Error message is specific when function is generically used.

PR #372, opened 2019-02-05
Email Join and Login

msya:
#### Summary
- Added the to join and login with email and password.
- Added the ability to send a custom user agent for every request.
- Updated `VimeoCallback` to return instances of a sealed class for success and error responses.


api/src/main/java/com/vimeo/networking2/config/RetrofitSetupModule.kt
```diff
@@ -17,47 +17,19 @@ import java.util.concurrent.TimeUnit
  */
 object RetrofitSetupModule {
 
-    private const val USER_AGENT_HEADER = "User-Agent"
     private const val AUTHORIZATION_HEADER = "Authorization"
+    private const val USER_AGENT_HEADER = "User-Agent"
+    private const val USER_AGENT_HEADER_VALUE = "Kotlin VimeoNetworking/$SDK_VERSION"
```
anthonycr: The current user agent is `VimeoNetworking/version (Java)`. Would it be a good idea to keep the same pattern and change this to `VimeoNetworking/$SDK_VERSION (Kotlin)`?

api/src/main/java/com/vimeo/networking2/config/RetrofitSetupModule.kt
```diff
@@ -76,36 +48,54 @@ object RetrofitSetupModule {
...
      */
-    fun retrofitCache(retrofit: Retrofit) = RetrofitServicesCache(retrofit)
+    private fun userAgentInterceptor(userAgent: String?) =
+        Interceptor { chain ->
+            val request = chain.request()
+            val builder = request.newBuilder()
+
+            if (request.header(USER_AGENT_HEADER).isNullOrBlank()) {
+
+                val customUserAgent = userAgent?.let {
```
anthonycr: The user agent should be additive to the `USER_AGENT_HEADER_VALUE` rather than completely replacing it. See how `RetrofitSetup.createUserAgent` works in the previous version of the library.

PR #373, opened 2019-02-06
File share DTO

anthonycr:
# Summary
We need to add internal support for the file transfer page. The link to the transfer page is held by the transfer page object. The API adds the information to the video as `file_transfer.link`, so I have created the appropriate `FileTransferPage` DTO which mimics the way the `ReviewPage` DTO works.

I created a legacy DTO in Java in `vimeo-networking` and a new DTO in Kotlin in `models`.


anthonycr:
Verified client side, merging.

PR #374, opened 2019-02-07
Update Facebook token field name

msya:
Facebook authentication requires the Facebook token name to be `token` whereas the token name for Google should be `id_token`. I created an enum `SocialAuthType` to determine which token param field name to use in the `socialAuthenticate` method.

### DIFF
--- a/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
+++ b/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
@@ -40,14 +40,32 @@ internal class AuthenticatorImpl(
         email: String,
         marketingOptIn: Boolean,
         authCallback: AuthCallback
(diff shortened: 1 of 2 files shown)


auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -58,12 +58,22 @@ internal class AuthenticatorImpl(
         token: String,
         email: String,
         marketingOptIn: Boolean,
-        authenticationErrorMessage: String,
+        socialAuthType: SocialAuthType,
         authCallback: AuthCallback
     ): VimeoRequest {
 
+        val tokenField = when(socialAuthType) {
```
anthonycr: This code could be simplified if instead of passing an `enum` into this function, we passed a `data class` that encapsulated the dynamic properties. e.g.
```kotlin
data class SocialAuthRequestParams(
  val tokenFieldName: String,
  val errorMessage: String
)
```
Or these parameters could be added to the `socialAuthenticate` declaration itself.

msya:
@anthonycr Back to you.

PR #375, opened 2019-02-07
Create Basic Auth and Authentication data classes

msya:
#### Summary

In this PR, I created the data classes `BasicAuthToken` and `AuthenticatedToken`. 

BasicAuthToken
- Returned by making a client credentials request

AuthenticatedToken
- Returned by making a Google, Facebook or email login request.
- Contains an access token and a user object.

AccessTokenProvider 
- Supplies the access token.
- Both of the data classes above inherit from this interface. 

The `AuthCallback` was removed as it was returning the access token itself. The client could pass a `VimeoCallback` into the API request methods.


models/src/main/java/com/vimeo/networking2/AuthenticatedAccessToken.kt
```diff
@@ -0,0 +1,20 @@
+package com.vimeo.networking2
+
+import com.squareup.moshi.Json
+
+/**
+ * Authenticate token that you would receive from email, Google or Facebook login. In addition to
+ * the token, you get a user object representing the authenticated user.
+ */
+data class AuthenticatedAccessToken(
```
anthonycr: Great idea 👍

PR #379, opened 2019-02-22
VA-3792 Album editing

howardrigberg:
#### Issue
https://github.com/vimeo//vimeo-networking-java/issues/VA-3792

#### Summary
Support the ability for users to add, delete, and edit albums. Album editing includes changing title, description, and privacy as well as adding and removing videos. 

#### How to Test
Tests can be performed on the application level.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -29,6 +29,8 @@ import com.vimeo.networking.callbacks.VimeoCallback;
 import com.vimeo.networking.callers.GetRequestCaller;
 import com.vimeo.networking.interceptors.LanguageHeaderInterceptor;
 import com.vimeo.networking.logging.ClientLogger;
+import com.vimeo.networking.model.Album;
+import com.vimeo.networking.model.AlbumPrivacy;
 import com.vimeo.networking.model.Comment;
 import com.vimeo.networking.model.Document;
 import com.vimeo.networking.model.PictureCollection;
@@ -321,7 +323,7 @@ public class VimeoClient {
         final String state = queryMap.get(Vimeo.CODE_GRANT_STATE);
 
         if (code == null || code.isEmpty() || state == null || state.isEmpty() ||
-                !state.equals(mCurrentCodeGrantState)) {
+            !state.equals(mCurrentCodeGrantState)) {
             mCurrentCodeGrantState = null;
 
             callback.failure(new VimeoError("Code grant code/state is null or state has changed"));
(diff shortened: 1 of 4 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -946,10 +948,140 @@ public void failure(VimeoError error) {
...
+     * @param parameters   Additional less common parameters that modify the album.
+     * @param callback     A callback that will receive the results of the network request.
+     * @return A Call object or null if a client-side initialization error has occurred. In the event of an error,
+     * the callback will still be notified.
+     */
+    @Nullable
+    public Call<Album> createAlbum(@NotNull final String name,
+                                   @NotNull final AlbumPrivacy albumPrivacy,
+                                   @Nullable final String description,
+                                   @Nullable final Map<String, Object> parameters,
```
msya: It will be easier to know which extra params could be passed in if it was associated with a type. For example, 

```
class AlbumParams {
    
       final Map<String, Object> parameters;

       public void reviewMode(boolean reviewMode) {
            parameters.put("review_mode", reviewMode);
      }
 
      ...

     public Map<String, Object> build() {
           return params;
      }
}
```

The benefit of this approach is that the client doesn't have look at the API docs to determine what extra params could be sent.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -946,10 +948,140 @@ public void failure(VimeoError error) {
...
+                                   @NotNull final VimeoCallback<Album> callback) {
+        if (VimeoNetworkUtil.validateString(name, "Name can't be empty in createAlbum.", callback)) {
+            final Map<String, Object> validParameters =
+                    VimeoNetworkUtil.prepareAlbumEditParameters(name, albumPrivacy, description, parameters, callback);
```
msya: The `prepareAlbumEditParameters` has multiple responsibilities. It is creating a map but it is also validating whether the client provided a password. The validation could be moved outside of the method and the method could be renamed to `createAlbumEditParamsMap` to signify what it is doing.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -946,10 +948,140 @@ public void failure(VimeoError error) {
...
+     * @param albumPrivacy A representation of the viewing privacy for the album. The expectation is that if the
+     *                     album is password protected that this object will also contain the required password.
+     * @param description  A nullable description for the album.
+     * @param parameters   Additional less common parameters that modify the album.
+     * @param callback     A callback that will receive the results of the network request.
+     * @return A Call object or null if a client-side initialization error has occurred. In the event of an error,
+     * the callback will still be notified.
+     */
+    @Nullable
+    public Call<Album> editAlbum(@NotNull final Album album,
```
msya: The same suggestions above apply to edit, delete and add to album methods.

PR #381, opened 2019-02-27
Implement certificate ssl pinning 

msya:
#### Summary

This PR implements certificate pinning for the new networking library. OkHttp 3 provides a way to specify a public key for a url pattern when setting up OkHttpClient. `RetrofitSetupModule` now has a method to do this. The `ServerConfig` object has a flag which is set to true by default to determine whether to setup SSL pinning or not. This is flag is checked in the module when setting up certificate pinning. 

This PR also fixes the issue that implementation and api were not being recognized by api, auth, models and vimeo-networking modules. This is because they were using the java gradle plugin. The java-library plugin does know about how to handle implementation vs api. They were updated in the specific gradle files.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -88,7 +88,7 @@ platform :android do
     # run the tests, but dont puke everywhere until the report is printed.
     begin
       gradle(
-        task: "vimeo-networking:test example-java-android:build example-kotlin-android:build models:build api:build auth:build"
+        task: "vimeo-networking:test example-java-android:build example-kotlin-android:build models:test api:test auth:test"
       )
     rescue => error
       begin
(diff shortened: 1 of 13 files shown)


fastlane/Fastfile
```diff
@@ -88,7 +88,7 @@ platform :android do
     # run the tests, but dont puke everywhere until the report is printed.
     begin
       gradle(
-        task: "vimeo-networking:test example-java-android:build example-kotlin-android:build models:build"
+        task: "vimeo-networking:test example-java-android:build example-kotlin-android:build models:build api:build auth:build"
```
anthonycr: should probably use `test` instead of `build` at least for the non-example modules. I don't know that we have any tests yet, but we should future proof it so that we don't check in failing tests.

Issue #385, opened 2019-04-03
JsonSyntaxException: Expected a string but was BEGIN_OBJECT at line 1 column 2 path $


ricardopereira:
#### Issue Summary

I want to know the link of a Video ID that I have in the database but an error occurs when deserializing the object.

I'm using the latest stable version (`implementation "com.vimeo.networking:vimeo-networking:1.1.3"`). It's an Android project with Kotlin.

#### Reproduction Steps

Just try to fetch the content of a Video like:

``` kotlin
VimeoClient.getInstance().fetchNetworkContent("/videos/"+id, object : ModelCallback<Video>(Video::class.java) {

    override fun success(video: Video) {
        onSuccess(video, videoInfo, directory, fetch)
    }

    override fun failure(error: VimeoError) {
        Timber.e(error.exception,"Vimeo download error occurred ${error.developerMessage}")
    }

})
```

#### Expected Behavior

A deserialized `Video` object where I can use it and access the video link.

#### Actual Behavior

A `com.google.gson.JsonSyntaxException` exception occurs:

``` console
I/System.out: --------- RESPONSE ---------
I/System.out: ENDPOINT: /videos/<redacted>
    STATUS CODE: 200
I/System.out: REQUEST TIME: 1322.8ms
I/System.out: --------- RESPONSE END ---------
I/System.out: Error when deserializing object!
W/System.err: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 1 column 2 path $
        at com.google.gson.Gson.fromJson(Gson.java:900)
W/System.err:     at com.google.gson.Gson.fromJson(Gson.java:853)
        at com.google.gson.Gson.fromJson(Gson.java:802)
        at com.google.gson.Gson.fromJson(Gson.java:774)
        at com.vimeo.networking.GsonDeserializer.deserializeObject(GsonDeserializer.java:66)
        at com.vimeo.networking.GsonDeserializer.deserialize(GsonDeserializer.java:58)
        at com.vimeo.networking.VimeoClient$5.success(VimeoClient.java:1169)
        at com.vimeo.networking.callbacks.VimeoCallback.onResponse(VimeoCallback.java:55)
        at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:70)
        at android.os.Handler.handleCallback(Handler.java:873)
[...]

PR #386, opened 2019-04-03
Fix Gson Deserializer Null Crash

msya:
#### Summary

The builders below are setting the deserialization field to null.  Even though the deserialization field was set to default value, it was being overridden to null by the constructors. This problem was brought up in the  issue https://github.com/vimeo/vimeo-networking-java/issues/385.

```
 public Builder(String clientID, String clientSecret, String scope) {
            this(null, clientID, clientSecret, scope, null, null);
        }

        @Deprecated
        public Builder(String baseURLString, String clientID, String clientSecret, String scope) {
            this(baseURLString, clientID, clientSecret, scope, null, null);
        }

        public Builder(String clientId, String clientSecret, String scope,
                       @Nullable AccountStore accountStore, @Nullable GsonDeserializer deserializer) {
            this(null, clientId, clientSecret, scope, accountStore, deserializer);
        }
```

To resolve this issue, I added a null check in the constructor and initialize it when a non null deserializer is given. Otherwise, I set a default value for it in the constructor.

#### How to Test

Configure VimeoClient with a client id and client secret and verify the following request succeeds and parses:

```
 VimeoClient.getInstance().fetchNetworkContent("/videos/"+id, object : ModelCallback<Video>(Video::class.java) {

            override fun success(video: Video) {
                println("video = [${video}]")
            }

            override fun failure(error: VimeoError) {
                println("error = [${error}]")
            }

        })
```

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/Configuration.java
@@ -26,6 +26,7 @@ import com.vimeo.networking.Vimeo.LogLevel;
 import com.vimeo.networking.logging.LogProvider;
 import com.vimeo.networking.model.VimeoAccount;
 
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


anthonycr (approved):
LGTM. We discussed offline that this fixes a separate issue than the one posted.

PR #389, opened 2019-05-28
Downgrade Moshi

msya:
Version 1.8.0 of Moshi is causing proguard to run indefinitely. This PR downgrades the version to 1.7.0. This change was tested in the app.

### DIFF
--- a/api-core/build.gradle
+++ b/api-core/build.gradle
@@ -19,7 +19,4 @@ dependencies {
     // Moshi Adapters
     implementation "com.squareup.moshi:moshi-adapters:1.7.0"
 
-    // Okio used by Moshi
-    implementation "com.squareup.okio:okio:2.2.2"
-
 }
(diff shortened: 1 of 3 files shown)


api-core/build.gradle
```diff
@@ -17,9 +17,6 @@ dependencies {
     implementation "com.squareup.retrofit2:converter-moshi:2.5.0"
 
     // Moshi Adapters
-    implementation "com.squareup.moshi:moshi-adapters:1.8.0"
-
-    // Okio used by Moshi
-    implementation "com.squareup.okio:okio:2.2.2"
```
anthonycr: Didn't we recently run into a crash which required us to specify the okio type? https://github.com/vimeo/vimeo-networking-java/pull/388

msya: I added the okio dependency back.

anthonycr: It looks like it was only added to the `api-core` module, and was still removed from the other modules. Should it be added to all modules like it was before?

PR #390, opened 2019-05-31
VA-3896 showcase edit support

howardrigberg:
#### Issue
https://github.com/vimeo//vimeo-networking-java/issues/VA-3896

#### Summary
Videos will have:
`metadata.connections.available_channels`
`metadata.connections.available_albums`

These connections will have a uri like follows:
`/videos/:video_id/available_channels/videos/:video_id/available_albums`

The response from this endpoint provides all of the logged in user's albums/channels (paginated). Each album/channel object in the list will contain:
`metadata.interactions.add_to`

There will be a `metadata.interactions.add_to.uri` for the interaction to add/remove. The uri will be the same in both cases. In the event that the video is not yet added to the album/channel, metadata.interactions.add_to.options = PUT. In the event that the video is already added to the album/channel, metadata.interactions.add_to.options = DELETE.
```
GET /videos/123

{
"uri": "/videos/123",
"metadata": {
"connections": {
"available_albums":

{ 
"uri": "/videos/123/available_albums", 
"options": [ "GET" ], 
"total": 2 
}

}
}
}

GET /videos/123/available_albums

{
"uri": "/albums/100",
"metadata": {
"interactions": {
"add_to":
{ 
"uri": "/albums/100/videos/123", "options": [ "PUT" ] }
},
{
"uri": "/albums/200",
"metadata": {
"interactions": {
"add_to":
{ 
"uri": "/albums/200/videos/123", 
"options": [ "DELETE" ] 
}
}
```
Additionally, we will need the ability to get a list of a user's videos for a specific album where the response will be a list of videos that have interactions similar to those described above for albums where the there is an "album" interaction with a uri and a "PUT" or "DELETE" option depending on whether the video is already in the album. The specs for this endpoint are described here: 
https://vimean.atlassian.net/browse/COL-1061

Batch adding of videos to an album is currently supported by the API via a patch request to the album.metadata.interactions.add_videos.uri as follows:

```
PATCH /users/173181/albums/5260655/videos
{
“remove”: [
{“video”: {“uri”: “/videos/223022"}},
{“video”: {“uri”: “/videos/254785”}}
],
“set”: [
{“video”: {“uri”: “/videos/261621”, “position”: 2}},
{“video”: {“uri”: “/videos/345195”, “position”: 1}}
]
}
```

Batch adding of a video to multiple albums is still undefined at the time of writing and will need to be implemented as part of this ticket: https://vimean.atlassian.net/browse/VA-4037

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -33,6 +33,7 @@ import com.vimeo.networking.model.Album;
 import com.vimeo.networking.model.AlbumPrivacy;
 import com.vimeo.networking.model.Comment;
 import com.vimeo.networking.model.Document;
+import com.vimeo.networking.model.ModifyVideosInAlbum;
 import com.vimeo.networking.model.PictureCollection;
 import com.vimeo.networking.model.PictureResource;
 import com.vimeo.networking.model.PinCodeInfo;
@@ -40,6 +41,7 @@ import com.vimeo.networking.model.Privacy;
 import com.vimeo.networking.model.TextTrackList;
 import com.vimeo.networking.model.User;
 import com.vimeo.networking.model.Video;
+import com.vimeo.networking.model.VideoList;
 import com.vimeo.networking.model.VimeoAccount;
 import com.vimeo.networking.model.error.ErrorCode;
 import com.vimeo.networking.model.error.LocalErrorCode;
@@ -1081,6 +1083,23 @@ public class VimeoClient {
         call.enqueue(callback);
         return call;
     }
+
+    @Nullable
+    public Call modifyVideosInAlbum(@NotNull final Album album,
+                                    @NotNull final ModifyVideosInAlbum modificationSpecs,
+                                    @NotNull VimeoCallback<VideoList> callback) {
+        if (!VimeoNetworkUtil.validateString(album.getUri(),
+                                             "Album uri cannot be empty in modifyVideosInAlbum",
+                                             callback)) {
+            return null;
+        }
+        final Call<VideoList> call = mVimeoService.modifyVideosInAlbum(getAuthHeader(),
+                                                                       album.getUri() + "/videos",
+                                                                       modificationSpecs);
+        call.enqueue(callback);
+        return call;
+    }
+
 
     @Nullable
     public Call<Video> editVideo(@Nullable String uri,
--- a/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
@@ -33,7 +33,7 @@ import com.vimeo.networking.model.Album;
 import com.vimeo.networking.model.AlbumPrivacy;
 import com.vimeo.networking.model.Comment;
 import com.vimeo.networking.model.Document;
-import com.vimeo.networking.model.ModifyVideosInAlbum;
+import com.vimeo.networking.model.ModifyVideosInAlbumSpecs;
 import com.vimeo.networking.model.PictureCollection;
 import com.vimeo.networking.model.PictureResource;
 import com.vimeo.networking.model.PinCodeInfo;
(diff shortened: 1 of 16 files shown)


howardrigberg:
@msya, [this feature branch](https://github.vimeows.com/MobileApps/Vimeo-Android/tree/feature/VA-3791-Showcases-Phase-2) is currently pointing at the networking changes made in this PR. Some of the new models are being used (e.g. in the add videos to album screen for displaying available_videos with their selected states). Let me know if you want to test the batch add/remove videos to album and I can send you some testing code for that.

vimeo-networking/src/main/java/com/vimeo/networking/model/AddVideoToAlbum.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * An object that is used to patch video addition updates to an Album.
+ */
+@SuppressWarnings("unused")
+public class AddVideoToAlbum implements Serializable {
+
+    private static final long serialVersionUID = -855083653273120166L;
+
+    public AddVideoToAlbum(@NotNull String uri, int position) {
```
msya: Should this constructor be below the uri and position fields?

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
```diff
@@ -158,4 +162,13 @@ public void setIsAdded(boolean added) {
...
         return mReportingReasons;
     }
+
+    /**
+     * @return a {@code List<String>} containing the options that are available for the {@link #getUri()}
+     * endpoint in this {@link Interaction}. The options will be things like "PUT", "DELETE", etc.
+     */
+    @Nullable
+    public List<String> getOptions() {
+        return new ArrayList<String>(mOptions);
```
msya, line 172: In the new models, I had defined the options to be an enum `ApiOptionsType`. It's easier for the client if they wanted to check for an option. Should this also return an enum of options?

vimeo-networking/src/main/java/com/vimeo/networking/model/ModifyVideosInAlbumSpecs.java
```diff
@@ -0,0 +1,114 @@
...
+ * An object that is used to patch video addition and deletion updates to an Album.
+ */
+@SuppressWarnings("unused")
+public class ModifyVideosInAlbumSpecs implements Serializable {
+
+    private static final long serialVersionUID = -3094719083671086785L;
+
+    @Nullable
+    @SerializedName("remove")
+    private List<NamedWrapperForRemove> mRemoveVideoList;
```
msya: Should the videos lists be a `LinkedHashSet` to retain order of the insertions and prevent duplicates?

vimeo-networking/src/main/java/com/vimeo/networking/model/ModifyVideosInAlbumSpecs.java
```diff
@@ -0,0 +1,114 @@
...
+
+        /**
+         * @return The wrapped AddVideoToAlbum object.
+         */
+        AddVideoToAlbum getAddVideoToAlbum() {
+            return mAddVideoToAlbum;
+        }
+    }
+
+    private class NamedWrapperForRemove implements Serializable {
```
msya, line 98: Do these classes need to be annotated as `@UseStag(FieldOption.SERIALIZED_NAME)`?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1082,6 +1084,36 @@ public Call removeFromAlbum(@NotNull final Album album,
...
+     * @param callback          A callback to be notified when the album's video list is modified. A {@link VideoList}
+     *                          containing the modified list of videos in the album will be returned in the event of
+     *                          success.
+     * @return A Call object or null if a client-side initialization error has occurred. In the event of an error,
+     * the callback will still be notified.
+     */
+    @Nullable
+    public Call modifyVideosInAlbum(@NotNull final Album album,
+                                    @NotNull final ModifyVideosInAlbumSpecs modificationSpecs,
+                                    @NotNull VimeoCallback<VideoList> callback) {
```
msya, line 1103: The error code `3429 ` was added for albums. It could be added to `ErrorCode`.

msya (changes requested):
@howardrigberg I had some minor comments. Back to you.

vimeo-networking/src/main/java/com/vimeo/networking/model/Interaction.java
howardrigberg: That's a nice thought, but I wanted to keep this consistent with the `getOptions` that already exists in `Connection` for the old models.

vimeo-networking/src/main/java/com/vimeo/networking/model/ModifyVideosInAlbumSpecs.java
howardrigberg: They could be but that would require me to provide setters for the private members and expose private classes that it doesn't make sense to expose. In this case, I don't think we need the performance optimization since it is only used for making a patch to add videos to a user's album. Let me know if you disagree and we can talk more about it IRL.

howardrigberg:
Good suggestions, @msya. Thanks for the review. I've addressed everything in code or comments. Back to you.

msya (approved):
LGTM

PR #391, opened 2019-06-04
VimeoError API improvement

anthonycr:
# Summary
Previously, `VimeoError` did not supply enough information about the cause of the error. Specifically, if an error occurred because there was no internet connection, the only information that could be used to determine if it was caused by lack of internet was checking the developer message if it matched a certain pattern "Unable to resolve host...". Checking the developer readable string message is not a reliable way of checking the cause within the program, and instead it would be better to check the type. So, I changed the internal `mException` property in `VimeoError` to `mThrowable`, and now within the `VimeoCallback`, the throwable cause is set along with the developer message. This will enable downstream consumers to more accurately determine what caused their issue.

Additionally, I removed usage of `printStackTrace` and replaced it with `ClientLogger.e`, except in `VimeoCallback` where the print was removed instead of replaced since we were already notifying the consumer.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/error/VimeoError.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/error/VimeoError.java
@@ -74,7 +74,7 @@ public class VimeoError extends RuntimeException {
     @SerializedName("invalid_parameters")
     protected List<InvalidParameter> mInvalidParameters;
 
-    private Exception mException;
+    private Throwable mThrowable;
     private int mHttpStatusCode = Vimeo.NOT_FOUND;
 
     private boolean mIsCanceledError;
@@ -86,9 +86,9 @@ public class VimeoError extends RuntimeException {
         this.mDeveloperMessage = errorMessage;
     }
 
-    public VimeoError(String errorMessage, Exception exception) {
+    public VimeoError(String errorMessage, Throwable throwable) {
         this.mDeveloperMessage = errorMessage;
-        this.mException = exception;
+        this.mThrowable = throwable;
     }
 
     public Response getResponse() {
@@ -204,12 +204,12 @@ public class VimeoError extends RuntimeException {
         return getInvalidParameter() != null ? getInvalidParameter().getErrorCode() : null;
     }
 
-    public Exception getException() {
-        return mException;
+    public Throwable getThrowable() {
+        return mThrowable;
     }
 
-    public void setException(Exception exception) {
-        this.mException = exception;
+    public void setThrowable(Throwable throwable) {
+        this.mThrowable = throwable;
     }
 
     public int getHttpStatusCode() {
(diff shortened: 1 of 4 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
```diff
@@ -199,8 +200,7 @@ public void run() {
                         .responseBodyConverter(VimeoError.class, new Annotation[0]);
                 vimeoError = errorConverter.convert(response.errorBody());
             } catch (final Exception e) {
-                //noinspection CallToPrintStackTrace
-                e.printStackTrace();
+                ClientLogger.e("Error while attempting to convert response body to VimeoError", e);
```
msya, line 203: The methods `VimeoNetworkUtils.getSimpleQueryMap` and `VimeoNetworkUtils.getErrorFromResponse` could be updated to use the ClientLogger when an exception is caught.

vimeo-networking/src/main/java/com/vimeo/networking/model/error/VimeoError.java
```diff
@@ -204,12 +204,12 @@ public ErrorCode getInvalidParameterErrorCode() {
...
 
-    public Exception getException() {
-        return mException;
+    public Throwable getThrowable() {
+        return mThrowable;
     }
 
-    public void setException(Exception exception) {
-        this.mException = exception;
+    public void setThrowable(Throwable throwable) {
```
msya, line 211: The setters and getters for exception are used in multiple places in the app. Is there a corresponding PR to update the app?

msya (changes requested):
@anthonycr Minor questions. Back to you.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
anthonycr: That's what I did in this commit: https://github.com/vimeo/vimeo-networking-java/pull/391/commits/497dfb8e6c193957aff488e477c3939379648e98, or are you referring to something else?

vimeo-networking/src/main/java/com/vimeo/networking/model/error/VimeoError.java
anthonycr: I created a ticket for it, and I will tag you on it.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
msya: oh. Sorry.  I got mixed with something else.

msya (approved):
LGTM

PR #392, opened 2019-06-10
Add max-age to cache header

msya:
## Summary

API no longer includes a max age in the API response. Therefore, OkHttp no longer reads from the cache. It determines it to be invalid. This PR sets a max age of 7 hours in the cache header for every request. 

## Test

In the sample app of the SDK, modify the stack picks request to pull from cache and another request to pull from the network. Get data from the network. Verify the request is cached by looking at the data folder of the app. Get data from the cache. Verify a 504 exception is not thrown and you get data from the cache.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/interceptors/CacheControlInterceptor.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/interceptors/CacheControlInterceptor.java
@@ -33,8 +33,8 @@ import okhttp3.Response;
 
 /**
  * Rewrite the server's cache-control header because our server sets all {@code Cache-Control} headers
- * to {@code no-store}
- * Created by brentwatson on 8/7/17.
+ * to {@code no-store}.
+ *
  */
 public class CacheControlInterceptor implements Interceptor {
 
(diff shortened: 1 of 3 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/interceptors/CacheControlInterceptor.java
```diff
@@ -33,16 +33,16 @@
...
+ *
  */
 public class CacheControlInterceptor implements Interceptor {
 
     @Override
     public Response intercept(Chain chain) throws IOException {
         return chain.proceed(chain.request())
                 .newBuilder()
-                .header(Vimeo.HEADER_CACHE_CONTROL, Vimeo.HEADER_CACHE_PUBLIC)
+                .header(Vimeo.HEADER_CACHE_CONTROL, Vimeo.HEADER_CACHE_PUBLIC_WITH_MAX_AGE)
```
anthonycr: Rather than hardcoding the cache age, should we be using the value available on the configuration object via `Configuration.getCacheMaxAge()`?

msya: Yes. I made the change.

anthonycr (approved):
LGTM

PR #394, opened 2019-07-01
Fix Cache Control Header Formatting

msya:
The cache control header interceptor adds a max age. However, the formatting of the directive was incorrect. It should be `max-age=XXXX`.


rpassis:
LGTM 👍

rpassis (approved):

PR #395, opened 2019-07-17
Add Enterprise account type

msya:
Added enterprise account type enum value to user object.


vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
```diff
@@ -69,7 +70,8 @@
         LIVE_PRO,
         LIVE_PREMIUM,
         PRO_UNLIMITED,
-        PRODUCER
+        PRODUCER,
+        ENTERPRISE
```
anthonycr, line 74: Enterprise should also be added to `networking2.enums.AccountType`.

anthonycr (changes requested):
Networking V2 should also be updated.

msya:
@anthonycr Back to you.

anthonycr (approved):
LGTM

PR #396, opened 2019-08-15
VA-4137 Adding LiveQuota Status

howardrigberg:
#### Issue
https://vimean.atlassian.net/browse/VA-4137

#### Summary
Added a status to the LiveQuota object

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/LiveQuota.kt
+++ b/models/src/main/java/com/vimeo/networking2/LiveQuota.kt
@@ -5,6 +5,8 @@ package com.vimeo.networking2
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 import com.vimeo.networking2.annotations.Internal
+import com.vimeo.networking2.enums.LiveQuotaStatusType
+import com.vimeo.networking2.enums.asEnum
 
 /**
  * Live Quota information.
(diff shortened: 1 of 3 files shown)


models/src/main/java/com/vimeo/networking2/LiveQuota.kt
```diff
@@ -13,18 +13,25 @@ import com.vimeo.networking2.annotations.Internal
...
+        @Internal
+        @Json(name = "time")
+        val time: LiveTime? = null,
+
+        /**
+         * Live status data.
+         */
+        @Internal
+        @Json(name = "status")
+        val status: String? = null
```
anthonycr: Design documents indicate that this is an enum with values
- `available`
- `private_mode`
- `stream_limit`
- `time_limit`

Based on other enums in the library, we should create an enum named `StatusType` or `LiveQuotaStatusType` with the above values. Then an extension function should be created in this file as

```kotlin
val LiveQuota.statusType: StatusType
  get() = status.asEnum(StatusType.UNKNOWN)
```

See `Live` or `Privacy` for examples of how this enum is implemented.

Additionally, the legacy vimeo-networking should be updated to include the enum as well.

anthonycr (changes requested):

models/src/main/java/com/vimeo/networking2/LiveQuota.kt
```diff
@@ -13,18 +13,25 @@ import com.vimeo.networking2.annotations.Internal
 @JsonClass(generateAdapter = true)
 data class LiveQuota(
 
-    /**
-     * Live streams quota data.
-     */
-    @Internal
-    @Json(name = "streams")
-    val streams: LiveStreamsQuota? = null,
+        /**
```
anthonycr: Do you have special formatting on? When I format this file it does not add extra indentation (and I don't think Mohit's formatter adds the indentation either since he wrote these files), I think your formatter might be incorrectly configured for this project so the indentation should be reverted.

howardrigberg:
Good call outs @anthonycr. I added the enums and fixed the formatting. Back to you.

anthonycr (approved):
LGTM

PR #401, opened 2019-11-27
Add Error mapping for INVALID_INPUT_PASSWORD_TOO_LONG

WhosNickDoglio:
#### Issue
[VA-4146](https://vimean.atlassian.net/browse/VA-4146)

#### Summary
Added missing error code to `ErrorCode` enum to make it easier to show the proper error message.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/error/ErrorCode.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/error/ErrorCode.java
@@ -154,6 +154,8 @@ public enum ErrorCode {
     INVALID_INPUT_EMAIL_NOT_RECOGNIZED,
     @SerializedName("2218")
     INVALID_INPUT_PASSWORD_EMAIL_MISMATCH,
+    @SerializedName("3102")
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/error/ErrorCode.java
```diff
@@ -217,6 +217,8 @@
     URL_UNAVAILABLE,
     @SerializedName("2406")
     USER_NOT_AUTHORIZED_TO_DELETE_ACCOUNT,
+    @SerializedName("3102")
+    INVALID_INPUT_PASSWORD_TOO_LONG,
```
kvenn: Based on the comments/organization in this file it looks like this might fit better around line [156](https://github.com/vimeo/vimeo-networking-java/blob/f92daa0a47efc50d4d7ddc5c2b5af5301facd92c/vimeo-networking/src/main/java/com/vimeo/networking/model/error/ErrorCode.java#L156). That section looks like it's Auth->Input which this seems to fall into.

kvenn (approved):
Looks ggggrrrrrrreat!

PR #406, opened 2020-01-15
VA-4204 publish api cleanup

howardrigberg:
#### Issue
https://github.com/vimeo//vimeo-networking-java/issues/VA-4204

#### Summary
Changing Video ID requirement to uri since Video objects do not explicitly expose the numeric ID needed. It is expected that `video.metadata.connections.publish_to_social.uri` will be used instead.


WhosNickDoglio (approved):
LGTM! 👍

PR #408, opened 2020-01-24
Added YouTube no channel blocker to both models

WhosNickDoglio:
#### Summary
Added `yt_no_channel` blocker(When the connected google account doesn't have a YouTube channel connected to it) to both model versions.


anthonycr (approved):
LGTM

PR #410, opened 2020-02-07
Fixed isStaff in User Object

WhosNickDoglio:
#### Summary

The `isStaff` in the `User` object was never actually being returned so I updated `getIsStaff` to properly check if the `User` is staff.

### DIFF
--- a/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
+++ b/vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
@@ -24,6 +24,7 @@ package com.vimeo.networking.model;
 
 import com.google.gson.annotations.SerializedName;
 import com.vimeo.networking.Vimeo;
+import com.vimeo.networking.model.UserBadge.UserBadgeType;
 import com.vimeo.networking.model.live.LiveQuota;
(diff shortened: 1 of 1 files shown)


vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
```diff
@@ -443,7 +447,7 @@ public String getId() {
      */
     @Nullable
     public Boolean getIsStaff() {
-        return mIsStaff;
+        return getBadge() != null && getBadge().getBadgeType() == UserBadgeType.STAFF;
```
howardrigberg, line 449: Remove the `@Nullable` and make the return type `boolean`.

howardrigberg (changes requested):

vimeo-networking/src/main/java/com/vimeo/networking/model/User.java
WhosNickDoglio: Fixed in https://github.com/vimeo/vimeo-networking-java/pull/410/commits/3f30862b439cb201f4986865d4347cd968db7770

howardrigberg (approved):

PR #411, opened 2020-02-12
Modifying required fields for publish to social posts

howardrigberg:
We changed the required fields for Facebook and YouTube publish to social API posts. This ticket is to expose the new requirements.

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/params/PublishToFacebookPost.kt
+++ b/models/src/main/java/com/vimeo/networking2/params/PublishToFacebookPost.kt
@@ -13,13 +13,13 @@ data class PublishToFacebookPost(
          * The title of the post as it will appear on Facebook.
          */
         @Json(name = "title")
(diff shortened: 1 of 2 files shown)


models/src/main/java/com/vimeo/networking2/params/PublishToFacebookPost.kt
```diff
@@ -13,13 +13,13 @@ data class PublishToFacebookPost(
          * The title of the post as it will appear on Facebook.
          */
         @Json(name = "title")
-        val title: String,
+        val title: String?,
```
anthonycr: We should follow the pattern used elsewhere in these param classes and provide default `null` values

anthonycr (changes requested):

howardrigberg:
@anthonycr Added null defaults. Back to you.

anthonycr (approved):
Looks good!

PR #412, opened 2020-02-12
Enable detekt and unify code style

anthonycr:
# Summary
We want to enable detekt on this project. We also want a unified code style. In order to do this, I enabled detekt at the CI level and also added a git commit hook that runs detekt optionally on commit. I also committed the project code style that lives in the `.idea` folder. This allows all contributors to use the same codestyle.

I fixed a number of formatting issues as well as issues detekt complained about.

## How to test
Let the tests (and detekt) run on CI.


api-core/src/main/java/com/vimeo/networking2/internal/ErrorHandlingCallAdapter.kt
```diff
@@ -15,7 +15,7 @@ import java.util.concurrent.Executor
  * @param callbackExecutor          The executor upon which the caller is notified.
  * @param responseBodyConverter     Converter to parse the error response into a [ApiError].
  */
-internal class ErrorHandlingCallAdapter<T>(
+internal class ErrorHandlingCallAdapter<T : Any>(
```
WhosNickDoglio, line 18: Just curious but what was the lint warning to add this? `Any` doesn't seem necessary to me here.

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallAdapter.kt
```diff
@@ -32,7 +32,7 @@ internal class VimeoCallAdapter<T>(
 
                 if (response.hasBody()) {
                     callbackExecutor.sendResponse {
-                        callback.onSuccess(VimeoResponse.Success(response.body()!!))
+                        callback.onSuccess(VimeoResponse.Success(requireNotNull(response.body())))
```
WhosNickDoglio, line 35: Bye bye double bang 👋

build.gradle
```diff
@@ -14,6 +12,17 @@ buildscript {
     }
 }
 
+plugins {
+    id "io.gitlab.arturbosch.detekt" version "1.0.0-RC14"
```
WhosNickDoglio, line 16: Any reason we're using this version instead of the latest version?

WhosNickDoglio (approved):
Overall everything looks good, just a few small questions. 👍

api-core/src/main/java/com/vimeo/networking2/internal/ErrorHandlingCallAdapter.kt
anthonycr: The upper bound on `T` in `requireNotNull` is `T : Any`, so I can't pass a `T` to it without getting a build error. Since we never expect a nullable type back as a result of a call (null is an error), we can safely set the upper bound to `T : Any`.

build.gradle
anthonycr: For the first integration, I wanted to stick with the old RC version we are using in other projects. Once we prove that it's working as expected, I want to upgrade to the latest version, as this version is a year old. I am ticketing the upgrade to the latest version.

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallAdapter.kt
howardrigberg: 👋 🔫 🔫

howardrigberg (approved):

PR #414, opened 2020-02-19
VA-4233: Added Page interface

WhosNickDoglio:
#### Issue
[VA-4233](https://vimean.atlassian.net/browse/VA-4233)

#### Summary
Added a `Page` interface to denote collections of data that do not support pagination and updated `Pageable` and `ConnectedAppList` to inherit from it.


anthonycr (approved):
LGTM!

PR #417, opened 2020-03-06
Added missing moshi annotations

WhosNickDoglio:
#### Summary
After bumping Moshi in the networking library to the latest it's required all models either be annotated with `@JsonClass` or the Moshi instance being used needs to be passed an instance of `KotlinJsonAdapterFactory`.  [See the Moshi changelog for more details](https://github.com/square/moshi/blob/master/CHANGELOG.md#version-190)

InvalidParameter was missing the `@JsonClass` annotation and caused the app the throw an `java.lang.IllegalArgumentException`

Adding the annotation fixes the crash.

Also added `@JsonClass` to `AuthenticatedAccessToken` as it was also missing it.


anthonycr (approved):
Looks good

howardrigberg (approved):

PR #418, opened 2020-03-18
Improve java compatibility

anthonycr:
# Summary
I noticed that there were several issues when trying to convert code from the old library to the new library. First, there are several classes that had issues being consumed from Java as they had `ClassNameKt` names for utilities instead of `ClassNameUtils`. Second, I found there was a missing property in the `InvalidParameter` DTO. I corrected both issues.


WhosNickDoglio (approved):
LGTM 👍

PR #422, opened 2020-03-30
Adding missing properties and fixing properties in V2 DTOs

anthonycr:
# Summary
A number of DTOs have issues, missing properties, incorrect property types, incorrect JSON key names. This PR fixes many of these issues.

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/Privacy.kt
+++ b/models/src/main/java/com/vimeo/networking2/Privacy.kt
@@ -4,6 +4,7 @@ package com.vimeo.networking2
 
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
+import com.vimeo.networking2.annotations.Internal
 import com.vimeo.networking2.enums.CommentPrivacyType
 import com.vimeo.networking2.enums.EmbedPrivacyType
 import com.vimeo.networking2.enums.ViewPrivacyType
@@ -26,6 +27,7 @@ data class Privacy(
      * The token used to authenticate in playback scenarios where password entry is impossible, and
      * the user initiating playback has already entered the password.
      */
+    @Internal
     @Json(name = "_bypass_token")
     val bypassToken: String? = null,
 
--- a/models/src/main/java/com/vimeo/networking2/Privacy.kt
+++ b/models/src/main/java/com/vimeo/networking2/Privacy.kt
@@ -43,6 +43,12 @@ data class Privacy(
     val embedPrivacy: String? = null,
 
     /**
+     * The password for viewing the authenticated user's videos.
+     */
+    @Json(name = "password")
+    val password: String? = null,
+
+    /**
      * The privacy settings of the channel.
      * @see Privacy.viewPrivacyType
      */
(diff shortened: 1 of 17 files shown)


models/src/main/java/com/vimeo/networking2/Privacy.kt
```diff
@@ -22,6 +22,13 @@ data class Privacy(
     @Json(name = "add")
     val add: Boolean? = null,
 
+    /**
+     * The token used to authenticate in playback scenarios where password entry is impossible, and
+     * the user initiating playback has already entered the password.
+     */
+    @Json(name = "_bypass_token")
```
howardrigberg, line 31: Is the leading underscore intentional? I realize it matches Privacy.java, but I just wanted to double check since the leading underscore is unorthodox.

vimeo-networking/src/main/java/com/vimeo/networking/utils/PrivacySettingsParams.java
```diff
@@ -10,7 +12,7 @@
 
 /**
  * Builder for setting privacy params for a video.
- *
+ * <p>
  * Created by Mohit Sarveiya on 3/20/18.
```
howardrigberg: Minor: Remove the author attributes.

models/src/main/java/com/vimeo/networking2/ProgressiveVideoFile.kt
```diff
@@ -61,7 +61,7 @@ data class ProgressiveVideoFile(
      * The file size of the video.
      */
     @Json(name = "size")
-    val size: Int? = null,
+    val size: Long? = null,
```
howardrigberg, line 64: Optional, you could add kdocs to the params in this class that are missing them.

models/src/main/java/com/vimeo/networking2/Upload.kt
```diff
@@ -40,7 +40,7 @@ data class Upload(
      */
     @Internal
```
howardrigberg, line 41: Are we still planning to use the `@Internal` annotation?

howardrigberg (changes requested):
Some very minor comments.

models/src/main/java/com/vimeo/networking2/Privacy.kt
anthonycr: Yes, it is intentional, and documented in the API blueprint. I'm not sure why there is a leading underscore in the property name, although I expect it means that the property is internal only. I've added the `@Internal` annotation to this property to be safe.

models/src/main/java/com/vimeo/networking2/ProgressiveVideoFile.kt
anthonycr: The only params that I see without docs are overridden properties, which we do not document since they inherit the docs from the interface they are overriding.

models/src/main/java/com/vimeo/networking2/Upload.kt
anthonycr: Yes

WhosNickDoglio (approved):

howardrigberg (approved):

PR #423, opened 2020-03-30
Support V2 DTOs in VimeoClient

anthonycr:
# Summary
This PR adds support for requesting the new DTOs using the `VimeoClient`. All usage of the V1 DTOs has been removed in the library, except that of `VimeoError`. Future work will change over to using the `VimeoResponse.Error` class, but for now, we will rely on `VimeoError` since the conversion is not entirely straight forward.

In addition to adding support for these new DTOs, a few changes were made to the V2 DTOs, particularly adding `Serializable` to DTOs that were missing it, adding DTOs and properties that were missing, and redesigning the `VideoFile` class so that it supports differentiation between HLS and DASH.

- [x] Change target branch to `develop` after https://github.com/vimeo/vimeo-networking-java/pull/422 is merged.


vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
```diff
@@ -197,10 +223,7 @@ public void run() {
         VimeoError vimeoError = null;
         if (response != null && response.errorBody() != null) {
             try {
-                final Converter<ResponseBody, VimeoError> errorConverter = VimeoClient.getInstance()
-                        .getRetrofit()
-                        .responseBodyConverter(VimeoError.class, new Annotation[0]);
-                vimeoError = errorConverter.convert(response.errorBody());
+                vimeoError = getGson().fromJson(response.errorBody().string(), VimeoError.class);
```
howardrigberg, line 226: I am seeing errors upon log out in the app. Could be related?
```
2020-03-30 15:16:38.712 18181-18181/com.vimeo.android.videoapp E/BaseStreamFragment BaseStreamFragment.onError (643): error in getFetchCallback - Retrofit - Authorization Required
    com.vimeo.networking.model.error.VimeoError
        at com.vimeo.networking.model.error.VimeoError$TypeAdapter.read(VimeoError$TypeAdapter.java:80)
        at com.vimeo.networking.model.error.VimeoError$TypeAdapter.read(VimeoError$TypeAdapter.java:13)
        at com.google.gson.Gson.fromJson(Gson.java:927)
        at com.google.gson.Gson.fromJson(Gson.java:892)
        at com.google.gson.Gson.fromJson(Gson.java:841)
        at com.google.gson.Gson.fromJson(Gson.java:813)
        at com.vimeo.networking.utils.VimeoNetworkUtil.getErrorFromResponse(VimeoNetworkUtil.java:226)
        at com.vimeo.networking.callbacks.VimeoCallback.onResponse(VimeoCallback.java:53)
        at retrofit2.DefaultCallAdapterFactory$ExecutorCallbackCall$1.lambda$onResponse$0$DefaultCallAdapterFactory$ExecutorCallbackCall$1(DefaultCallAdapterFactory.java:81)
        at retrofit2.-$$Lambda$DefaultCallAdapterFactory$ExecutorCallbackCall$1$3wC8FyV4pyjrzrYL5U0mlYiviZw.run(Unknown Source:6)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
```diff
@@ -90,10 +90,10 @@
     @NotNull
     private Converter.Factory chooseFactoryForSerializer(@NotNull final Annotation[] annotations) {
         final Serializer serializer = findSerializerInArray(annotations);
-        if (serializer != null && serializer.converter() == ConverterType.MOSHI) {
```
WhosNickDoglio, line 93: Do we still need this class? Don't all the new models use Moshi?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -247,7 +245,7 @@ public void setVimeoAccount(@Nullable VimeoAccount vimeoAccount) {
         if (vimeoAccount == null) {
             // If the provided account was null but we have an access token, persist the vimeo account with
             // just a token in it. Otherwise we'll want to leave the persisted account as null.
-            vimeoAccount = new VimeoAccount(mConfiguration.mAccessToken);
+            vimeoAccount = new VimeoAccount(mConfiguration.mAccessToken, null, null, null, null, null);
```
WhosNickDoglio: `@JvmOverloads` on a data class constructor? Good or bad idea?

Or something like `VimeoAccountFactory.createAccountWithAccessToken(mConfiguration.mAccessToken)` ?

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
```diff
@@ -86,6 +93,23 @@ public static Gson getGson() {
...
 
+    /**
+     * Static helper method that automatically applies the VimeoClient Moshi preferences
+     * <p>
+     * This includes formatting for dates, and any serializers used.
+     *
+     * @return Moshi object that can be used to serialize and deserialize JSON.
+     */
+    @NotNull
+    public static Moshi getMoshi() {
```
WhosNickDoglio, line 104: Shouldn't `RetrofitSetup` use this method?

models/src/main/java/com/vimeo/networking2/params/ModifyVideosInAlbumSpecs.kt
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * The set of videos that should be removed.
+     */
+    @Json(name = "remove")
+    val removeVideoSet: Set<NamedWrapperForRemove>? = null,
+
+    /**
+     * The set of videos that should be added.
+     */
+    @Json(name = "set")
```
WhosNickDoglio, line 20: Is this the correct JSON key? Just want to confirm since `ModifyVideoInAlbumsSpecs` has a similar property with the key as "add".

WhosNickDoglio (changes requested):
Looking good! Just a few suggestions and question. 

Also can we depreciate or just remove completely all the v1 models?

example-kotlin-android/src/main/kotlin/com/vimeo/android/networking/example/kotlin/AccountPreferenceManager.kt
```diff
@@ -3,9 +3,9 @@ package com.vimeo.android.networking.example.kotlin
 import android.content.Context
 import android.content.SharedPreferences
 import android.preference.PreferenceManager
-import com.vimeo.networking.model.User
-import com.vimeo.networking.model.VimeoAccount
 import com.vimeo.networking.utils.VimeoNetworkUtil
+import com.vimeo.networking2.User
+import com.vimeo.networking2.VimeoAccount
```
howardrigberg, line 8: Optional: there's a trivial lint error in this file on line 50 if you want to fix it.

example-kotlin-android/src/main/kotlin/com/vimeo/android/networking/example/kotlin/MainActivity.kt
```diff
@@ -10,10 +10,9 @@ import com.vimeo.networking.VimeoClient
 import com.vimeo.networking.callbacks.AuthCallback
 import com.vimeo.networking.callbacks.VimeoCallback
 import com.vimeo.networking.callers.GetRequestCaller
-import com.vimeo.networking.callers.MoshiGetRequestCaller
-import com.vimeo.networking.model.User
-import com.vimeo.networking.model.VideoList
 import com.vimeo.networking.model.error.VimeoError
+import com.vimeo.networking2.User
```
howardrigberg, line 14: Optional. Some lint warnings in this file if you want to fix them.

models/src/main/java/com/vimeo/networking2/params/ModifyVideosInAlbumSpecs.kt
anthonycr: I copied the original Java class, @howardrigberg maybe you can weigh in on why this is named `set` instead of `add`?

models/src/main/java/com/vimeo/networking2/DashVideoFile.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.vimeo.networking2.annotations.Internal
+import com.vimeo.networking2.common.VideoFile
+import java.io.Serializable
+import java.util.*
+
+/**
+ * A video file that represents a DASH stream.
+ */
+@JsonClass(generateAdapter = true)
+data class DashVideoFile(
```
howardrigberg, line 14: When are we planning to delete the java model equivalents?

vimeo-networking/src/main/java/com/vimeo/networking/AnnotatedConverterFactory.java
anthonycr: I'm going to remove all the V1 models and associated code in an upcoming ticket.

models/src/main/java/com/vimeo/networking2/Drm.kt
```diff
@@ -0,0 +1,25 @@
...
+ * library and only Apple products support `fairplay`, that type is omitted. Clients will only receive these if given
+ * the app-specific permission, essentially this class is not consumable by the general public.
+ */
+@Internal
+@JsonClass(generateAdapter = true)
+data class Drm(
+    /**
+     * The video file containing the info about the DRM protected stream.
+     */
+    @Json(name = "widevine")
```
howardrigberg, line 19: Do you know it is possible for us to have a problem where models get serialized with the old system and then deserialized with the new system upon update in the app?

models/src/main/java/com/vimeo/networking2/DashVideoFile.kt
howardrigberg, line 14: There is an inconsistency with respect to the `VideoFile` class. This class does not inherit from it even though it shares members while `HlsVideoFile` is derived from it.

models/src/main/java/com/vimeo/networking2/params/AddVideoToAlbum.kt
```diff
@@ -0,0 +1,22 @@
...
+     */
+    @Json(name = "uri")
+    val uri: String,
+
+    /**
+     * The position of the video.
+     */
+    @Json(name = "position")
+    val position: Int?
+)
```
howardrigberg, line 22: Should we add `Serializable` for the params?

models/src/main/java/com/vimeo/networking2/params/ModifyVideosInAlbumSpecs.kt
howardrigberg: It's how API decided to do it. I believe that I asked the question at the time, but I don't remember the rationale. 🤷‍♂

WhosNickDoglio (approved):

models/src/main/java/com/vimeo/networking2/params/ModifyVideosInAlbumSpecs.kt
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * An object that is used to patch video addition and deletion updates to an Album.
+ */
+@JsonClass(generateAdapter = true)
+data class ModifyVideosInAlbumSpecs(
+    /**
+     * The set of videos that should be removed.
+     */
+    @Json(name = "remove")
+    val removeVideoSet: Set<NamedWrapperForRemove>? = null,
```
howardrigberg: The original design of hiding the `NamedWrapper` classes was intentional. It was an implementation detail that is necessary to get the data to serialize to json properly, but it should not be exposed to consumers of the networking library. I wanted to avoid an unnecessary layer of complication in the networking api by exposing this class.

models/src/main/java/com/vimeo/networking2/params/NamedWrapperForAdd.kt
```diff
@@ -0,0 +1,16 @@
+package com.vimeo.networking2.params
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Container for [AddVideoToAlbum].
```
howardrigberg: Is there a reason why this class can't be private and internalized within `ModifyVideosInAlbumSpecs`? It is an internal detail to accommodate the api and is not needed outside of the class itself.

models/src/main/java/com/vimeo/networking2/params/NamedWrapperForRemove.kt
```diff
@@ -0,0 +1,16 @@
+package com.vimeo.networking2.params
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Container for [RemoveVideoFromAlbum].
+ */
+@JsonClass(generateAdapter = true)
+data class NamedWrapperForRemove(
```
howardrigberg: Is there a reason why this class can't be private and internalized within `ModifyVideosInAlbumSpecs`? It is an internal detail to accommodate the api and is not needed outside of the class itself.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -762,8 +760,8 @@ public void success(VimeoAccount vimeoAccount) {
             if (vimeoAccount.getUser() != null && (mEmail == null || mEmail.isEmpty())) {
                 // We must always have a `name` field, which is used by the Android Account Manager for
                 // display in the device Settings -> Accounts [KZ] 12/17/15
-                final String userName = vimeoAccount.getUser().mName;
-                final String name = userName != null ? userName : vimeoAccount.getUser().mUri;
+                final String userName = vimeoAccount.getUser().getName();
+                final String name = userName != null ? userName : vimeoAccount.getUser().getName();
```
howardrigberg: This should be `getUri()` not `getName()` at the end of this statement, right?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1483,13 +1482,13 @@ public Call modifyVideoInAlbums(@NotNull final Video video,
      * @param callback The VimeoCallback containing PictureResource data
```
howardrigberg: change to `PictureCollection`

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -1483,13 +1482,13 @@ public Call modifyVideoInAlbums(@NotNull final Video video,
...
      */
     @Nullable
-    public Call<PictureResource> createPictureResource(String uri, VimeoCallback<PictureResource> callback) {
+    public Call<PictureCollection> createPictureResource(String uri, VimeoCallback<PictureCollection> callback) {
```
howardrigberg: Just want to double check that this was intentional and without side-effects. `PictureCollection` and `PictureResource` are technically different objects.  `PictureResource` represents a subset of `PictureCollection` and nullability of members probably means this will function properly, but it is a little confusing. Maybe `PictureCollection` from `PictureResource` in the new models and use  `PictureResource` here to keep things the same?  I also noticed that you did not change the object type from `PictureResource` to `PictureCollection` in `activatePictureResource` below.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -425,11 +415,10 @@
                                @Header("Cache-Control") String cacheHeaderValue);
 
     @GET
-    @Serializer(converter = ConverterType.MOSHI)
-    Call<com.vimeo.networking2.FeedList> getFeedListMoshi(@Header("Authorization") String authHeader,
-                                                          @Url String uri,
-                                                          @QueryMap Map<String, String> options,
-                                                          @Header("Cache-Control") String cacheHeaderValue);
+    Call<FeedList> getFeedListMoshi(@Header("Authorization") String authHeader,
```
howardrigberg: Change the name to `getFeedList`?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -462,10 +451,10 @@
...
-                               @Url String uri,
-                               @QueryMap Map<String, String> options,
-                               @Header("Cache-Control") String cacheHeaderValue);
+    Call<TvodItemList> getTvodList(@Header("Authorization") String authHeader,
+                                   @Url String uri,
+                                   @QueryMap Map<String, String> options,
+                                   @Header("Cache-Control") String cacheHeaderValue);
 
     @GET
     Call<UserList> getUserList(@Header("Authorization") String authHeader,
```
howardrigberg, line 454: Remove the `getUserListMoshi` and `getVideoListMoshi` methods?

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -551,13 +540,13 @@
     Call<Object> POST(@Header("Authorization") String authHeader,
                       @Url String uri,
                       @Header("Cache-Control") String cacheHeaderValue,
-                      @Body HashMap<String, String> parameters);
+                      @Body Map<String, String> parameters);
```
howardrigberg: Is this version of `POST` being used? Can we delete it? Ditto for the `GET`.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoService.java
```diff
@@ -294,15 +291,15 @@
     // <editor-fold desc="POSTs">
     @POST
     Call<Comment> comment(@Header("Authorization") String authHeader, @Url String uri,
-                          @QueryMap Map<String, String> options, @Body HashMap<String, String> parameters);
+                          @QueryMap Map<String, String> options, @Body Map<String, String> parameters);
 
 
     @POST
-    Call<PictureResource> createPictureResource(@Header("Authorization") String authHeader, @Url String uri);
+    Call<PictureCollection> createPictureResource(@Header("Authorization") String authHeader, @Url String uri);
```
howardrigberg: See comments elsewhere about `PictureCollection` vs. `PictureResource`.

howardrigberg (changes requested):

models/src/main/java/com/vimeo/networking2/DashVideoFile.kt
anthonycr: Yes, also mentioned in a reply to nick, there will be a separate ticket for the full removal of the V1 models.

models/src/main/java/com/vimeo/networking2/Drm.kt
anthonycr: As far as I know, there are no issues with serializing from the old models to the new, since the JSON representations are unchanged.

models/src/main/java/com/vimeo/networking2/DashVideoFile.kt
anthonycr: One of the design principles of V2 is that we are not using inheritance if possible. The design of these classes was completely changed, as it was not correctly designed. The `quality` property existed on the `HlsVideoFile` and the `DashVideoFile` even though it was unnecessary (the quality was always `HLS` or `DASH`). If you look at the way the files were used, the `VideoFile` class was used as an interface to accept HLS, DASH, or progressive streams. Adding `DASH` and `HLS` as artificial qualities was just a hacky way to differentiate between stream types. The use of inheritance also obfuscated common properties when viewing the class. Since the type system already has a built-in way of determining whether it's an HLS or DASH stream, by simply using `instanceof` in java and `is` in kotlin, I simplified the designs of these classes and cleaned them up. They still share the common `VideoFile` interface, where the properties are defined. 

The only other common property shared between `DashVideoFile` and `HlsVideoFile` is the `live` property, which was originally shared in the `VideoFile` abstract class. However, this property does not exist on the `ProgressiveVideoFile` DTO, since live streams are always DASH or HLS, so I opted to duplicate it across the two files, as I don't think it warrants creating a separate interface.

models/src/main/java/com/vimeo/networking2/params/AddVideoToAlbum.kt
anthonycr: No, the `VimeoService` now serializes parameters using Moshi (See the modifications to `AnnotatedConverterFactory`)

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: I realize that one is a subset of the other, but in the API code, they share the same definition, so I intentionally removed `PictureResource` and replaced it with `PictureCollection`. I will change the function names and docs to reference the picture collection.

models/src/main/java/com/vimeo/networking2/params/ModifyVideosInAlbumSpecs.kt
anthonycr: I could make them inner classes, but we can't make them private, because Moshi requires the properties be visible to the adapters (they aren't visible if they are private), and we can't make the properties public but the class type private as that results in a kotlin error. Let me know if you have other ideas about how you want to design this class in Kotlin + Moshi. Personally, it doesn't bother me that these classes are public, but I understand the original design you implemented was user friendly. Perhaps a custom constructor would be suitable to improve the UX. Let me know what you think.

models/src/main/java/com/vimeo/networking2/params/NamedWrapperForAdd.kt
anthonycr: Cannot be made `private` as the property using it must be publicly exposed for use with Moshi.

models/src/main/java/com/vimeo/networking2/params/NamedWrapperForRemove.kt
anthonycr: Cannot be made `private` as the property using it must be publicly exposed for use with Moshi.

anthonycr:
@howardrigberg I addressed the issues you pointed out. The only thing outstanding is your comments on the album parameters classes, but I need input from you before I decide to make any changes.

anthonycr:
@howardrigberg changes made, back to you for review.

howardrigberg (approved):

PR #424, opened 2020-04-23
Removing V1 models

anthonycr:
# Summary
Now that the library has been migrated to the V2 models, the V1 models can be removed. This removes the V1 models served by the library as well as tests associated with them.


WhosNickDoglio (approved):

howardrigberg (approved):
👻 R.I.P. old models! 

Just want to confirm that you tested that the app can build against this version of networking?

PR #426, opened 2020-04-29
MPS-538: Remove wildcard imports from models module

WhosNickDoglio:
#### Summary
The library (KotlinPoet) that is being used to generate new models doesn’t work with wildcard imports, so removing them is necessary to generate new models.  

We currently only use them for `java.util.Date` so just updated those imports.


anthonycr (approved):
Looks good

PR #428, opened 2020-05-07
Vimeo error migration

anthonycr:
# Summary

`VimeoError` has been removed in favor of `VimeoResponse.Error`. The new error type has 3 subtypes that allows for further analysis. The root `Error` type has two properties, an `httpStatusCode`, and a `message`. These two fundamental properties are enough to describe an error in the library. The 3 subtypes of errors allow for further elaboration and are described below.

### Api

This error type provides an instance of an `ApiError` and is parsed from the error response of a network request. Occasionally, the library creates internal instances of this error itself in order to eagerly catch issues, such as when a parameter wrongly provided. In cases where the response is from the API, the `httpStatusCode` will provide the appropriate status code, otherwise local errors will populate it with -1.

### Exception

This error type represents exceptions that are thrown by the library. Examples of this are the network request failing because of a timeout or bad connection, in either case an `IOException` will probably be thrown. In the case of an exception being thrown, we most likely haven't received a response from the API, so the `httpStatusCode` will always be set to -1. 

### Unknown

This error type is reserved for cases that don't fall into either the `Api` case or the `Exception` case. The instance in which this is used is if the response should result in an `Api` error and no exception was thrown, but we were unable to parse the error response correctly. This type should never be created by the consumer as it requires the response for debugging.

## Other changes
- `VimeoError` has been completely removed
- Gson usage has been removed
- Stag usage has been removed


api-core/src/main/java/com/vimeo/networking2/VimeoResponse.kt
```diff
@@ -32,37 +32,40 @@ sealed class VimeoResponse<in T>(open val httpStatusCode: Int) {
          * Vimeo API returned an error response for the request you made.
          *
          * @param reason Info on the error.
-         * @param httpStatusCode HTTP status code.
+         * @param httpStatusCode HTTP status code, -1 if not applicable or if the error was created locally.
```
howardrigberg: Maybe create a named constant for -1 == LOCAL_ERROR?

api-core/src/main/java/com/vimeo/networking2/VimeoResponse.kt
```diff
@@ -32,37 +32,40 @@ sealed class VimeoResponse<in T>(open val httpStatusCode: Int) {
...
          */
         data class Api(
             val reason: ApiError,
             override val httpStatusCode: Int
-        ) : Error("API error: ${reason.errorCode ?: "unknown"}", httpStatusCode)
+        ) : Error("API error: ${reason.errorCode ?: NA}", httpStatusCode)
 
         /**
-         * Exception was thrown when making the request. This maybe due to no internet.
+         * Exception was thrown when making the request, e.g. the internet connection failed. This should only be used
```
howardrigberg: The kdocs for `Api`, `Exception` and `Unknown` could be copied from your PR description. They are a bit clearer.

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallAdapter.kt
```diff
@@ -42,7 +41,7 @@ internal class VimeoCallAdapter<T : Any>(
                     } else {
                         callbackExecutor.sendResponse {
                             callback.onError(
-                                VimeoResponse.Error.Generic(
+                                VimeoResponse.Error.Unknown(
```
howardrigberg, line 44: Run the formatter on this file.

example-kotlin-android/src/main/kotlin/com/vimeo/android/networking/example/kotlin/AccountPreferenceManager.kt
```diff
@@ -51,7 +53,8 @@ object AccountPreferenceManager {
```
howardrigberg, line 53: Redundant qualifier name on line 52.

example-kotlin-android/src/main/kotlin/com/vimeo/android/networking/example/kotlin/AccountPreferenceManager.kt
```diff
@@ -37,11 +37,13 @@ object AccountPreferenceManager {
...
+            return VimeoNetworkUtil.getMoshi()
+                    .adapter(VimeoAccount::class.java).fromJson(accountJSON)
         }
         set(vimeoAccount) {
             if (vimeoAccount != null) {
-                VimeoNetworkUtil.getGson().toJson(vimeoAccount)?.also {
+                VimeoNetworkUtil.getMoshi()
+                        .adapter(VimeoAccount::class.java).toJson(vimeoAccount)?.also {
                     sharedPreferences.edit().putString(CLIENT_ACCOUNT_JSON, it).apply()
                 } ?: removeClientAccount()
```
howardrigberg, line 48: Run the formatter on this file.

models/src/main/java/com/vimeo/networking2/enums/ErrorCodeType.kt
```diff
@@ -197,7 +199,7 @@ enum class ErrorCodeType(override val value: String?) : StringValue {
 
     INVALID_INPUT_BAD_CLIP_PRIVACY_VIEW("2232"),
 
-    INVALID_INPUT_CLIP_PRIVACY_PASSWORD_MISSING_PASSWORD2233("2233"),
+    INVALID_INPUT_CLIP_PRIVACY_PASSWORD_MISSING_PASSWORD("2233"),
```
howardrigberg: I already changed this in another recent PR. It also looks like a lot of the new error types I added are missing here.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
```diff
@@ -209,30 +172,62 @@ public void run() {
...
+     * @return a {@link VimeoResponse.Error} object extracted from the response or null
      */
-    @Nullable
-    public static <ResponseType_T> VimeoError getErrorFromResponse(@Nullable final Response<ResponseType_T> response) {
-        if (response != null && response.isSuccessful()) {
-            return null;
+    @NotNull
+    public static <ResponseType_T> VimeoResponse.Error getErrorFromResponse(@NotNull final Response<ResponseType_T> response) {
+        if (response.isSuccessful()) {
+            throw new IllegalStateException("Cannot get error from a successful response");
```
WhosNickDoglio, line 209: Like this a lot more than returning null.

vimeo-networking/src/main/java/com/vimeo/networking/logging/LoggingInterceptor.java
```diff
@@ -47,6 +49,9 @@
  */
 public class LoggingInterceptor implements Interceptor {
 
+    @NotNull
+    private final Gson gson = new GsonBuilder().setPrettyPrinting().create();
```
WhosNickDoglio, line 53: Do we need `Gson` for this? If we use `Moshi` for this we could completely remove it as a dependency right?

WhosNickDoglio (approved):
Overall this looks great, excited to get this merged. 👍

vimeo-networking/src/main/java/com/vimeo/networking/logging/LoggingInterceptor.java
anthonycr: I can't find a way to do this with `Moshi`. There doesn't seem to be any printing options available, so I stuck with gson here. I'm thinking in the future we just remove this functionality from the library, since we mostly debug networking responses using other tools like Stetho.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
anthonycr: Yeah, I looked at all the usage of it and in every case we were checking whether or not the response was unsuccessful before calling the function, so failing like this makes more sense.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -871,8 +907,9 @@ public void run() {
                     VimeoClient.sContinuePinCodeAuthorizationRefreshCycle = false;
                     mTimer.cancel();
                     if (authCallback != null && now >= mExpiresInNano) {
-                        final VimeoError error = new VimeoError("Pin code expired.");
-                        error.setLocalErrorCode(LocalErrorCode.UNABLE_TO_LOGIN_PINCODE_EXPIRED);
+                        final VimeoResponse.Error error = new VimeoResponse.Error.Exception(
```
howardrigberg, line 910: As-is, this will break the TV implementation of pin code auth. If we are no longer supporting TV, we could simply remove the pin code support completely from VimeoClient since this was created solely for TV and is not used by any other consumers AFAIK.

vimeo-networking/src/main/java/com/vimeo/networking/utils/PinCodeExpiredException.kt
```diff
@@ -0,0 +1,8 @@
+package com.vimeo.networking.utils
+
+import java.lang.IllegalStateException
+
+/**
+ * An exception indicating that a pin code has expired.
+ */
+class PinCodeExpiredException(override val message: String) : IllegalStateException(message)
```
howardrigberg, line 8: See comments above regarding PinCode.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
```diff
@@ -209,30 +172,62 @@ public void run() {
...
+                }
             } catch (final Exception e) {
                 ClientLogger.e("Error while attempting to convert response body to VimeoError", e);
+                if (errorBody == null) {
+                    errorBody = "Unable to read error body";
+                }
+                return new VimeoResponse.Error.Unknown(errorBody, response.code());
             }
+        } else {
+            return new VimeoResponse.Error.Unknown("Null error body", response.code());
```
howardrigberg, line 230: Maybe "Null error body" and "Unable to read error body" could be pulled out into error constants so callers could potentially distinguish these errors? Or maybe we should pull these out as additional typed error classes? It seems weird to categorize them as "Unknown" types when they are very clearly known in advance.

howardrigberg (changes requested):
Nice job, Anthony. Just a few minor comments.

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallAdapter.kt
anthonycr: Recent Android Studio change messed with the default formatting, this formatting is actually the closest to the previous version of AS, and is properly formatted for me. I will commit the changes to the code style.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: It certainly does break the current pincode code, but I provided a very simple alternative approach. Instead of the old `vimeoError.localErrorCode == LocalErrorCode.UNABLE_TO_LOGIN_PINCODE_EXPIRED` check, the new code would just check `vimeoError is VimeoResponse.Error.Exception && vimeoEror.throwable is PinCodeExpiredException`. It's a direct conversion that's easy to do. I'm going to leave it in the code base for now.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoNetworkUtil.java
anthonycr: These errors are not expected to happen on a frequency which would justify us enumerating the cases in which they are created. Even when the Vimeo API returns a 500 it is supposed to return us a valid `ApiError`. The cases in which these `Unknown` errors will be received are exceptional and unless we see a large number of these errors, I don't want the class to be overengineered to handle scenarios that are unlikely to happen. 

These errors are "Unknown" in the sense that it is unknown _why_ they are happening. In the case of an `Api` or `Exception`, the reason for their occurrence is known, but in the case of an `Unknown`, it is unclear, for instance, why we would receive an unsuccessful response with a null error body, or why the error body would be non-null but unreadable. The purpose of `Unknown` is to act as a net for errors that don't fit into the other two categories. The reason we are not using `Exception` is that `Exception` is reserved for cases without an `httpStatusCode` (i.e. the server never responded).

anthonycr:
@howardrigberg back to you for review.

howardrigberg (approved):

PR #431, opened 2020-06-02
Adding InvalidToken error

anthonycr:
# Summary
In the previous definition of `VimeoError`, an "invalid token" error was defined as an error where a specific header `WWW-Authenticate` was set with value `Bearer error="invalid_token"`. Using this definition was strange considering the available information in the `ApiError`, so when the `VimeoResponse.Error` type was created, the invalid token error was defined as a `VimeoResponse.Error.Api` where the `errorCode` was either 8000 or 8003. However, testing showed that requests to the `/feed` URI with a logged out token would return an 8000 error code under certain circumstances. As a result, this check is not reliable. Instead, I opted to go back to the original error definition and check the headers manually when the response is constructed.

I created a new error type, `VimeoResponse.Error.InvalidToken`, explicitly for an invalid token. Any token used to make a request that results in this response should be discarded.

## Testing
Make a request with an invalid token, verify that an `InvalidToken` response is returned.

### DIFF
--- a/api-core/src/main/java/com/vimeo/networking2/VimeoResponse.kt
+++ b/api-core/src/main/java/com/vimeo/networking2/VimeoResponse.kt
@@ -47,6 +47,20 @@ sealed class VimeoResponse<in T>(open val httpStatusCode: Int) {
         ) : Error("API error: ${reason.errorCode ?: NA}", httpStatusCode)
 
         /**
+         * The Vimeo API returned a fatal error that indicates that the token used to make the request is invalid. The
+         * token should be discarded and a new token should be acquired. Further requests should not be made with the
+         * token.
+         *
+         * @param reason Info on the error.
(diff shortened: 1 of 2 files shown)


WhosNickDoglio (approved):
LGTM 👍

rpassis:
Hey @anthonycr FYI we just ran into a very similar issue.

I don't know if this is the case on Android but one small gotcha is that the `WWW-Authenticate` header key sent by our API was getting changed to `Www-Authenticate` under the hood on iOS, and my check was failing because dictionary keys are case sensitive in Swift.

This is from the iOS docs

> HTTP headers are case insensitive. To simplify your code, URL Loading System canonicalizes certain header field names into their standard form. For example, if the server sends a content-length header, it’s automatically adjusted to be Content-Length.

PR #434, opened 2020-06-23
Add the video status endpoint

abalcer-ls:
# Summary
Add support the /videos/{id}/status endpoint to the VimeoService

Was added:

- the data class `VideoStatus`
- the enum `VideoStateType`
- the new caller `GetRequestCaller.VIDEO_STATUS`

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/VideoStatus.kt
+++ b/models/src/main/java/com/vimeo/networking2/VideoStatus.kt
@@ -1,0 +1,43 @@
+package com.vimeo.networking2
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import com.vimeo.networking2.enums.VideoStateType
+import com.vimeo.networking2.enums.asEnum
+import java.io.Serializable
+
+/**
+ * The current status of the video transcoding process.
+ */
+@JsonClass(generateAdapter = true)
+data class VideoStatus(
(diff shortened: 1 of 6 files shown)


models/src/main/java/com/vimeo/networking2/VideoStatus.kt
```diff
@@ -0,0 +1,43 @@
...
+         * The current state of the transcoding process.
+         */
+        @Json(name = "state")
+        val state: String? = null,
+
+        /**
+         * The percentage of the transcoding process that is complete.
+         */
+        @Json(name = "progress")
+        val progress: Int? = 0,
```
anthonycr: Default should be `null`.

models/src/main/java/com/vimeo/networking2/VideoStatus.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        /**
+         * The remaining time in seconds before transcoding is complete.
+         */
+        @Json(name = "time_left")
+        val timeLeft: Long? = 0
```
anthonycr: Default should be `null`.

anthonycr (changes requested):
Nice job, just two minor changes need to be made regarding the nullability of two properties.

anthonycr (approved):
Looks good.

PR #435, opened 2020-06-25
VimeoService refactor

anthonycr:
# Summary
The first step toward refactoring the `VimeoClient` is to rewrite the `VimeoService`. In this PR I created a Kotlin version of the `VimeoService` and removed the authentication related functions from it and put them into the `AuthService`. As part of this rewrite of the service in Kotlin, I made many parameters take actual classes or enums instead of raw types like `String`. Specifically I created a `Scopes` class and converters for the `Scopes` and `GrantType` enum. This prevents consumers from incorrectly using the APIs. 

Additionally, `VimeoAccount` was changed so that the `user` and `accessToken` properties are non null. The reason for this is that `VimeoAccount` can now conform to the `AccessTokenProvider` and can work in conjunction with the `UnauthenticatedAccessToken` class. It now replaces the `AuthenticatedAccessToken` class. I refactored the `AuthenticatorImpl` to support these new changes such as making the `AuthService` more explicit in the required parameters.

The `VimeoService` now lives in a `request` module that will be the home of the new `VimeoClient` when it is written.

Note: I am pointing this PR into a non-develop branch as the changes to `VimeoAccount` are breaking (see the temporary changes to the `VimeoAccountFactory`).


settings.gradle
```diff
@@ -1,3 +1,4 @@
+include ':request'
```
WhosNickDoglio: Optional: This should probably be included in the next line and not separate right?

build.gradle
```diff
@@ -17,7 +17,7 @@ plugins {
 
 ext {
     supportLibraryVersion = '28.0.0'
-    kotlin_version = '1.3.61'
+    kotlin_version = '1.3.72'
     retrofitVersion = '2.7.1'
```
WhosNickDoglio: Optional: Retrofit, Moshi and Okio also have newer versions if you want to update as part of this PR.

auth/src/main/java/com/vimeo/networking2/internal/VimeoParametersConverterFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.vimeo.networking2.GrantType
+import com.vimeo.networking2.Scopes
+import retrofit2.Converter
+import retrofit2.Retrofit
+import java.lang.reflect.Type
+
+/**
+ * A [Converter.Factory] used for auth parameter conversion.
+ */
+class VimeoParametersConverterFactory : Converter.Factory() {
```
WhosNickDoglio, line 33: This is cool.

api-core/src/main/java/com/vimeo/networking2/GrantType.kt
```diff
@@ -1,9 +1,11 @@
 package com.vimeo.networking2
 
+import com.vimeo.networking2.enums.StringValue
+
 /**
  * Grant types.
```
WhosNickDoglio: Maybe something more descriptive could be here?

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -129,19 +129,16 @@ interface Authenticator {
         fun create(serverConfig: ServerConfig): Authenticator {
```
WhosNickDoglio: Optional: Could we override `invoke` instead of adding a `create` method here? So Kotlin callers could just call `Authenticator(serverConfig)`?


We could also do something like the following: 

```
    companion object {

        @JvmName("create")
        @JvmStatic
        operator fun invoke(serverConfig: ServerConfig): Authenticator {
            return AuthenticatorImpl(authService, authHeaders, Scopes(serverConfig.scopes))
        }
    }
```

So Java callers would have `Authenticator.create(serverConfig)` and Kotlin callers would have `Authenticator(serverConfig)`

**This suggestion also requires targeting Java 8** so we could circle back to this when we update to that.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -129,19 +129,16 @@ interface Authenticator {
...
+                        serverConfig.clientId,
+                        serverConfig.clientSecret
+                    )
 
-            val scopes = serverConfig.scopes
-                .joinToString(separator = " ", transform = { it.name.toLowerCase() })
-
-            return AuthenticatorImpl(authService, authHeaders, scopes)
+            return AuthenticatorImpl(authService, authHeaders, Scopes(serverConfig.scopes))
         }
```
WhosNickDoglio: There's a few extra blank lines at the bottom of this file that can be removed.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoAccountFactory.kt
```diff
@@ -2,11 +2,14 @@
...
 import com.vimeo.networking2.VimeoAccount
 
 /**
  * Create a [VimeoAccount] with only the provided [accessToken] set.
  */
+@Suppress("UnsafeCallOnNullableType")
 fun createVimeoAccountWithToken(accessToken: String?): VimeoAccount = VimeoAccount(
-    accessToken = accessToken
+    accessToken = accessToken!!,
+    user = User()
```
WhosNickDoglio, line 15: Can we add a TODO here so we remember to remove this later?

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -0,0 +1,478 @@
+/*
```
WhosNickDoglio, line 1: This looks like the only file with copyright, shouldn't it be applied to all files?

WhosNickDoglio, line 1: This looks like the only file with copyright, shouldn't it be applied to all files?

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -0,0 +1,478 @@
...
+
+    @GET("me/connected_apps/{type}")
+    fun getConnectedApp(
+        @Header(AUTHORIZATION) authorization: String,
+        @Path("type") type: String,
+        @Header(CACHE_CONTROL) cacheControl: CacheControl?
+    ): VimeoCall<ConnectedApp>
+
+    @PUT("me/connected_apps/{type}")
+    @Headers("Cache-Control: no-cache, no-store")
```
WhosNickDoglio: `"Cache-Control: no-cache, no-store"` gets repeated a couple of times and could be a `const` instead

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -0,0 +1,478 @@
...
+        @Path("type") type: String,
+        @Body parameters: Map<String, Any>
+    ): VimeoCall<ConnectedApp>
+
+    @DELETE("me/connected_apps/{type}")
+    @Headers("Cache-Control: no-cache, no-store")
+    fun deleteConnectedApp(
+        @Header(AUTHORIZATION) authorization: String,
+        @Path("type") type: String
+    ): VimeoCall<Void>
```
WhosNickDoglio: Shouldn't this be `Unit` instead of `Void`?

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -0,0 +1,478 @@
...
+import retrofit2.http.POST
+import retrofit2.http.PUT
+import retrofit2.http.Path
+import retrofit2.http.QueryMap
+import retrofit2.http.Url
+
+/**
+ * Retrofit service for the Vimeo API.
+ */
+@Suppress("unused", "UndocumentedPublicFunction", "ComplexInterface")
```
WhosNickDoglio, line 46: Optional: Maybe a TODO to remember to come back for documentation?

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -1,62 +1,180 @@
...
+import retrofit2.http.FormUrlEncoded
 import retrofit2.http.Header
+import retrofit2.http.Headers
 import retrofit2.http.POST
 
 /**
  * All the authentication endpoints.
  */
+@Suppress("unused", "LongParameterList", "ComplexInterface")
 internal interface AuthService {
```
WhosNickDoglio, line 43: Optional: Maybe a TODO to remember to come back for documentation?

WhosNickDoglio (changes requested):
Overall this looks really good.  I'm excited for when we can starting using this in the app.

Just a handful of small nits and suggestions. 👍

anthonycr:
@WhosNickDoglio thanks for the review, i made all the changes.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -126,24 +147,19 @@ interface Authenticator {
          *                     interceptors, read timeouts, base url etc...) that can be set for
          *                     authentication and making requests.
          */
-        fun create(serverConfig: ServerConfig): Authenticator {
-
+        @JvmName("create")
```
WhosNickDoglio, line 151: This should also have `@JvmStatic` so it can be `Authenticator.create(serverConfig)` instead of `Authenticator.Companion.create(serverConfig)` for Java users.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -1,3 +1,24 @@
...
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
```
WhosNickDoglio, line 21: The documentation for this class should be updated to reflect the new usage. 
`val authenticator = Authenticator(serverConfig)`

WhosNickDoglio (changes requested):
Just a few tiny more changes before this is ready to be merged 👍

anthonycr:
@WhosNickDoglio thanks, missed those, should be good now.

WhosNickDoglio (approved):
LGTM! 👏

PR #436, opened 2020-07-10
Vimeo Configuration refactor

anthonycr:
# Summary
In this PR, I rewrote the `Configuration` class and its builder, and converted a few classes to Kotlin in order to support this refactor. All new classes live in the `request` module at the moment.

In this PR, I created
- `Configuration`
- `Configuration.Builder`
- `AccountStore` - converted from Kotlin
- `InMemoryAccountStore` - default implementation of `AccountStore`
- `AcceptHeaderInterceptor` - converted from Kotlin
- `CacheControlHeaderInterceptor` - converted from Kotlin
- `LanguageHeaderInterceptor` - converted from Kotlin
- `UserAgentHeaderInterceptor` - converted from Kotlin
- `LogDelegate` - replaces `LogProvider`
- `LogDelegate.Level` - replaces `LogLevel`
- `VimeoLogger` - replaces `ClientLogger`
- `DefaultLogDelegate` - replaces default behavior of `ClientLogger` when no `LogProvider` is specified

There are a couple differences with the new `Configuration`. 
- We now supply a default `AccountStore`.
- We now explicitly supply a default `LogDelegate` instead of implicitly logging if no provider was specified.
- We no longer allow a `Configuration` to be constructed using an access token. ID, secret, and scope are now always required.

### DIFF
--- a/request/src/main/java/com/vimeo/networking2/Configuration.kt
+++ b/request/src/main/java/com/vimeo/networking2/Configuration.kt
@@ -246,27 +246,27 @@ data class Configuration(
         /**
          * The default max age of the cache, 2 hours.
          */
-        private const val DEFAULT_CACHE_MAX_AGE = 60 * 60 * 2
+        const val DEFAULT_CACHE_MAX_AGE = 60 * 60 * 2
 
         /**
-         * The default size of the cache, 10 Mib
+         * The default size of the cache, 10 Mib.
          */
-        private const val DEFAULT_CACHE_SIZE = 10L * 1024 * 1024
+        const val DEFAULT_CACHE_SIZE = 10L * 1024 * 1024
 
         /**
          * Default response timeout, 60 seconds.
          */
-        private const val DEFAULT_TIMEOUT = 60
+        const val DEFAULT_TIMEOUT = 60
 
         /**
          * An unspecified user agent.
          */
-        private const val DEFAULT_USER_AGENT = "UNSPECIFIED"
+        const val DEFAULT_USER_AGENT = "UNSPECIFIED"
 
         /**
          * The default base URL for the Vimeo API.
          */
-        private const val DEFAULT_BASE_URL = "https://api.vimeo.com/"
+        const val DEFAULT_BASE_URL = "https://api.vimeo.com/"
     }
 
 }
--- a/request/src/main/java/com/vimeo/networking2/Configuration.kt
+++ b/request/src/main/java/com/vimeo/networking2/Configuration.kt
@@ -115,14 +115,14 @@ data class Configuration(
         private var cacheMaxAgeSeconds: Int = DEFAULT_CACHE_MAX_AGE
 
         /**
-         * Use a different base URL. Defaults to [DEFAULT_BASE_URL].
+         * Specify a base URL. Defaults to [DEFAULT_BASE_URL].
          *
          * @see Configuration.baseUrl
          */
         fun withBaseUrl(baseUrl: String) = apply { this.baseUrl = baseUrl }
 
         /**
-         * Use a different code grant redirect. Defaults to `vimeo{clientId}://auth`.
+         * Specify a code grant redirect. Defaults to `vimeo{clientId}://auth`.
          *
          * @see Configuration.codeGrantRedirectUri
          */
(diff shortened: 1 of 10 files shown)


request/src/main/java/com/vimeo/networking2/logging/LogDelegate.kt
```diff
@@ -0,0 +1,72 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.vimeo.networking2.logging
+
+/**
+ * A logger interface for the networking library.
+ */
+@Suppress("FunctionMinLength")
+interface LogDelegate {
```
WhosNickDoglio, line 30: Like this a lot 👏

request/src/main/java/com/vimeo/networking2/Configuration.kt
```diff
@@ -0,0 +1,272 @@
...
+            isCertPinningEnabled = isCertPinningEnabled,
+            logDelegate = logDelegate,
+            logLevel = logLevel,
+            cacheDirectory = cacheDirectory,
+            cacheMaxSizeBytes = cacheMaxSizeBytes,
+            cacheMaxAgeSeconds = cacheMaxAgeSeconds
+        )
+    }
+
+    companion object {
```
WhosNickDoglio, line 245: Optional: With the  `Configuration` constructor being public I think these could also be public so if the consumer uses the constructor they still have access to the defaults without having to define them themselves?

request/src/main/java/com/vimeo/networking2/interceptors/CacheControlHeaderInterceptor.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import com.vimeo.networking2.Configuration
+import okhttp3.Interceptor
+import okhttp3.Response
+
+/**
+ * Rewrite the server's cache-control header because our server sets all `Cache-Control` headers to `no-store`. To get
+ * data from the cache, we set a max age to [Configuration.cacheMaxAgeSeconds]. This prevents OkHttp from throwing a 504
+ * exception. API no longer sends a max age in the header. OkHttp determines the cache to be invalid. This is a work
+ * around for this issue.
```
WhosNickDoglio: > This prevents OkHttp from throwing a 504 exception. API no longer sends a max age in the header. OkHttp determines the cache to be invalid. This is a work around for this issue.

Optional: the phrasing here feels awkward to me, maybe can rephrase this?

request/src/main/java/com/vimeo/networking2/Configuration.kt
```diff
@@ -0,0 +1,272 @@
...
+        private var cacheDirectory: File? = null
+        private var cacheMaxSizeBytes: Long = DEFAULT_CACHE_SIZE
+        private var cacheMaxAgeSeconds: Int = DEFAULT_CACHE_MAX_AGE
+
+        /**
+         * Specify a base URL. Defaults to [DEFAULT_BASE_URL].
+         *
+         * @see Configuration.baseUrl
+         */
+        fun withBaseUrl(baseUrl: String) = apply { this.baseUrl = baseUrl }
```
WhosNickDoglio, line 122: I _think_ we already talked about this but I like the choice to go with a `with` prefix over a `set` prefix for the `Builder` methods.  👏

request/src/main/java/com/vimeo/networking2/Configuration.kt
```diff
@@ -0,0 +1,272 @@
...
+        private var networkInterceptors: List<Interceptor> = emptyList()
+        private var interceptors: List<Interceptor> = emptyList()
+
+        private var userAgent: String = DEFAULT_USER_AGENT
+
+        private var requestTimeoutSeconds: Int = DEFAULT_TIMEOUT
+
+        private var isCertPinningEnabled: Boolean = true
+
+        private var logDelegate: LogDelegate? = DefaultLogDelegate()
```
WhosNickDoglio, line 110: Optional: Should we default to logging enabled? I think it would make sense to only enable logging if the consumer chooses to.

WhosNickDoglio (approved):
Overall I think this looks great. 👏

Left a handful of optional suggestions 👍

request/src/main/java/com/vimeo/networking2/interceptors/CacheControlHeaderInterceptor.kt
anthonycr: Yes, I had just copied the docs from the previous java class and didn't even realize how awkward the language was. Will fix, thanks for pointing it out.

request/src/main/java/com/vimeo/networking2/Configuration.kt
anthonycr: I think that's a good idea. They might not be necessary, but I don't want to assume too many use cases and perhaps users will need them.

request/src/main/java/com/vimeo/networking2/Configuration.kt
anthonycr: You're right. We already use other "production-ready" defaults, so it makes sense that logging behaved similarly. However, I think it's be better to set the `LogDelegate.Level` to `NONE` instead of defaulting to no log delegate. If the default `LogDelegate.Level` is `VERBOSE` but the default `LogDelegate` is null, it might be confusing to the consumer why the library isn't logging.

anthonycr:
@WhosNickDoglio thanks for the review, good callouts. I made a few changes and would like your re-approval before merging.

WhosNickDoglio (approved):
LGTM 👏

PR #438, opened 2020-07-21
MPS-537: Remove Serializable

WhosNickDoglio:
#### Issue
[MPS-537](https://vimean.atlassian.net/browse/MPS-537)

#### Summary
Removed all Serializable related code from base models as they soon will be generated. 🎉

#### How to Test
Confirm projects still builds.


anthonycr (approved):
Looks good, very exciting step forward

PR #439, opened 2020-07-22
Request Redesign - VimeoApiClient & Authenticator

anthonycr:
# Summary
This PR rewrites the `VimeoClient` in Kotlin in `VimeoApiClient` and implementation `VimeoApiClientImpl`. The main request design is as follows:
```kotlin
val vimeoRequest: VimeoRequest = VimeoApiClient.instance().fetchData(
    uri: String, 
    parameters: Map<String, String>,
    callback: vimeoCallback(
        onSuccess = { success: VimeoResponse.Success<Data> -> },
        onError = { error: VimeoResponse.Error -> }
    )
)
```
`VimeoRequest` can be used to cancel a request. Initialization is as follows:
```kotlin
VimeoApiClient.initialize(configuration, Authenticator.instance())
```
`Authenticator` is used for all authentication requests. It in turn must be configured:
```kotlin
Authenticator.initialize(configuration)
```
Authenticating with the `Authenticator` instance used to initialize the client will also authorize the client. If requests are made with the client before properly authorizing, then the requests may fail as unauthenticated.

While the above configuration should be used by most clients, it is not the only initialization route. For consumers that need multiple client instances, the client can be constructed manually. The initialization sequence for that is shown below:
```kotlin
val authenticator = Authenticator.create(configuration)
val vimeoApiClient = VimeoApiClient.create(configuration, authenticator)
```
In this case, the consumer will need to keep around the references to the `Authenticator` and `VimeoApiClient` since the internal singleton instance is not being initialized.

The implementation of the `VimeoApiClient` differs from the previous `VimeoClient` in that it is designed as a thin wrapper around the `VimeoService`. For the most part, local errors are not created, except where the client design introduces nullability issues. Generally, the extent of the client implementation complexity is packaging together parameters into a map to be used as a request body or query map.


WhosNickDoglio (approved):
LGTM 👍

PR #440, opened 2020-07-22
VimeoClient -> VimeoApiClient

anthonycr:
# Summary
This PR rewrites the `VimeoClient` in Kotlin in `VimeoApiClient` and implementation `VimeoApiClientImpl`. The main request design is as follows:
```kotlin
val vimeoRequest: VimeoRequest = VimeoApiClient.instance().fetchData(
    uri: String, 
    parameters: Map<String, String>,
    callback: vimeoCallback(
        onSuccess = { success: VimeoResponse.Success<Data> -> },
        onError = { error: VimeoResponse.Error -> }
    )
)
```
`VimeoRequest` can be used to cancel a request. Initialization is as follows:
```kotlin
VimeoApiClient.initialize(configuration, Authenticator.instance())
```
`Authenticator` is used for all authentication requests. It in turn must be configured:
```kotlin
Authenticator.initialize(configuration)
```
Authenticating with the `Authenticator` instance used to initialize the client will also authorize the client. If requests are made with the client before properly authorizing, then the requests may fail as unauthenticated.

While the above configuration should be used by most clients, it is not the only initialization route. For consumers that need multiple client instances, the client can be constructed manually. The initialization sequence for that is shown below:
```kotlin
val authenticator = Authenticator.create(configuration)
val vimeoApiClient = VimeoApiClient.create(configuration, authenticator)
```
In this case, the consumer will need to keep around the references to the `Authenticator` and `VimeoApiClient` since the internal singleton instance is not being initialized.

The implementation of the `VimeoApiClient` differs from the previous `VimeoClient` in that it is designed as a thin wrapper around the `VimeoService`. For the most part, local errors are not created, except where the client design introduces nullability issues. Generally, the extent of the client implementation complexity is packaging together parameters into a map to be used as a request body or query map.


api-core/src/main/java/com/vimeo/networking2/account/InMemoryAccountStore.kt
```diff
@@ -28,12 +30,6 @@ class InMemoryAccountStore : AccountStore {
...
     override fun loadAccount(): VimeoAccount? = vimeoAccount
-
-    override fun saveAccount(vimeoAccount: VimeoAccount, email: String) {
-        this.vimeoAccount = vimeoAccount
-    }
-
-    override fun deleteAccount(vimeoAccount: VimeoAccount) {
-        this.vimeoAccount = null
-    }
+    override fun storeAccount(vimeoAccount: VimeoAccount) { this.vimeoAccount = vimeoAccount }
```
WhosNickDoglio, line 35: Optional: These would look nicer spaced out.

vimeo-networking/src/main/java/com/vimeo/networking/utils/VimeoAccountFactory.kt
```diff
@@ -11,6 +10,5 @@ import com.vimeo.networking2.VimeoAccount
 @Suppress("UnsafeCallOnNullableType")
```
WhosNickDoglio: Pretty sure this Suppression can be removed

models/src/main/java/com/vimeo/networking2/params/SearchDateType.kt
```diff
@@ -0,0 +1,13 @@
+package com.vimeo.networking2.params
+
+import com.vimeo.networking2.enums.StringValue
+
+/**
+ * Created by anthony.restaino on 7/22/20.
+ */
```
WhosNickDoglio: Replace this with actual doc.

models/src/main/java/com/vimeo/networking2/params/SearchDurationType.kt
```diff
@@ -0,0 +1,12 @@
+package com.vimeo.networking2.params
+
+import com.vimeo.networking2.enums.StringValue
+
+/**
+ * Created by anthony.restaino on 7/22/20.
+ */
```
WhosNickDoglio: Replace this with actual doc.

models/src/main/java/com/vimeo/networking2/params/SearchFacetType.kt
```diff
@@ -0,0 +1,15 @@
+package com.vimeo.networking2.params
+
+import com.vimeo.networking2.enums.StringValue
+
+/**
+ * Created by anthony.restaino on 7/22/20.
+ */
```
WhosNickDoglio: Replace this with actual doc.

models/src/main/java/com/vimeo/networking2/params/SearchFilterType.kt
```diff
@@ -0,0 +1,14 @@
+package com.vimeo.networking2.params
+
+import com.vimeo.networking2.enums.StringValue
+
+/**
+ * Created by anthony.restaino on 7/22/20.
+ */
```
WhosNickDoglio: Replace this with actual doc.

models/src/main/java/com/vimeo/networking2/params/SearchSortDirectionType.kt
```diff
@@ -0,0 +1,11 @@
+package com.vimeo.networking2.params
+
+import com.vimeo.networking2.enums.StringValue
+
+/**
+ * Created by anthony.restaino on 7/22/20.
+ */
```
WhosNickDoglio: Replace this with actual doc.

models/src/main/java/com/vimeo/networking2/params/SearchSortType.kt
```diff
@@ -0,0 +1,15 @@
+package com.vimeo.networking2.params
+
+import com.vimeo.networking2.enums.StringValue
+
+/**
+ * Created by anthony.restaino on 7/22/20.
+ */
```
WhosNickDoglio: Replace this with actual doc.

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallAdapter.kt
```diff
@@ -30,87 +31,58 @@ import retrofit2.Response
...
+ * thread and return the result in the calling thread.
  *
- * @param call                  Retrofit call object.
- * @param callbackExecutor      Callback executor set by Retrofit to return the result. Retrofit
- *                              itself sets it to the main thread on Android. If the executor is
- *                              null, then the callback will be executed on the thread provided by
- *                              OkHttp's dispatcher.
+ * @param call Retrofit call object.
+ * @param callbackExecutor Callback executor set by Retrofit to return the result. Retrofit itself sets it to the main
+ * thread on Android. If the executor is null, then the callback will be executed on the thread provided b OkHttp's
```
WhosNickDoglio: "thread provided b**y** OkHttp's"

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallEmptyResponseAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+import retrofit2.Response
+import java.util.concurrent.Executor
+
+/**
+ * A custom Retrofit call adapter for requests with an expected empty response. This was created in order to parse an
+ * error response from the API in a background thread and return the result in the calling thread.
+ *
+ * @param call Retrofit call object with empty response.
+ * @param callbackExecutor Callback executor set by Retrofit to return the result. Retrofit itself sets it to the main
+ * thread on Android. If the executor is null, then the callback will be executed on the thread provided b OkHttp's
```
WhosNickDoglio: "thread provided b**y** OkHttp's"

request/src/main/java/com/vimeo/networking2/internal/LocalVimeoCallAdapter.kt
```diff
@@ -0,0 +1,48 @@
...
+     * Enqueue the [VimeoCallback] with the provided [ApiError].
+     */
+    fun <T> enqueueError(apiError: ApiError, callback: VimeoCallback<T>): VimeoRequest {
+        retrofit.callbackExecutor()
+            .sendResponse { callback.onError(VimeoResponse.Error.Api(apiError, VimeoResponse.HTTP_NONE)) }
+
+        return NoOpVimeoRequest
+    }
+
+    private fun Executor?.sendResponse(action: () -> Unit) {
```
WhosNickDoglio: This same extension function was removed in other CallAdapters, just want to confirm we still need it here?

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -41,34 +41,34 @@ import retrofit2.http.Url
...
         @Header(AUTHORIZATION) authorization: String,
+        @QueryMap options: Map<String, @JvmSuppressWildcards String>,
         @Header(CACHE_CONTROL) cacheControl: CacheControl?
     ): VimeoCall<ConnectedAppList>
 
     @GET("me/connected_apps/{type}")
     fun getConnectedApp(
         @Header(AUTHORIZATION) authorization: String,
         @Path("type") type: String,
+        @QueryMap options: Map<String, @JvmSuppressWildcards String>,
```
WhosNickDoglio: Why do we need `@JvmSuppressWildcards` for all the second types in the `Maps` parameters?

WhosNickDoglio (changes requested):
This is excellent and I'm looking forward to getting this merged! 

Left some comments and suggestions 👍

request/src/main/java/com/vimeo/networking2/VimeoService.kt
anthonycr: It's because of `in` and `out` variance in Kotlin. In Java, type parameters are often invariant, while in Kotlin they are encouraged to vary if possible for increased utility. For example, here, the `Map` interface in Kotlin has a covariant value type, while in Java's `Map` the value type is invariant. Metaprogramming utilities like Dagger and Retrofit have difficulty handling co- and contravariance in type parameters since it is much harder to validate types when the definition accepts a superset or subset of a type. As a result, these utilities usually throw errors when encountering wildcards, so here we need to suppress the wildcard introduced by Kotlin's `Map` type.

You can read more about the specific issue here: https://github.com/square/retrofit/issues/1805

api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
```diff
@@ -34,8 +34,74 @@ object ApiConstants {
...
+        const val ENDPOINT_RECOMMENDATIONS = "/recommendations"
+        const val ENDPOINT_TERMS_OF_SERVICE = "documents/termsofservice"
+        const val ENDPOINT_PRIVACY_POLICY = "documents/privacy"
+        const val ENDPOINT_PAYMENT_ADDENDUM = "documents/paymentaddendum"
+    }
+
+    /**
+     * Parameters
+     */
+    object Parameters {
```
howardrigberg, line 53: I'm confused by the intention of these parameters. Were they meant to replace the ones that are currently in Vimeo.java? If so, we should point all of the references here and delete the other file. If not, we should probably remove these redundant values.

api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
```diff
@@ -34,8 +34,74 @@ object ApiConstants {
...
+        const val PARAMETER_REDIRECT_URI = "redirect_uri"
+        const val PARAMETER_RESPONSE_TYPE = "response_type"
+        const val PARAMETER_STATE = "state"
+        const val PARAMETER_SCOPE = "scope"
+        const val PARAMETER_TOKEN = "token"
+        const val PARAMETER_ID_TOKEN = "id_token"
+        const val PARAMETER_CLIENT_ID = "client_id"
+        const val PARAMETER_USERS_NAME = "name"
+        const val PARAMETER_EMAIL = "email"
+        const val PARAMETER_PASSWORD = "password"
```
howardrigberg, line 63: This constant is reused locally, but other redundant strings like "name" aren't.

api-core/src/main/java/com/vimeo/networking2/VimeoCallback.kt
```diff
@@ -42,3 +44,17 @@ interface VimeoCallback<ResponseType_T> {
     fun onError(error: VimeoResponse.Error)
 
 }
+
+/**
+ * Create a lambda friendly instance of the [VimeoCallback].
```
howardrigberg, line 48: "A lambda friendly..."

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
```diff
@@ -483,26 +483,26 @@ public VimeoAccount authorizeWithClientCredentialsGrantSync() {
                         Vimeo.FIELD_NAME,
                         ErrorCodeType.INVALID_INPUT_NO_NAME,
                         "An empty or null name was provided."
-                ));
+                                                                            ));
```
howardrigberg, line 486: Is the future deletion of this file and related dependencies ticketed?

api-core/src/main/java/com/vimeo/networking2/account/CachingAccountStore.kt
```diff
@@ -0,0 +1,49 @@
...
+ */
+package com.vimeo.networking2.account
+
+import com.vimeo.networking2.VimeoAccount
+
+/**
+ * An [AccountStore] that caches the stored account for fast retrieval and delegates actual storage to another
+ * [AccountStore].
+ *
+ * @param actualAccountStore The store which will be used to persist the account outside the lifecycle of this
```
howardrigberg, line 30: Unfinished thought?

api-core/src/main/java/com/vimeo/networking2/account/AccountStore.kt
```diff
@@ -19,29 +19,28 @@
...
+    fun storeAccount(vimeoAccount: VimeoAccount)
 
     /**
-     * Delete the provided account from the store.
-     *
-     * @param vimeoAccount The account to delete.
+     * Delete the stored account.
      */
-    fun deleteAccount(vimeoAccount: VimeoAccount)
+    fun removeAccount()
```
howardrigberg, line 45: Not a big deal, but I noticed some inconsistency as to whether or not there is a line between this final function and the trailing interface brackets (see VimeoCalllback). Do you have a preference?

api-core/src/main/java/com/vimeo/networking2/account/InMemoryAccountStore.kt
```diff
@@ -19,7 +19,9 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-package com.vimeo.networking2
+package com.vimeo.networking2.account
+
+import com.vimeo.networking2.VimeoAccount
 
 /**
  * An account store that just holds the accounts in memory.
```
howardrigberg, line 27: Do you have a preference with leading lines for the first method inside of a class/interface? Other places have leading lines but this class doesn't.

api-core/src/main/java/com/vimeo/networking2/config/Configuration.kt
```diff
@@ -82,15 +88,20 @@ data class Configuration(
...
      * The builder for the [Configuration].
      *
      * @param clientId The Vimeo API client ID, required to construct an instance.
      * @param clientSecret The Vimeo API client secret, required to construct an instance.
-     * @param scope The permission scopes requested by the client, required to construct an instance.
+     * @param scopes The permission scopes requested by the client, required to construct an instance.
      */
-    class Builder(private val clientId: String, private val clientSecret: String, private val scope: Scopes) {
-        private var baseUrl: String = DEFAULT_BASE_URL
+    class Builder(private val clientId: String, private val clientSecret: String, private val scopes: List<ScopeType>) {
```
howardrigberg: Could write tests for this builder.

api-core/src/main/java/com/vimeo/networking2/config/ConfigurationExtensions.kt
```diff
@@ -0,0 +1,13 @@
+@file:JvmName("ConfigurationUtils")
```
howardrigberg: Licensing?

api-core/src/main/java/com/vimeo/networking2/config/RetrofitSetupModule.kt
```diff
@@ -24,123 +24,87 @@ package com.vimeo.networking2.config
...
-            val request = chain.request()
-            val builder = request.newBuilder()
-
-            if (request.header(AUTHORIZATION_HEADER).isNullOrBlank()) {
-                builder.addHeader(AUTHORIZATION_HEADER, "Bearer $accessToken")
-            }
-            chain.proceed(builder.build())
-        }
+    @JvmStatic
+    fun clearRequestCache(configuration: Configuration) {
```
howardrigberg: This method is not called anywhere internally in the library, nor is it called at the application level refactor. What is the context in which this method should be used?

api-core/src/main/java/com/vimeo/networking2/config/Configuration.kt
```diff
@@ -70,7 +76,7 @@ data class Configuration(
 
     val userAgent: String,
 
-    val requestTimeoutSeconds: Int,
+    val requestTimeoutSeconds: Long,
```
howardrigberg: Optional: I realize that this is a bit outside of the scope of this ticket, but it would be better if we changed the name of `interceptors` to `applicationInterceptors` to align with Square's nomenclature. It would be a little clearer for consumers.

Also wondering whether we want to consider changing the name of this class to something like `VimeoApiConfiguration`. It is a little confusing as-is because it clashes with `VimeoApiClient.fetchAppConfiguration`. Just a thought...

api-core/src/main/java/com/vimeo/networking2/internal/ResponseExtensions.kt
```diff
@@ -0,0 +1,70 @@
+package com.vimeo.networking2.internal
```
howardrigberg, line 22: Licensing?

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallAdapter.kt
```diff
@@ -30,87 +31,58 @@ import retrofit2.Response
...
     private val call: Call<T>,
-    private val callbackExecutor: Executor?,
-    private val responseBodyConverter: Converter<ResponseBody, ApiError>
+    private val callbackExecutor: Executor,
+    private val responseBodyConverter: Converter<ResponseBody, ApiError>,
+    private val vimeoLogger: VimeoLogger
 ) : VimeoCall<T> {
 
+    /**
+     * Determine is the response has a body.
```
howardrigberg: is -> if

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallEmptyResponseAdapter.kt
```diff
@@ -0,0 +1,66 @@
+package com.vimeo.networking2.internal
```
howardrigberg, line 22: Licensing?

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCallEmptyResponseAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+import okhttp3.ResponseBody
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Converter
+import retrofit2.Response
+import java.util.concurrent.Executor
+
+/**
+ * A custom Retrofit call adapter for requests with an expected empty response. This was created in order to parse an
+ * error response from the API in a background thread and return the result in the calling thread.
```
howardrigberg: "in the calling thread." -> "in the thread determined by the [callbackExecutor]." ?

api-core/src/main/java/com/vimeo/networking2/internal/interceptor/AcceptHeaderInterceptor.kt
```diff
@@ -32,11 +32,15 @@ import okhttp3.Response
...
+            HEADER_ACCEPT,
+            HEADER_ACCEPT_VALUE
+        ).build()
+    )
 
     companion object {
         private const val HEADER_ACCEPT = "Accept"
         private const val HEADER_ACCEPT_VALUE: String =
-                "application/vnd.vimeo.*+json; version=${ApiConstants.API_VERSION}"
+            "application/vnd.vimeo.*+json; version=${ApiConstants.API_VERSION}"
```
howardrigberg, line 44: Why did we move away from using an interceptor for the token but kept it for the version?

api-core/src/main/java/com/vimeo/networking2/internal/params/ScopesConverter.kt
```diff
@@ -29,5 +29,5 @@ import retrofit2.Converter
  */
 class ScopesConverter : Converter<Scopes, String> {
     override fun convert(value: Scopes): String =
-            value.scopes.joinToString(separator = " ", transform = { it.name.toLowerCase() })
+            value.scopes.joinToString(separator = " ", transform = { it.value.toLowerCase() })
```
howardrigberg: `it.value.toLowerCase(Locale.ROOT)` will get rid of the lint warning

api-core/src/main/java/com/vimeo/networking2/internal/params/StringValueConverter.kt
```diff
@@ -0,0 +1,11 @@
+package com.vimeo.networking2.internal.params
```
howardrigberg, line 24: Licensing?

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -43,37 +45,36 @@ import okhttp3.Credentials
...
- *
- *       override fun onSuccess(authResponse: VimeoResponse.Success<BasicAuthToken>) {
- *
- *       }
- *
- *       override fun onError(error: VimeoResponse.Error) {
- *
- *       }
- * })
+ * authenticator.clientCredentials(vimeoCallback(
```
howardrigberg, line 47: In docs above:

"API that allow you to make the following authentication requests:"

Maybe modify to something like:

"Interface that provides the following ways to authenticate with the Vimeo api:"

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -43,37 +45,36 @@ import okhttp3.Credentials
...
 
     /**
-     * Authenticate client id and client secret.
+     * The currently authenticated account, if the client has successfully authenticated. May represent either a logged
+     * in or logged out user.
+     */
+    val currentAccount: VimeoAccount?
+
+    /**
+     * Authenticate using the client id and client secret to obtain a logged out token.
```
howardrigberg: Maybe provide a little more context for where the client id/secret come from:

"Authenticate using the client id and client secret (as set in the Configuration) to obtain a logged out token."

api-core/src/main/java/com/vimeo/networking2/internal/ErrorHandlingUnitCallAdapter.kt
```diff
@@ -0,0 +1,28 @@
...
+ * @param responseBodyConverter Converter to parse the error response into a [ApiError].
+ * @param vimeoLogger The logger used to log information about error handling.
+ */
+internal class ErrorHandlingUnitCallAdapter(
+    private val callbackExecutor: Executor,
+    private val responseBodyConverter: Converter<ResponseBody, ApiError>,
+    private val vimeoLogger: VimeoLogger
+) : CallAdapter<Unit, VimeoCall<Unit>> {
+    override fun adapt(call: Call<Unit>): VimeoCall<Unit> =
+        VimeoCallEmptyResponseAdapter(call, callbackExecutor, responseBodyConverter, vimeoLogger)
```
howardrigberg, line 25: Run the formatter on this file.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -134,33 +135,69 @@ interface Authenticator {
...
-                        serverConfig.clientSecret
-                    )
-
-            return AuthenticatorImpl(authService, authHeaders, Scopes(serverConfig.scopes))
+        operator fun invoke(configuration: Configuration): Authenticator {
+            val authService = RetrofitSetupModule.retrofit(configuration).create(AuthService::class.java)
+            val basicAuthHeader: String = Credentials.basic(configuration.clientId, configuration.clientSecret)
+
+            return AuthenticatorImpl(
+                authService,
```
howardrigberg: Run the formatter on this file.

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -228,249 +220,276 @@ interface VimeoService {
...
-        @Body parameters: Map<String, String>
+        @Body parameters: Map<String, @JvmSuppressWildcards String>
     ): VimeoCall<Unit>
 
     @Headers(HEADER_NO_CACHE)
     @GET("search")
     fun search(
         @Header(AUTHORIZATION) authorization: String,
-        @QueryMap queryParams: Map<String, String>
+        @QueryMap queryParams: Map<String, @JvmSuppressWildcards String>
```
anthonycr, line 513: There are two generic parameter types supported in this service `@QueryMap` and `@Body`. Ideally, we wouldn't have any generic parameters and we would have the fully supported type here. However, we don't have the time to do that for this iteration, so we have to live with the generic parameters for now. The issue I have run into is regarding consistency in naming and the supported types.

`@QueryMap` is always of the type `Map<String, String>`, but the name differs in this file between `queryParams` and `options`. Additionally, it is almost always called `refinementMap` in the `VimeoApiClient`. So my question is, what name should we choose for the `@QueryMap`? It is almost exclusively used to refine the results coming back from the API.

`@Body` varies between being of type `Any`, `List<Any>`, `Map<String, String>`, and `Map<String, Any>`, as well as the concrete types that fully encapsulate the values in a type safe way. The names differ between `parameters` and `body` as well as `postBody` and `patchBody` in the `VimeoApiClient`.

So... how can we consolidate these names and types? The differences are mostly leftover from the old `VimeoClient` and `VimeoService` which served as the basis for these classes.

@WhosNickDoglio @howardrigberg

anthonycr: My personal suggestion would be to choose `body` as the single name used for `@Body` and change the type to `Any` for anything that isn't a fully enumerated body DTO. I'm not sure what the name for `@QueryMap` params should be.

howardrigberg:
@anthonycr I vote for: `queryParams` and `bodyParams` or if you are allergic to brevity: `queryParameters` and `bodyParameters`. If I imagine how someone external to the company might use them, I think this is the clearest and most descriptive. I think of someone going to a reference like: https://developer.vimeo.com/api/reference/videos#edit_video

If they wanted to edit part of a video like "description", they'd see it in the chart identified as a body parameter. Likewise with query string parameters. They are always described as "query parameters".

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -117,50 +125,90 @@ interface Authenticator {
...
-        authCallback: VimeoCallback<VimeoAccount>
+        callback: VimeoCallback<VimeoAccount>
     ): VimeoRequest
 
+    /**
+     * Exchange an OAuth 1 token and secret for a new OAuth 2 token.
+     *
+     * @param token The old token to use in the exchange.
+     * @param tokenSecret The old token secret to use in the exchange.
+     * @param callback Callback to be notified of the result of the request.
```
howardrigberg, line 151: [VimeoCallback] to be notified...

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -117,50 +125,90 @@ interface Authenticator {
...
+
+            return AuthenticatorImpl(
+                authService,
+                basicAuthHeader,
+                configuration.scope,
+                CachingAccountStore(configuration.accountStore)
+            )
         }
+
+        private val delegate: MutableAuthenticatorDelegate = MutableAuthenticatorDelegate()
```
howardrigberg, line 175: Doesn't our style dictate this at the top of the object?

auth/src/main/java/com/vimeo/networking2/internal/MutableAuthenticatorDelegate.kt
```diff
@@ -0,0 +1,79 @@
...
+import com.vimeo.networking2.VimeoAccount
+import com.vimeo.networking2.VimeoCallback
+import com.vimeo.networking2.VimeoRequest
+
+/**
+ * An [Authenticator] that delegates its implementation to an internal mutable instance [actual].
+ *
+ * @param actual The actual implementation of [Authenticator], defaults to null.
+ */
+class MutableAuthenticatorDelegate(var actual: Authenticator? = null) : Authenticator {
```
howardrigberg, line 37: internal?

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -117,50 +125,90 @@ interface Authenticator {
...
+
+        /**
+         * Initialize the singleton instance of the [Authenticator] with a [Configuration]. If the authenticator was
+         * already initialized, this will reconfigure it. This function must be called before [instance] is used.
+         *
+         * @param configuration The configuration used by the authenticator.
+         */
+        @JvmStatic
+        fun initialize(configuration: Configuration) {
+            delegate.actual = Authenticator(configuration)
```
howardrigberg: So, re-initialization is allowed? Will it cause any other downstream problems?

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -36,28 +34,26 @@ import retrofit2.http.POST
...
      *
-     * @return A [VimeoCall] that provides a [BasicAccessToken] that can be used to perform unauthenticated requests.
+     * @return A [VimeoCall] that provides a [VimeoAccount] that can be used to perform unauthenticated requests.
      */
     @FormUrlEncoded
     @POST("oauth/authorize/client")
     fun authorizeWithClientCredentialsGrant(
         @Header(AUTHORIZATION) authorization: String,
         @Field(GRANT_TYPE) grantType: GrantType,
         @Field(SCOPE) scope: Scopes
```
howardrigberg: Shouldn't "scope" be "scopes" throughout this file?

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -67,7 +63,7 @@ internal interface AuthService {
      * @param email The email of the account that should be created.
      * @param password The password of the account that should be created.
      * @param scope The permissions scope that should be granted to the client.
-     * @param params Extra parameters associated with the account creation, such as marketing opt in.
+     * @param marketingOptIn True if the use is opting into marketing emails, false otherwise.
```
howardrigberg: "True if the user..."

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -77,30 +73,54 @@ internal interface AuthService {
...
     ): VimeoCall<VimeoAccount>
 
+    /**
+     * Used to sign up for Vimeo using a Facebook authorization token.
+     *
+     * @param authorization Created from the client id and client secret.
+     * @param email The email the user uses to log into Facebook.
+     * @param token The Facebook token used to authorize with Facebook.
+     * @param scope The permissions scope that should be granted to the client.
+     * @param marketingOptIn True if the use is opting into marketing emails, false otherwise.
```
howardrigberg: "True if the user..."

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -67,7 +63,7 @@ internal interface AuthService {
      * @param email The email of the account that should be created.
      * @param password The password of the account that should be created.
      * @param scope The permissions scope that should be granted to the client.
-     * @param params Extra parameters associated with the account creation, such as marketing opt in.
+     * @param marketingOptIn True if the use is opting into marketing emails, false otherwise.
      *
      * @return A [VimeoCall] that provides a [VimeoAccount] that can be used to perform authenticated requests and also
      * contains a user object.
```
howardrigberg, line 69: Do these endpoints support a field filter? If I recall correctly this was a source of subtle bugs in the app in the past. The initial user object was unfiltered but subsequent objects would be filtered.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -143,15 +185,38 @@ internal interface AuthService {
...
     fun authenticateWithCodeGrant(
         @Header(AUTHORIZATION) authorization: String,
-        @Field("redirect_uri") redirectURI: String,
+        @Field("redirect_uri") redirectUri: String,
         @Field("code") code: String,
         @Field(GRANT_TYPE) grantType: GrantType
     ): VimeoCall<VimeoAccount>
 
+    /**
+     * Exchanges an old OAuth One token for a shiny OAuth2 token.
```
howardrigberg: "a shiny NEW OAuth2 token"

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -22,27 +22,33 @@
...
+        get() = accountStore.loadAccount()
+
+    override fun clientCredentials(callback: VimeoCallback<VimeoAccount>): VimeoRequest {
         val params = mapOf(
             AuthParam.FIELD_GRANT_TYPE to GrantType.CLIENT_CREDENTIALS.value,
             AuthParam.FIELD_SCOPES to scopes
         )
         val call = authService.authorizeWithClientCredentialsGrant(
-            authorization = authHeader,
+            authorization = basicAuthHeader,
```
howardrigberg, line 52: Run the formatter on this file

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -54,17 +60,17 @@ internal class AuthenticatorImpl(
                 "Client credentials authentication error.",
```
howardrigberg: `developerMessage = "Client credentials authentication error."`

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -156,25 +162,24 @@ internal class AuthenticatorImpl(
                 "Email join error.",
```
howardrigberg: `developerMessage = "Email join error."`

models/src/main/java/com/vimeo/networking2/params/SearchDateType.kt
```diff
@@ -0,0 +1,13 @@
+package com.vimeo.networking2.params
```
howardrigberg, line 24: licensing?

models/src/main/java/com/vimeo/networking2/params/SearchDateType.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.vimeo.networking2.enums.StringValue
+
+/**
+ * The search options for the time range in which a video was uploaded.
```
howardrigberg, line 29: Is it exclusively used for videos?

models/src/main/java/com/vimeo/networking2/params/SearchDurationType.kt
```diff
@@ -0,0 +1,12 @@
+package com.vimeo.networking2.params
```
howardrigberg, line 24: licensing?

models/src/main/java/com/vimeo/networking2/params/SearchFacetType.kt
```diff
@@ -0,0 +1,15 @@
+package com.vimeo.networking2.params
```
howardrigberg, line 24: licensing?

models/src/main/java/com/vimeo/networking2/params/SearchFilterType.kt
```diff
@@ -0,0 +1,14 @@
+package com.vimeo.networking2.params
```
howardrigberg, line 24: licensing?

models/src/main/java/com/vimeo/networking2/params/SearchSortDirectionType.kt
```diff
@@ -0,0 +1,12 @@
+package com.vimeo.networking2.params
```
howardrigberg, line 24: licensing

models/src/main/java/com/vimeo/networking2/params/SearchSortType.kt
```diff
@@ -0,0 +1,16 @@
+package com.vimeo.networking2.params
```
howardrigberg, line 24: licensing?

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+ * final Authenticator authenticator = Authenticator.create(Configuration)
+ * final VimeoApiClient client = VimeoApiClient.create(Configuration, authenticator)
+ * ```
+ * For consumers that don't want to manage the instance of the client themselves, there is also a convenient singleton
+ * instance that can be initialized and used.
+ * ```
+ * Authenticator.initialize(Configuration)
+ * VimeoApiClient.initialize(Configuration, Authenticator.instance())
+ * val clientInstance = VimeoApiClient.instance()
+ * ```
```
howardrigberg, line 68: Maybe it would also be helpful to mention authentication here as well.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+         */
+        @JvmStatic
+        @JvmName("create")
+        operator fun invoke(configuration: Configuration, authenticator: Authenticator): VimeoApiClient {
+            val retrofit = RetrofitSetupModule.retrofit(configuration)
+            val vimeoService = retrofit.create(VimeoService::class.java)
+            val basicAuthHeader = Credentials.basic(configuration.clientId, configuration.clientSecret)
+            val synchronousExecutor = Executor { it.run() }
+            return VimeoApiClientImpl(
+                vimeoService,
```
howardrigberg, line 1504: Run the formatter on this file

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+ * Authenticator.initialize(Configuration)
+ * VimeoApiClient.initialize(Configuration, Authenticator.instance())
+ * val clientInstance = VimeoApiClient.instance()
+ * ```
+ */
+@Suppress("ComplexInterface", "LongParameterList")
+interface VimeoApiClient {
+
+    /**
+     * Create an album.
```
howardrigberg, line 74: A few general thoughts on the documentation here. 

1. Perhaps we could reference the public Vimeo API documentation for each of these functions? It would be incredibly helpful for external consumers of the library. The only drawback would be if the links changed in the future. For example, this method would link out to: https://developer.vimeo.com/api/reference/showcases#create_showcase
2. Related to your other general question, should we consider naming `parameters` as `bodyParameters`? The only source that an external user would consult for information about what to pass here is our documentation which refers to these as "body parameters".
3. Why no mention of the `@return A [VimeoRequest] object that can be used to cancel the request`?
4. I am not sure how we should handle the "Album" vs. "Showcase" naming issue. Our api docs and the Vimeo website call everything showcases, but the api itself still uses albums on endpoints and such. It's not necessarily something we have to do anything about, but maybe we could at least address the issue in kdocs? I was thinking that maybe we could do something like refer to the object as album/showcase in the kdocs. I am reluctant to change the naming in code because there will still end up being a naming discrepancy at some level.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+
+    /**
+     * Update the like interaction of a [Video].
+     *
+     * @param isLiked True if the video should be liked, false otherwise.
+     * @param video The [Video] which should be liked or disliked.
+     * @param password The optional password that will be needed to like the [Video] if it is password protected.
+     * @param callback The callback which will be notified of the request completion.
+     */
+    fun updateVideoLike(
```
howardrigberg, line 551: Maybe we can remove this redundant endpoint? We are not using it in our code.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+    /**
+     * Update the watch later interaction of a [Video].
+     *
+     * @param isWatchLater True if the video should be added to watch later, false otherwise.
+     * @param video The [Video] which should be added/removed to/from watch later.
+     * @param password The optional password that will be needed to add the the [Video] watch later if it is password
+     * protected.
+     * @param callback The callback which will be notified of the request completion.
+     */
+    fun updateVideoWatchLater(
```
howardrigberg, line 587: Maybe we can remove this redundant endpoint? We are not using it in our code.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+        password: String?,
+        callback: VimeoCallback<Unit>
+    ): VimeoRequest
+
+    /**
+     * Update the watch later interaction of a [Video].
+     *
+     * @param isWatchLater True if the video should be added to watch later, false otherwise.
+     * @param uri The URI of the [Video].
+     * @param password The optional password that will be needed to add the the [Video] watch later if it is password
```
howardrigberg: remove extra "the"

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+        password: String?,
+        callback: VimeoCallback<Unit>
+    ): VimeoRequest
+
+    /**
+     * Update the watch later interaction of a [Video].
+     *
+     * @param isWatchLater True if the video should be added to watch later, false otherwise.
+     * @param video The [Video] which should be added/removed to/from watch later.
+     * @param password The optional password that will be needed to add the the [Video] watch later if it is password
```
howardrigberg: remove extra "the"

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+
+    /**
+     * Create a comment on a [Video].
+     *
+     * @param video The video upon which the user is commenting.
+     * @param comment The content of the comment.
+     * @param password The optional password will be needed to comment on the [Video] if it is password protected.
+     * @param callback The callback which will be notified of the request completion.
+     */
+    fun createComment(
```
howardrigberg, line 621: Maybe we can remove this redundant endpoint? We are not using it in our code.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -0,0 +1,1184 @@
...
+     */
+    fun fetchProduct(
+        uri: String,
+        fieldFilter: String?,
+        cacheControl: CacheControl?,
+        callback: VimeoCallback<Product>
+    ): VimeoRequest
+
+    /**
+     * Fetch the user that is currently logged into the client.
```
howardrigberg: [User]

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -35,40 +36,41 @@ import retrofit2.http.PATCH
...
 import retrofit2.http.Url
 
 /**
  * Retrofit service for the Vimeo API.
  *
- * TODO: Add documentation to functions.
+ * @see VimeoApiClient
  */
 @Suppress("unused", "UndocumentedPublicFunction", "ComplexInterface")
 interface VimeoService {
```
howardrigberg: internal?

request/src/main/java/com/vimeo/networking2/internal/MutableVimeoApiClientDelegate.kt
```diff
@@ -0,0 +1,518 @@
...
+import com.vimeo.networking2.params.SearchDateType
+import com.vimeo.networking2.params.SearchDurationType
+import com.vimeo.networking2.params.SearchFacetType
+import com.vimeo.networking2.params.SearchFilterType
+import com.vimeo.networking2.params.SearchSortDirectionType
+import com.vimeo.networking2.params.SearchSortType
+import okhttp3.CacheControl
+
+/**
+ * A [VimeoApiClient] that delegates its implementation to an internal mutable instance [actual].
```
howardrigberg: Optional: it might help future developers to understand the need for this by explaining a bit about the initialization use case. It isn't obvious or intuitive.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -0,0 +1,835 @@
...
+            invalidParameters = listOf(InvalidParameter(
+                errorCode = ErrorCodeType.INVALID_URI.value
+            ))
+        ), callback)
+    }
+
+    private fun String?.asPasswordParameter(): Map<String, String> =
+        this?.let { mapOf(ApiConstants.Parameters.PARAMETER_PASSWORD to it) } ?: emptyMap()
+
+    private fun <T, V> Map<T, V>?.intoMutableMap(): MutableMap<T, V> {
```
howardrigberg: 👨‍🍳 💋 BUT! Style-wise, shouldn't the declaration be the following?
```
private fun <T, V> Map<T, V>?.intoMutableMap(): MutableMap<T, V> = 
            this?.toMutableMap() ?: mutableMapOf()
```

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -0,0 +1,835 @@
...
+
+    override fun createAlbum(
+        name: String,
+        albumPrivacy: AlbumPrivacy,
+        description: String?,
+        parameters: Map<String, Any>?,
+        callback: VimeoCallback<Album>
+    ): VimeoRequest {
+        val body = parameters.intoMutableMap()
+        body[ApiConstants.Parameters.PARAMETER_ALBUM_NAME] = name
```
howardrigberg, line 75: The logic here and in some other methods is different than in `VimeoClient` in that certain defensive client side code has been removed. For example, in the case of this method the name parameter was checked in `VimeoClient` to confirm it was not empty before making the network call. Can you explain the rationale? Excuse the repeated comment here, I brought this up at the app-level review and I'm not sure if you answered it yet because I haven't circled back there for a second round.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -0,0 +1,835 @@
...
+) : VimeoApiClient {
+
+    private val authHeader: String
+        get() {
+            val vimeoAccount = authenticator.currentAccount
+
+            return vimeoAccount?.accessToken?.let { "Bearer $it" } ?: basicAuthHeader
+        }
+
+    override fun createAlbum(
```
howardrigberg, line 67: Do we have plans to group and move related APIs (e.g. Album) out of VimeoApiClient into some encapsulating class? This could help consumers to navigate the API features more easily. We could group the features in a similar way to the organization that is presented in the API reference: https://developer.vimeo.com/api/reference

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -0,0 +1,835 @@
...
+        callback: VimeoCallback<Video>
+    ): VimeoRequest {
+        val body = parameters.intoMutableMap()
+        if (title != null) {
+            body[ApiConstants.Parameters.PARAMETER_VIDEO_NAME] = title
+        }
+        if (description != null) {
+            body[ApiConstants.Parameters.PARAMETER_VIDEO_DESCRIPTION] = description
+        }
+        if (password != null) {
```
howardrigberg, line 213: Follow up to my other question regarding validation, this is a good specific example. The previous code was doing a client side check to make sure that password was not null when the privacy type was set to password. Do we know what the client/server consequences are of letting errors like this one pass through?

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -0,0 +1,835 @@
...
+        }
+    }
+
+    override fun updateVideoLike(
+        isLiked: Boolean,
+        video: Video,
+        password: String?,
+        callback: VimeoCallback<Unit>
+    ): VimeoRequest {
+        val uri = video.uri ?: return localVimeoCallAdapter.enqueueEmptyUri(callback)
```
howardrigberg: Why just checking against non-null and not non-blank?

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -0,0 +1,835 @@
...
+    override fun fetchProductList(
+        fieldFilter: String?,
+        cacheControl: CacheControl?,
+        callback: VimeoCallback<ProductList>
+    ): VimeoRequest {
+        return vimeoService.getProducts(authHeader, fieldFilter, cacheControl).enqueue(callback)
+    }
+
+    override fun fetchProduct(
+        uri: String,
```
howardrigberg, line 602: Why not uniformly validate uri parameters? What happens if this uri is empty for example?

howardrigberg (changes requested):
Fantastic job! 🎉 

We need to update the readme and the `example-java-android` and `example-kotlin-android` for the new approach. Additionally, we should port the recent additions that I made for Folders to the new VimeoApiClient.

vimeo-networking/src/main/java/com/vimeo/networking/VimeoClient.java
anthonycr: Yes.

api-core/src/main/java/com/vimeo/networking2/VimeoCallback.kt
anthonycr: Are you suggesting remove "Create?" The usage of create is accurate, as it creates an instance of a callback.

api-core/src/main/java/com/vimeo/networking2/config/RetrofitSetupModule.kt
anthonycr: The method is used by the client application, so i think this comment is outdated.

api-core/src/main/java/com/vimeo/networking2/internal/interceptor/AcceptHeaderInterceptor.kt
anthonycr: The other headers that are set are fixed for the duration of of the lifecycle of a client instance. Originally when we were designing the new client, we thought that it would be better to set the authorization header in an interceptor. However, when building the new client, it didn't seem very practical. 

The reason is that an interceptor is injected into an OkHttp instance, which is injected into a Retrofit instance. This would create coupling between the interceptor instance and both the OkHttp instance and the `VimeoApiClient` instance. The relationships between the classes get a little complicated, and maintaining the proper states becomes difficult. The interceptor either needs to be mutable, rely on an `AccountStore`, or rely on the `Authenticator` instance. Making the interceptor mutable feels bad, but we can't use an `AccountStore` unless it is the same one used by the `Authenticator`. We can't use the `Authenticator` because it isn't created when we construct the interceptor. Overall, it seemed more straight forward to manage the auth state within the `Authenticator` and pass the header manually within the `VimeoApiClient`. We also avoid the problem of needing to use different auth headers for `VimeoApiClient` and `Authenticator`.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: Yes re-initialization is allowed, as mentioned in the documentation of this method, just like the previous implementation of `VimeoClient`. It is necessary for us to support re-initialization due to us configuring our client based on a request to the configs endpoint. We need to perform an initial initialization based on cache or static config, then reconfigure after making a request. 

It won't cause any downstream problems, as the use of delegates allows us to reconfigure the singleton on the fly, but it is possible to misuse. For example, if you were to turn off cert pinning in the `Authenticator` without reconfiguring the `VimeoApiClient`, then cert pinning will only be disabled for auth requests. The `VimeoApiClient.initialize` function takes an instance of a `Configuration` and an `Authenticator` in order to make it clear that the states of the two classes are not self updating. When we rewrite the readme we will make sure that the initialization sequence is clear.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: No. I want to support both `doThing(uri)` and `doThing(object)` functions for convenience in this class.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: No, see comment on video function.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: No, see comment on video like function.

api-core/src/main/java/com/vimeo/networking2/VimeoCallback.kt
howardrigberg: Sorry. I had a brain fart writing that. I meant to write "Creates a lambda friendly..." obviously a completely trivial and optional remark. I just thought we tend to write these from the perspective of what the function does rather than what the caller can do with the function.

Feel free to ignore it.

api-core/src/main/java/com/vimeo/networking2/internal/interceptor/AcceptHeaderInterceptor.kt
howardrigberg: That makes a lot of sense. Thanks for the clear explanation.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
anthonycr: I was hoping to, but at this point the lines are not clear enough to justify the increased complexity. There are a couple features that have clear boundaries like albums and connected apps, but most parts of the API are not fully developed. For example, channels utilize generic `fetchVideo`, `putContent`, and `deleteContent` functions. In order to create a "channels client," we would have to create separate `addToChannel`, `removeFromChannel`, `fetchVideoInChannel` functions. It's not feasible for us to do right now given time constraints. 

Given our limited resources, I'm not sure that it's completely responsible for us to do this in the near term, as it will increase the overall complexity (while improving discoverability) and will increase the maintenance costs associated with the library due to an increased number of classes and (presumably) new rules about adding new endpoints.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: From what I see on the API docs and in the API code, field filters are not supported on these endpoints.

api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
anthonycr: These are intended to be used in place of `Vimeo.java`. We don't need all the constants present in `Vimeo.java`, so I didn't add them all here, as a lot are related to parameters or enum values that are present in a different way. When we remove the old code from the library, we will migrate any necessary parameters to the new `ApiConstants` object.

api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
anthonycr: I'm going to work on de-duplicating these strings when I remove `Vimeo.java` from the library later.

api-core/src/main/java/com/vimeo/networking2/config/Configuration.kt
anthonycr: Tests are a good idea, so I created a ticket to write tests for `api-core` and `request`.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: It's redundant since the type is clearly visible in the docs that display in AS on hover. "Callback" is an accurate description of what it is. 

Screen Shot 2020-08-04 at 2 44 55 PM

models/src/main/java/com/vimeo/networking2/params/SearchDateType.kt
anthonycr: Yes it seems to be only used for videos.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: Authentication would be better discussed in the `Authenticator` where it is already discussed or in the readme, which will be rewritten at another time.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: 1. That would be helpful, I think it's something we could do later to improve the library. We don't have to stop with the refactoring here!
2. Done, renamed everything `queryParams` or `bodyParams`.
3. Done.
4. Perhaps you could address this yourself after this PR is merged? The DTOs have been named `Album` for some time, so renaming them to `Showcase` or writing documentation to explain the difference feels a little out of scope, as it's not something we have acknowledged before in the library.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
anthonycr: It's my opinion that the defensive client side code was redundant and better off left out. It complicates the client logic and runs the risk of being out of date with the actual API requirements. My position is that it is better to let the API fail the request if there is an issue with a parameter. Now certain errors we must fail for client side, specifically those where there is a type mismatch. 

Coincidentally, the parameter this comment is on is a good demonstration of the above logic. The check in the old `VimeoClient` that prevented an album being created with an empty name was unnecessary. The API accepts empty names in the create request and substitutes "Untitled" in its place. I'm not sure if it was originally an invalid parameter when the old `VimeoClient` code was written, but I think it is a better idea to let the API tell us of such an issue rather than make those determinations here as much as possible.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
anthonycr: In the case of passing a null password with a privacy set to password, the API response as expected with an error indicating that the password is missing. Response below:
```
{
  "invalid_parameters": [
    {
      "field": "password",
      "error_code": 2233,
      "error": "Please enter a password for this video.",
      "developer_message": "User is attempting to set video privacy to `password` without providing a password."
    }
  ],
  "error": "You have provided an invalid parameter. Please contact developer of this application.",
  "link": null,
  "developer_message": "The parameters passed to this API endpoint didn't pass Vimeo's validation. Please check the invalid_parameters list for more information.",
  "error_code": 2204
}
```

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
anthonycr: Added in a `.notEmpty()` extension that filters out the URI if it is null or blank.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
howardrigberg: @anthonycr I agree less client validation logic as a general directive, but I still think there is more that can be done here in cases that are simply going to fail (for example, an empty request uri). There is never a case where API requirements will accommodate that. 

Additionally, I was also pointing this out because unless the removal of client-side logical changes have been individually checked for their impact on the client app, we are vulnerable to new bugs and/or crashes when these edge cases occur. I'm not sure whether you've vetted each of these instances or if it is just a risk you are willing to take.

anthonycr:
@WhosNickDoglio @howardrigberg I've finished making changes to this PR, please rereview.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -0,0 +1,977 @@
...
+        ), callback)
+    }
+
+    /**
+     * @return The [String] if it is not empty or blank, otherwise returns null.
+     */
+    private fun String?.notEmpty(): String? = this?.takeIf { it.isNotBlank() }
+
+    private fun <T : StringValue> T.validate(): T =
+        this.takeIf { it.value?.isNotEmpty() == false } ?: error(INVALID_ENUM_MESSAGE)
```
howardrigberg: gonna break stuff with this one as is. should be:

 `this.takeIf { it.value?.isNotEmpty() == true } ?: error(INVALID_ENUM_MESSAGE)`

auth/src/main/java/com/vimeo/networking2/internal/MutableAuthenticatorDelegate.kt
WhosNickDoglio: @anthonycr Any reason this isn't `internal`?

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -21,163 +21,167 @@
...
 import retrofit2.http.QueryMap
 import retrofit2.http.Url
 
 /**
  * Retrofit service for the Vimeo API.
  *
- * TODO: Add documentation to functions.
+ * @see VimeoApiClient
  */
 @Suppress("unused", "UndocumentedPublicFunction", "ComplexInterface")
```
WhosNickDoglio: Can we remove "unused" from the Supression? Assuming since this class is internal all its method should be exposed via `VimeoApiClient` right?

Removing it for me locally shows `getVideoStatus()`  is the only unused method, this should be expose via the client right?

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -21,163 +21,167 @@
...
     @GET("me/connected_apps/{type}")
     fun getConnectedApp(
         @Header(AUTHORIZATION) authorization: String,
-        @Path("type") type: String,
+        @Path("type") type: ConnectedAppType,
+        @Query(PARAMETER_GET_FIELD_FILTER) fieldFilter: String?,
         @Header(CACHE_CONTROL) cacheControl: CacheControl?
     ): VimeoCall<ConnectedApp>
 
     @PUT("me/connected_apps/{type}")
```
howardrigberg, line 80: This needs `@FormUrlEncoded`

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -21,163 +21,167 @@
...
-        @Body parameters: Map<String, Any>
-    ): VimeoCall<Unit>
-
     @PATCH
     fun editVideo(
         @Header(AUTHORIZATION) authorization: String,
         @Url uri: String,
-        @Body parameters: Map<String, Any>
+        @Body bodyParams: Map<String, @JvmSuppressWildcards Any>
     ): VimeoCall<Video>
```
howardrigberg, line 158: This needs @FormUrlEncoded

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -21,163 +21,167 @@
...
     fun editUser(
         @Header(AUTHORIZATION) authorization: String,
         @Url uri: String,
-        @Body parameters: Map<String, Any>
+        @Field(PARAMETER_USERS_NAME) name: String?,
+        @Field(PARAMETER_USERS_LOCATION) location: String?,
+        @Field(PARAMETER_USERS_BIO) bio: String?
     ): VimeoCall<User>
 
     @PATCH
```
howardrigberg, line 170: This needs @FormUrlEncoded

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -21,163 +21,167 @@
...
     ): VimeoCall<PictureCollection>
 
     @PATCH("me/notifications/subscriptions")
     fun editNotificationSubscriptions(
         @Header(AUTHORIZATION) authorization: String,
-        @Body parameters: Map<String, Boolean>
+        @Body subscriptionMap: Map<NotificationType, Boolean>
     ): VimeoCall<NotificationSubscriptions>
 
     @POST
```
howardrigberg, line 184: This needs @FormUrlEncoded

howardrigberg (changes requested):

anthonycr:
@WhosNickDoglio @howardrigberg thanks, fixed those issues.

howardrigberg (approved):
Nice job with this, @anthonycr! You covered a lot of ground and made pragmatic improvements that put us in a much better position with the library going forward. Thanks for your contribution. Another great milestone!

WhosNickDoglio (approved):
👏👏👏

LGTM! This looks great, huge step forward from the previous client.

PR #441, opened 2020-07-22
MPS-921: Integrate Model Generator Plugin

WhosNickDoglio:
#### Issue
[MPS-921](https://vimean.atlassian.net/browse/MPS-921)

#### Summary
Primary changes: 
- Adding the Model generator gradle plugin along with the integration tests.
- Creating empty modules for `Parcelable` and `Serializable` models that will be generated using new plugin.
- Update the library to no longer expose the base models transitively (`api` to `compileOnly`) so consumers now to need to explicitly declare which models they want to use, `models`, `models-serializable` or `models-parcelable`.

Minor Changes: 
- Migrate to AndroidX (Necessary for `Parcelable` module to target API 29).
- Bump AGP to 3.6.3 and Gradle to 5.6.4 (Necessary because Model generator plugin uses newer APIs then in the version of Gradle we had).
- Bumped Kotlin to 1.3.72 for methods used in the integration tests. 
- Added a `publish_android.gradle` file to enable publishing the `models-Parcelable` artifact.

#### How to Test
Run `gradle build` and confirm all new models are generated without issues.  I'm also making an internal PR to use the new models that will verify them as well.


anthonycr (approved):
Looks good, since I already reviewed the majority of the code (the plugin) elsewhere, this was a quick review.

PR #442, opened 2020-07-23
[TECH] Don't include generated files in detekt or git

WhosNickDoglio:
#### Summary
- Added generated directories to gitignore 
- Skipped apply detekt to modules that contain generated code

#### How to Test
Run `gradle build detekt` and confirm Android Studio doesn't yell at you.


anthonycr (approved):
Excellent

PR #445, opened 2020-07-28
Support response origin

anthonycr:
# Summary
Determine the origin of a response, either cache or network. This is useful when a `CacheControl` is not specified in a request, but the consumer might need to know how up to date the response is. Potentially unnecessary.


howardrigberg (approved):
nice useful addition!

PR #446, opened 2020-08-06
Remove old VimeoClient and related code in vimeo-networking

anthonycr:
# Summary
- Removed all code from `vimeo-networking` module. This module now simply aggregates `api-core`, `auth`, and `request` modules together. Consumers can rely on the modules individually, or they can use `vimeo-networking` like before. Now they are required to specify a `model-{type}` implementation. If they don't care about serialization, then they can just use `models`. If they are java consumers, then `models-serializable` might be useful, and if android consumers `models-parcelable`.

- Rewrote example. Example now showcases configuration of the `Authenticator` and `VimeoApiClient`, authentication via client credentials and code grant, and fetching the current user's videos. Example now resides in `example` and is only written in Kotlin.

- Updated the models plugin to only register tasks once. This fixes an issue with compiling the `models-parcelable` module.

- Renamed `Connection` to `BasicConnection`, and created a common `Connection` interface that can be shared by all connections, similar to the way `Interaction` works.

- Exposed moshi configuration in `RetrofitSetupModule`.

- Removed `example-android` and `example-java` that used old `VimeoClient`.

- Added support in `Authenticator` for
  - Log in via pin code.
  - Log in via code grant.
  - Exchanging one access token for another.


example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+    private fun VimeoAccount?.asLoginStatus() = when {
+        this == null -> "Logged Out - Unauthenticated"
+        this.isLoggedIn -> "Logged In - Authenticated"
+        else -> "Logged Out - Authenticated"
+    }
+
+    private fun toast(message: String) = Toast.makeText(this, message, Toast.LENGTH_SHORT).show()
+
+    companion object {
+        const val CLIENT_ID = "CLIENT ID"
```
howardrigberg, line 168: Some line comments here would be helpful to tell the user where/how to get the client id and secret.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -56,7 +58,11 @@ import okhttp3.Credentials
  * Authenticator.initialize(Configuration)
  * val instance = Authenticator.instance()
  * ```
+ *
+ * See [https://developer.vimeo.com/api/authentication] for more information about the different authentication
+ * approaches.
```
howardrigberg, line 97: It might also be helpful to point them at the example since that demonstrates how to authenticate with the client: `example/src/main/java/com/vimeo/example/MainActivity.kt`

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -85,6 +91,7 @@ interface Authenticator {
      *
      * @return A [VimeoRequest] object to cancel API requests.
      */
+    @Internal
```
howardrigberg, line 128: Yes! Great to see this annotation in action. Do you think we should ticket an audit of the networking library to tag any outstanding internals or do you feel confident that you tagged all of them?

example/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">Vimeo Networking 2 Example</string>
```
howardrigberg: Maybe something like: "Vimeo Networking Authorization Example"

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+        loginStatus.text = authenticator.currentAccount.asLoginStatus()
+
+        clientCredentials.setOnClickListener {
+            // Obtain logged out credentials which can be used to make logged out requests.
+            authenticator.clientCredentials(vimeoCallback(
+                onSuccess = {
+                    toast("Obtained client credentials.")
+                    loginStatus.text = authenticator.currentAccount.asLoginStatus()
+                },
+                onError = { toast("Couldn't obtain client credentials") }
```
howardrigberg: It would be a little more helpful and self documenting to handle `it.reason.errorCode == EMPTY_AUTHENTICATION` and give a warning to the user asking if they have set their `CLIENT_ID` and `CLIENT_SECRET`.

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+ */
+class MainActivity : AppCompatActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_main)
+
+        // Construct a configuration instance.
+        val configuration = VimeoApiConfiguration.Builder(CLIENT_ID, CLIENT_SECRET, listOf(ScopeType.PUBLIC))
+            .withCodeGrantRedirectUrl("https://example.com")
```
howardrigberg: It would be clearer to have a named static for this near the `CLIENT_ID` and `CLIENT_SECRET` with line comments that explain why and how it will be used.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -158,6 +183,53 @@ interface Authenticator {
...
 
+    /**
+     * Obtain the URI which can be used to log in the user and receive back a URI that can be exchanged using
+     * [authenticateWithCodeGrant] for a logged in account.
+     *
+     * @param responseCode The response code that can be used to identify the origin of the redirect URI.
+     *
+     * @return The URI which can be opened in a browser.
+     */
+    fun obtainCodeGrantAuthorizationUri(responseCode: Int): String
```
howardrigberg: Might want to indicate that it is ok to call this on the same thread. No network requests are made.

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+        // Initialize the authenticator instance with the configuration and the api client.
+        val authenticator = Authenticator(configuration)
+        val apiClient = VimeoApiClient(configuration, authenticator)
+
+        // If we are handling a code grant redirect, this URI will be present, otherwise it's a normal app start.
+        val redirectUri = intent?.data
+
+        if (redirectUri != null) {
+            // Log in using the code grant URI provided to us by the redirect.
+            authenticator.authenticateWithCodeGrant(redirectUri.toString(), vimeoCallback(
```
howardrigberg: I am confused about how this would ever get called with the current code. The browser is launched and then the user will be navigated to the redirect url that they created in the web console, but the app won't get called back with the redirectUri. Is there something I am missing here? 

If we don't want to deal with the complications around managing the extra code that would be required to do webview based auth, we could give instructions for the user to copy paste the url into a edittext manually and click a button to complete the flow manually. It is an ugly approach, but it gets the job done for demo purposes with a minimal amount of work. It could come along with line comments about how you might actually handle it for real in mobile code.

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+            authenticator.clientCredentials(vimeoCallback(
+                onSuccess = {
+                    toast("Obtained client credentials.")
+                    loginStatus.text = authenticator.currentAccount.asLoginStatus()
+                },
+                onError = { toast("Couldn't obtain client credentials") }
+            ))
+        }
+
+        login.setOnClickListener {
```
howardrigberg, line 77: It would be a little more work, but it would be nice to have a single login/logout button that changed purpose depending on the logged in state.

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        loginStatus.text = authenticator.currentAccount.asLoginStatus()
+
+        clientCredentials.setOnClickListener {
+            // Obtain logged out credentials which can be used to make logged out requests.
+            authenticator.clientCredentials(vimeoCallback(
+                onSuccess = {
+                    toast("Obtained client credentials.")
+                    loginStatus.text = authenticator.currentAccount.asLoginStatus()
```
howardrigberg: Maybe we could add a button that would hit staff picks or something else that required client credentials after successful authentication? It would be nice to illustrate the next step. Also, it might be illustrative to label that button as "use client credentials" rather than "get staff picks" because the focus here is on the authorization flow and that will act as a final confirmation step in the process. It also might help to put the related buttons for authorization code grant and client credentials into separate visual groups in the demo app.

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+            authenticator.logOut(vimeoCallback(
+                onSuccess = {
+                    toast("Successfully logged out.")
+                    loginStatus.text = authenticator.currentAccount.asLoginStatus()
+                },
+                onError = { toast("Unable to log out.") }
+            ))
+        }
+
+        getMyVideos.setOnClickListener {
```
howardrigberg, line 96: Alternatively, it might help to label the button as "use authorization code grant" rather than "get my videos" since the point is to validate and demonstrate the last step in the authorization process.

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+            // Fetch the currently logged in user's videos. Will fail if the user is not logged in.
+            val uri = authenticator.currentAccount?.user?.metadata?.connections?.videos?.uri ?: ""
```
howardrigberg: It would be helpful and illustrative to actively report when the user is logged out since you will know it at this point rather than blindly making the network request.

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+        this.isLoggedIn -> "Logged In - Authenticated"
+        else -> "Logged Out - Authenticated"
+    }
+
+    private fun toast(message: String) = Toast.makeText(this, message, Toast.LENGTH_SHORT).show()
+
+    companion object {
+        const val CLIENT_ID = "CLIENT ID"
+        const val CLIENT_SECRET = "CLIENT SECRET"
+        const val REQUEST_CODE = 12345
```
howardrigberg: Would be helpful to explain where and how the request code is used in a line comment either here or near `obtainCodeGrantAuthorizationUri` above. (i.e. The REQUEST_CODE will be passed as a `state` query string parameter to the redirect uri  that is provided in `withCodeGrantRedirectUrl` when logging in the user with code grant authorization. It can be used for security validation by the receiving code.)

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -158,6 +183,53 @@ interface Authenticator {
...
+
+    /**
+     * Log in using an authorization code grant received from the request to made to the URI returned by
+     * [obtainCodeGrantAuthorizationUri].
```
howardrigberg: Maybe explain the flow a little more here: You need to open the uri in a browser or webview and then pass the entire resultant redirect uri to this method.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -146,7 +148,8 @@ internal interface AuthService {
     ): VimeoCall<VimeoAccount>
 
     /**
-     * Used to log into Vimeo using a Facebook authorization token.
+     * Used to log into Vimeo using a Facebook authorization token. If an account does not exist, this request will
+     * fail. For this reason, [joinWithFacebook] is preferred since it also supports login.
```
howardrigberg: Why don't we just remove `logInWithFacebook` and `logInWithGoogle` and rename the other existing `joinWith...` methods as `loginOrJoinWithFacebook` and `loginOrJoinWithGoogle`? The `logInWith...` methods are internal and unused and it probably just confusing to keep them around at all at this point.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -185,12 +189,29 @@ internal interface AuthService {
         @Field(SCOPE) scopes: Scopes
     ): VimeoCall<VimeoAccount>
 
+    /**
+     * Creates a request to the code grant authorization endpoint. Used to direct the user to a URL where they can log
+     * in and then redirect back to the app.
+     *
+     * @param clientId The client id of the app.
+     * @param redirectUri The URI which should be redirected back to.
+     * @param state A random number that will be returned in the redirect to identify the origin of the request.
```
howardrigberg: An arbitrary value. We return this value to you later as a security measure in a query string parameter named `state` in the callback to the [redirectUri] you specified above.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -185,12 +189,29 @@ internal interface AuthService {
         @Field(SCOPE) scopes: Scopes
     ): VimeoCall<VimeoAccount>
 
+    /**
+     * Creates a request to the code grant authorization endpoint. Used to direct the user to a URL where they can log
+     * in and then redirect back to the app.
+     *
+     * @param clientId The client id of the app.
+     * @param redirectUri The URI which should be redirected back to.
```
howardrigberg, line 202: The uri needs to be preregistered at https://developer.vimeo.com/apps/ for your developer accoount with Vimeo.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -185,12 +189,29 @@ internal interface AuthService {
         @Field(SCOPE) scopes: Scopes
     ): VimeoCall<VimeoAccount>
 
+    /**
+     * Creates a request to the code grant authorization endpoint. Used to direct the user to a URL where they can log
+     * in and then redirect back to the app.
+     *
+     * @param clientId The client id of the app.
```
howardrigberg, line 201: The client identifier of your app. To get this value, access your app's information page from https://developer.vimeo.com/apps/.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
WhosNickDoglio: Just curious but why isn't this method or the related ones using `VimeoRequest` and `VimeoCallback` ?

example/src/main/java/com/vimeo/example/MainActivity.kt
WhosNickDoglio, line 77: I'm getting an error when I try this. 
Screenshot_1597088056

WhosNickDoglio (approved):
This looks great! 

Excited to have this fully in Kotlin. 

Left a question and an issue I noticed with the example 👍

vimeo-networking/src/main/java/com/vimeo/networking/AccountStore.java
```diff
@@ -1,40 +0,0 @@
-/*
```
howardrigberg, line 1: You left vimeo-networking/src/main/resources/keystore.bks. Is this intentional?

howardrigberg (changes requested):
Nice job, I've got a few line comments. Mostly on the example.

Also, wondering how you feel about changing `com.vimeo.networking2` to `com.vimeo.networking` now that all of the old stuff is deleted.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: Not sure what changes you made to the example before running it, but because the docs aren't clear - you need to create an API app here: https://developer.vimeo.com/apps/new and then register the redirect URI (in the example it's set to `https://example.com`) on the API page - navigate to https://developer.vimeo.com/apps and then choose your app and at the bottom click "add URL." Once this is done replace the client id and client secret placeholders in the example.

anthonycr:
Regarding changing the package name, we made the decision to use the package name `networking2` in order to facilitate migration to the new networking library, as this is a complete rewrite of the library. Since we aren't the only consumers, and everyone else is still on V1, it is better to leave it as is.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: Well this function is just constructing a URL so it doesn't enqueue a request or require a callback.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: The app fires off an intent to open the code grant URL in the browser. The website prompts the user to log in and authorize the app to access Vimeo. Once that is done, the browser navigates to the redirect URI specified in the configuration. The application has to register that URI in the manifest (see `AndroidManifest.xml` where `https://example.com` is registered in an intent filter). When the browser redirects to `https://example.com?code={auth code}` the app is reopened and the redirect URI is present in the `intent`. The application then extract the URI from the `intent` and pass it to `authenticateWithCodeGrant`, which extracts the authorization code and authenticates with the API. Since `MainActivity` is restarted when a new intent is fired, `onCreate` is called. If `MainActivity` was brought to the front instead, this code would need to be in `onNewIntent`.

The 3 things the user has to do is
1. Set the client ID in this file.
2. Set the client secret in this file.
3. Add `https://example.com` as a callback URL in the API dashboard.

Once those three things are done, this code works as expected.

howardrigberg: Ah. I see. Thanks for the explanation. I thought I might be missing something. That's definitely something that needs to be called out clearly in the docs. Did you test the entire flow to confirm that it works?

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: I changed the name to `createCodeGrantAuthorizationUri` and updated the docs a little. It feels a bit weird to call out this method as being thread safe when all the functions on `Authenticator` are thread safe and we don't call those out.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: I want to keep the example as simple as possible.

example/src/main/res/values/strings.xml
anthonycr: I'll change it to just `Vimeo Networking Example`. It also demonstrates how to make a request with the `VimeoApiClient` so it encompasses a little more than auth.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: Added documentation to the class definition instead that explains setting up the project.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
WhosNickDoglio: Ah, I saw the use of `authService` in `AuthenticatorImpl` and assumed it was a call, didn't notice the `request().url().toString()`, makes sense now 👍

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: This is an internal class, so this function doesn't need comprehensive docs as if it was publicly consumable. I added a link to the documentation URL to the `VimeoApiConfiguration` constructor and builder just to be safe.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
howardrigberg: FWIW, Nick's confusion around this was shared. Most of what the Authenticator does is make requests to our servers. This method is the exception. When I first saw it in action in the example, I wasn't immediately sure if the source of the string was from the server or not or why it was different. Once I looked at the implementation, I understood what was happening, but I was initially a little confused. That's why I made the thread suggestion initially. If you have any other ideas of docs that could call this out in some way, it would be helpful. Completely optional though.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: This is an internal class not accessible to the consumer so this documentation isn't necessary. There is already public documentation about this requirement on `VimeoApiConfiguration.codeGrantRedirectUri`.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
howardrigberg: Perfect! That is what I was literally about to suggest.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: This is an internal class that is not visible to the consumer, so I added this documentation to the `Authenticator` instead.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: Yes I created my own API app that I used to verify this example while building it. This is the same flow supported by the V1 library.

example/src/main/java/com/vimeo/example/MainActivity.kt
WhosNickDoglio: I was setting up the code grant URL wrong, updated docs helped me get it working properly. 👍

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: I want to keep them around as documentation of their existence and how they can be used in contrast with the `joinWith...` functions. Since `AuthService` is internal, this will not be confusing to the user, only us as maintainers. I added some more docs to these functions to explain that they are intentionally unused and added suppression of the unused warnings.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: Instead of linking to the `MainActivity` which is a bit awkward since the example isn't packaged with the library, I added more to the `Authenticator` documentation block about client credentials vs basic authentication vs code grant. We will be linking to the example in the readme as well.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: They definitely aren't all tagged, it's something we can tackle as we go along. I finished adding `@Internal` annotations to the `Authenticator` in a later PR, but we don't have any in the `VimeoApiClient`. I think that's fine for now as the user will be able to see in the response from the server that they don't have access to the content.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: Now that the `MainActivity` explicitly calls this out in the documentation at the top, this is no longer necessary?

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: I added a staff picks fetch and removed the client credentials button. This brings the example more in line with the original, where client credentials are implicitly requested at startup rather than by clicking a button. Now if the user clicks on the get my videos button without being logged in it will set the message to say to log in. However, for staff picks, the consumer will see that they can make the request without being logged in.

anthonycr: Overall, I want to keep the example as short as possible. The current example is less than 200 lines of code, and my goal is to keep the `MainActivity` under that. The longer the example and the more complicated it is, the harder it will be for users to grok.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: Discussed on the staff picks comment.

example/src/main/java/com/vimeo/example/MainActivity.kt
howardrigberg: I think it is still a helpful thing to have. People will make mistakes regardless of the docs (many never read them) and this would be a quick way to point them in the right direction.

anthonycr:
@howardrigberg @WhosNickDoglio thanks for the review, I improved the example and documentation so that configuring the example and using the `Authenticator` to perform a code grant is hopefully more clear now.

example/src/main/java/com/vimeo/example/MainActivity.kt
anthonycr: Since this isn't the only request that can fail due to the id and secret not being set, I am just going to add a dialog at the beginning of `onCreate` that stops them in their tracks.

howardrigberg: Great idea!

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -22,41 +22,76 @@
...
- *
- * Ex:
+ * If no authentication is performed, by default basic authentication will be used for requests made by this SDK. It is
+ * advised that client credentials are used instead of basic authentication if the consumer intends to use the API as a
+ * "logged out" user. If the consumer intends to use the API as a "logged in" user, then obtaining credentials using the
+ * code grant mechanism is the way to authenticate.
  *
+ * Create an instance of the [Authenticator] by using its invoke function as follows.
  * ```
  * val authenticator = Authenticator(Configuration)
```
howardrigberg, line 54: Example doesn't match description above. `invoke` is not... ahem... invoked.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -22,41 +22,76 @@
...
  * ```
+ *
+ * In order to obtain client credentials, make a request as follows.
+ * ```
+ * authenticator.clientCredentials(vimeoCallback(
+ *     onSuccess = { authResponse: VimeoResponse.Success<VimeoAccount> -> }
+ *     onError = { error: VimeoResponse.Error -> }
+ * ))
+ * ```
+ * Once this request succeeds, all subsequent requests with the client will use the token obtained.
```
howardrigberg: You could mention that in practice the credentials are automatically stored in the AccountStore associated with the Authenticator. It might provide some clarity about the magic at play with how the credentials are retained.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -22,41 +22,76 @@
...
+ * startActivity(intent)
+ * ```
+ * Once the user logs in at the URI, the API will redirect back to the redirect URI mentioned above.
+ * ```
+ * // Use the URI the browser redirects back to to log in.
+ * authenticator.authenticateWithCodeGrant(redirectedUri, vimeoCallback(
+ *     onSuccess = { authResponse: VimeoResponse.Success<VimeoAccount> -> }
+ *     onError = { error: VimeoResponse.Error -> }
+ * ))
+ * ```
```
howardrigberg, line 94: It would be helpful to also mention the need to register an `intent-filter` for the redirect uri in the manifest so your app will get called back.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -158,6 +212,58 @@ interface Authenticator {
...
+
+    /**
+     * Log in using an authorization code grant. Before authenticating with this function, the user should be directed
+     * to log into Vimeo in a browser by navigating to the URI returned by [createCodeGrantAuthorizationUri]. After the
+     * user logs into Vimeo, the API will redirect back to the URI specified in
+     * [VimeoApiConfiguration.codeGrantRedirectUri] along with some additional parameters. The consumer should intercept
+     * this URI and use it to authenticate by passing it to this function in the [uri] parameter.
+     *
+     * @param uri The URI which was redirected back to you. Must contain a `code` query parameter that contains the
+     * authorization code.
```
howardrigberg, line 240: and also a `state` query parameter that matches the response code set in createCodeGrantAuthorizationUri that can be used to validate that uri originated from your app

example/src/main/res/layout/activity_main.xml
```diff
@@ -0,0 +1,88 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="com.vimeo.example.MainActivity">
+
+    <Button
+        android:id="@+id/login"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="Login"
```
howardrigberg: "Log In"

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,176 @@
...
+                    loginStatus.text = authenticator.currentAccount.asLoginStatus()
+                },
+                onError = { toast("Unable to log out.") }
+            ))
+        }
+
+        getMyVideos.setOnClickListener {
+            // Fetch the currently logged in user's videos. Will fail if the user is not logged in.
+            val uri = authenticator.currentAccount?.user?.metadata?.connections?.videos?.uri
+                ?: return@setOnClickListener run { myVideosList.text = "Cannot fetch videos until user is logged in!" }
```
howardrigberg: Can you show it as a toast instead so it is consistent with the other errors?

example/src/main/java/com/vimeo/example/MainActivity.kt
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * In order to use this example, first you need to follow these steps:
+ * 1. Go to [https://developer.vimeo.com/apps] and create an API app if you haven't created one.
+ * 2. After creating the API app, find the client ID and client secret on the app page and replace the [CLIENT_ID] and
+ * [CLIENT_SECRET] constants with those values.
+ * 3. On the same page that had the client ID and client secret, add a callback URL. Ideally this URL is one that you
+ * own and not one that another app will be handling.
+ * 4. Replace [CODE_GRANT_REDIRECT_URL] with the URL added on the app page and replace intent filter in the
+ * `AndroidManifest.xml` to handle the set URL instead of `https://example.com`.
+ * 5. Run the app and make requests.
```
howardrigberg, line 49: This is perfect!

howardrigberg (approved):
Looks good, Anthony. I like the documentation improvements. Things are much clearer now. I made a few more suggestions to docs and code. Feel free to adopt as many as you agree with. I'm approving it now though.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: `Authenticator(Configuration)` calls `Authenticator.invoke(Configuration)` in Kotlin, it's syntactic sugar. The `invoke` function is an operator function like `plus` or `minus`.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: While it's true that the `state` parameter will be present in the redirect URI, we don't validate it for the consumer and leave it up to them to validate. Adding it as a requirement in the param doc is a bit misleading.

PR #448, opened 2020-08-12
Support enterprise SSO login

anthonycr:
# Summary
We want to support login for enterprise customers via SSO. This is an internal use only API, but is very similar in usage to the publicly accessible code grant authentication API (via `createCodeGrantAuthorizationUri` and `authenticateWithCodeGrant`).

The API consists of 3 functions that were added to the `Authenticator`.
- `fetchSsoDomain(domain: String, VimeoCallback<SsoDomain>)`
- `createSsoGrantAuthorizationUri(domain: SsoDomain, responseCode: String): String`
- `ssoCodeGrant(uri: String, marketingOptIn: Boolean, VimeoCallback<VimeoAccount>)`

`createSsoGrantAuthorizationUri` is very similar to `createCodeGrantAuthorizationUri` in both its definition and implementation. The main difference is that the SSO version requires an `SsoDomain` instance. The reason for this is that while the base code grant auth URI is fixed for public code grants, for enterprise SSO the base URI is provided by the API. As a result, the consumer needs to make a request to `fetchSsoDomain` before creating the auth URI.

`ssoCodeGrant` is almost identical to `authenticateWithCodeGrant` in definition, implementation, and usage. The only difference in definition is that the marketing opt in is used. The implementation differs in that it hits a different endpoint. The usage is the same in that it requires the consumer to receive a callback with a redirect URI and then pass that redirect URI to the `ssoCodeGrant` function.

## Changes
- Added `SsoDomain` DTO.
- Added functions to `Authenticator` mentioned in above summary and counterparts in the `AuthService`.
- Added `url` property to a `VimeoCall` that allows the `HttpUrl` of the request to be returned.
- Added convenience `enqueueWithAccountStore` extension function on `VimeoCall<VimeoAccount>` that stores the result of a successful request as the currently logged in account.
- Extracted all string literals in `AuthService` into constants (not sure if this is necessarily a good idea since it makes seeing the actual values more difficult?)
- Added new errors to `AuthParam`

## Testing
NOTE: `:example` can be easily configured to test this.

1. Configure the client with an id and secret that has access to enterprise SSO login. 
2. Make sure to set the redirect URI in the configuration. 
3. Configure the app to handle the redirect URI via an `intent-filter` in the manifest.
4. Make a request to `fetchSsoDomain` with a domain that is supported by enterprise login.
5. Create the auth URL using `createSsoGrantAuthorizationUri` and the domain received back from the previous request.
6. Direct the user to log in via the browser using the URL created.
7. Receive the redirect URL back.
8. Log in using `ssoCodeGrant` and the URL that you received in the redirect.


api-core/src/main/java/com/vimeo/networking2/internal/VimeoCall.kt
```diff
@@ -50,6 +51,11 @@ interface VimeoCall<T> {
      */
     fun cancel()
 
+    /**
+     * The [HttpUrl] to which this call will be made.
+     */
+    val url: HttpUrl
```
anthonycr: Perhaps this should be a `String` instead so that there is looser coupling to okhttp. `VimeoCall` is an internal class though, so it doesn't feel like violating the contract we have of not exposing implementation details.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -247,6 +247,51 @@ interface Authenticator {
...
+     *
+     * @param uri The URI which was redirected back to you. Must contain a `code` query parameter that contains the
+     * authorization code.
+     * @param marketingOptIn True if the user is opting into marketing emails, false otherwise.
+     * @param callback Callback to be notified of the result of the request.
+     *
+     * @return A [VimeoRequest] object to cancel API requests.
+     */
+    @Internal
+    fun ssoCodeGrant(
```
anthonycr: The other (public) code grant function is called `authenticateWithCodeGrant`. I did not call this `authenticateWithSsoCodeGrant` because I was unsure if it would confuse consumers or not, as the API is extremely similar both literally and semantically, but hits different endpoints.

It would be good to unify the naming scheme for authentication functions similar to how `VimeoApiClient` uses `create-`, `edit-`, `delete-`, and `fetch-` prefixes.

I would propose `authorizeWith-` as the designated prefix for all authentication functions. This would result in
- `authenticateWithClientCredentials`
- `authenticateWithGoogle`
- `authenticateWithFacebook`
- `authenticateWithEmailJoin`
- `authenticateWithEmailLogin`
- `exchangeAccessToken` - unchanged
- `exchangeOAuth1Token` - unchanged
- `authenticateWithCodeGrant` - unchanged
- `authenticateWithSsoCodeGrant` - unchanged, this should probably change so that it's not clashing with the other code grant function, but i don't have any ideas
- `authenticateWithPinCode` - unchanged
- `logOut`, `fetch-`, and `create-` functions remain unchanged.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -326,10 +406,26 @@ internal interface AuthService {
...
+        private const val TOKEN = "token"
+        private const val TOKEN_SECRET = "token_secret"
+        private const val DOMAIN = "domain"
+        private const val USER_CODE = "user_code"
+        private const val DEVICE_CODE = "device_code"
+        private const val ACCESS_TOKEN = "access_token"
+        private const val CODE = "code"
+        private const val CLIENT_ID = "client_id"
+        private const val ID_TOKEN = "id_token"
+        private const val USERNAME = "username"
```
anthonycr: I'm not sure it's worth pulling all these constants out. We originally did it with `AUTHORIZATION` since it is a constant header across all authentication methods, but maybe this could get out of hand. Not all parameters in `VimeoService` have been extracted for instance. I would like to either go all in or get rid of these constants as it looks sloppy to mix constants with inline values.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -230,24 +230,77 @@ internal class AuthenticatorImpl(
...
+        uri: String,
+        marketingOptIn: Boolean,
+        callback: VimeoCallback<VimeoAccount>
+    ): VimeoRequest {
+        val httpUrl: HttpUrl = HttpUrl.parse(uri) ?: return localVimeoCallAdapter.enqueueEmptyUri(callback)
+        val authorizationCode = httpUrl.queryParameter(PARAM_CODE).orEmpty()
+        val invalidAuthParams = mapOf(
+            AuthParam.AUTHORIZATION_CODE to authorizationCode,
+            AuthParam.REDIRECT_URI to redirectUri
+        ).validate()
```
anthonycr: This original approach to local validation feels sloppy and hard to maintain. I had to add 3 entries to the `AuthParam` enum for this PR and I am not sure it's worth the effort. It complicates the implementation of this class, increasing the burden on the reader and the maintenance effort. 

The simpler approach I took for URIs in `VimeoApiClient` (used for the httpUrl here) is easier to maintain and read.
```
val httpUrl: HttpUrl = HttpUrl.parse(uri) ?: return localVimeoCallAdapter.enqueueEmptyUri(callback)
```
I'm wondering if perhaps we should generify it to take a parameter name like
```
val httpUrl: HttpUrl = HttpUrl.parse(uri) ?: return localVimeoCallAdapter.enqueueInvalidParameter("uri", callback)
```
I'm not really sure how to report the empty authorization code in this case though as it's a parameter in the `uri` parameter. So would we add a secondary message to the error or would we just call out the `authorizationCode` as being missing?

I still want to keep these verifications as simple as possible and rely on the API as much as possible, but in cases like a URI being empty or null, either the API error will be so fatal as to not be helpful or we will be unable to pass the parameter to the API since it could be null and non null is required.

api-core/src/main/java/com/vimeo/networking2/internal/VimeoCall.kt
WhosNickDoglio: I'd vote for making this a `String`. 

All the current instances of it being called immediately convert it to a `String`, so unless you foresee a reason down the line where we'd need the URL as a `HttpUrl` instead of a `String` I see no reason for it to not be a `String`.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
WhosNickDoglio: I would rather go all in then remove them. 👍

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
WhosNickDoglio: I like adding the `authorizeWith` prefix. 

> authenticateWithSsoCodeGrant - unchanged, this should probably change so that it's not clashing with the other code grant function, but i don't have any ideas

Could this be `authenticateWithSso`? I think removing the code grant part would remove some possible confusion or clashing.

anthonycr: I think `authenticateWithSso` is a good idea, thanks.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: Converted `VimeoService` params to all constants as well.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -247,6 +247,51 @@ interface Authenticator {
         callback: VimeoCallback<VimeoAccount>
     ): VimeoRequest
 
+    /**
+     * Find a supported SSO domain that matches the [domain] parameter.
+     *
+     * @param domain A domain, also known as hostname, that might be supported for SSO by the Vimeo API.
+     * @param callback Callback to be notified of the result of the request.
```
howardrigberg, line 256: Optional: "[VimeoCallback] to be notified..."

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -247,6 +247,51 @@ interface Authenticator {
         callback: VimeoCallback<VimeoAccount>
     ): VimeoRequest
 
+    /**
+     * Find a supported SSO domain that matches the [domain] parameter.
```
howardrigberg: Optional: you could connect the dots in the docs that this is a multi step process and provide the order of operations.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -247,6 +247,51 @@ interface Authenticator {
...
+    /**
+     * Obtain the URI which can be used to log in the user and receive back a URI that can be exchanged using
+     * [authenticateWithSso] for a logged in account. Obtain the [SsoDomain] required by this function first using
+     * [fetchSsoDomain].
+     *
+     * @param ssoDomain The domain info obtained from the API that specifies the [SsoDomain.connectUrl] which will be
+     * used as the base for this URI.
+     * @param responseCode The response code that can be used to identify the origin of the redirect URI.
+     *
+     * @return The URI which can be opened in a browser.
```
howardrigberg: "...which needs to be opened in a browser or webview to receive a redirect uri that will be passed to [authenticateWithSso] as the [uri] parameter." 

Am I correct in my understanding? If so, it might help to be a little more explicit about how it needs to be used since it isn't intuitive.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -247,6 +247,51 @@ interface Authenticator {
...
+     */
+    @Internal
+    fun createSsoAuthorizationUri(ssoDomain: SsoDomain, responseCode: String): String
+
+    /**
+     * Authenticate with the server using an authorization code grant from a supported enterprise SSO domain.
+     *
+     * @param uri The URI which was redirected back to you. Must contain a `code` query parameter that contains the
```
howardrigberg, line 290: "...back to you after opening the uri returned from [createSsoAuthorizationUri] in a browser or webview."

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -71,7 +71,7 @@ internal interface VimeoService {
     @GET("me/connected_apps/{type}")
     fun getConnectedApp(
         @Header(AUTHORIZATION) authorization: String,
-        @Path("type") type: ConnectedAppType,
+        @Path(TYPE) type: ConnectedAppType,
         @Query(PARAMETER_GET_FIELD_FILTER) fieldFilter: String?,
```
howardrigberg: It probably makes sense to move this from `ApiConstants` into the local companion object now that you have it. This class is the only place it is being used.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -308,11 +310,89 @@ internal interface AuthService {
...
+     * parameter.
+     *
+     * @param authorization Created from the client id and client secret.
+     * @param domain A domain, also known as hostname, that might be supported for SSO by the Vimeo API.
+     *
+     * @return A [VimeoCall] that provides a [SsoDomain] that can be used to perform SSO.
+     */
+    @Internal
+    @GET("sso_domains")
+    fun getSsoDomain(
```
howardrigberg, line 329: Match naming with `Authenticator` for clarity?

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -308,11 +310,89 @@ internal interface AuthService {
...
+     * Creates a request to the SSO grant authorization endpoint. Used to direct the user to a URL where they can log
+     * in via SSO and then redirect back to the app.
+     *
+     * @param uri The base URI obtained from [SsoDomain.connectUrl].
+     * @param redirectUri The URI which should be redirected back to.
+     * @param state A random value that will be returned in the redirect to identify the origin of the request.
+     */
+    @Internal
+    @GET
+    fun ssoGrantRequest(
```
howardrigberg: Match naming with `Authenticator` for clarity?

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -308,11 +310,89 @@ internal interface AuthService {
...
+     */
+    @Internal
+    @GET("sso_domains")
+    fun getSsoDomain(
+        @Header(AUTHORIZATION) authorization: String,
+        @Query(DOMAIN) domain: String
+    ): VimeoCall<SsoDomain>
+
+    /**
+     * Creates a request to the SSO grant authorization endpoint. Used to direct the user to a URL where they can log
```
howardrigberg, line 335: This description is a little clearer than the one in Authenticator.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -230,24 +230,77 @@ internal class AuthenticatorImpl(
...
+                invalidParameters = invalidAuthParams
+            )
+            call.enqueueError(apiError, callback)
+        } else {
+            call.enqueue(callback)
+        }
+    }
+
+    override fun createSsoAuthorizationUri(ssoDomain: SsoDomain, responseCode: String): String {
+        val connectUrl = requireNotNull(ssoDomain.connectUrl) { "SsoDomain.connectUrl cannot be null" }
```
howardrigberg: In `fetchSsoDomain` we enqueue an error when a string is empty, but here we throw an exception if it is null and don't check if it is empty. Ideally we would consistently deal with required parameters unless I am missing something.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
howardrigberg: I agree that it is easier to understand and maintain when the approach addresses the string parameters directly as you have done rather than indirectly through the parameter map. Perhaps we could take a hybrid approach to what you came up with and the existing `validate` extension?

It might be something like this:

```
    data class ApiParameter(
        val fieldName: String,
        val fieldValue: String?,
        val errorCodeType: ErrorCodeType = ErrorCodeType.DEFAULT
    )

    fun <T> LocalVimeoCallAdapter.validateParameters(
        callback: VimeoCallback<T>,
        developerMessage: String?,
        vararg params: ApiParameter
    ): VimeoRequest? {
        val invalidParameters = mutableListOf<InvalidParameter>()
        for (param in params) {
            if (param.fieldValue.isNullOrBlank()) {
                invalidParameters.add(InvalidParameter(
                    field = param.fieldName,
                    errorCode = param.errorCodeType.value,
                    developerMessage = "${param.fieldName} is empty or null"
                ))
            }
        }
        return if (invalidParameters.isNotEmpty()) {
            enqueueError(ApiError(
                invalidParameters = invalidParameters,
                developerMessage = developerMessage
            ), callback)
        } else {
            null
        }
    }
```
Calling it in the context of your method would look like this:
```
localVimeoCallAdapter.validateParameters(
            callback,
            "SSO code grant error",
            ApiParameter("uri", httpUrl?.toString(), ErrorCodeType.INVALID_URI),
            ApiParameter("authorization_code", authorizationCode),
            ApiParameter("redirect_uri", redirectUri)
        )?.let { errorResponse ->
            return errorResponse
        }
```

Taking it a step further, I am still in favor of having a consistent client-side approach to handling any required string parameters to our api methods. I think we should always check that they are not empty and enqueue an ApiError for empty parameters. We have gone back and forth on this in the past, but if we have an easy way to validate string parameters client-side, I think the cost benefit of always doing it will be worth it.

howardrigberg (changes requested):
Cool to see the SSO is relatively straightforward. I've got a few suggestions.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: For this comment and the next one on documentation, I opted to copy the docs from the `authenticateWithCodeGrant` instead of just changing the return doc. We worked a lot on the code grant docs in a previous PR, and SSO follows the same flow, so we can just reuse them on these functions and slightly adapt them for SSO.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: See other documentation comment, this URI doc is identical to the code grant param doc, but I updated the actual function doc to contain the information you mentioned.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: Updated the authenticator one to match that of the code grant that is clearer.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: The naming convention we have been using for get requests in the library has been `get[THING]` in the service layer and then `fetch[THING]` in the client layer (see `VimeoService` and `VimeoApiClient`). The reason for the difference is to emphasize that the functions are not doing the same thing, as one creates a request and another actually executes the request. The service functions have loosely matched the HTTP method annotation, there are a number of exceptions in this service. I don't think we should rename this function without renaming all the `VimeoService` functions.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: The authenticator function does something different than this one. The authenticator function extracts the URL from the request, while this one creates the request. I renamed this `createSsoGrantRequest` and the other one `createCodeGrantRequest` to be more clear, but changing the name to match the authenticator exactly would be semantically incorrect.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
anthonycr: This function just creates a URL, it doesn't actually make a request, so there is no callback to notify of an error. Our only option is to throw an exception or return null. Since crashing is a little harsh and there's no way to guarantee that the API won't return a null URL, I made the return type null and added docs that state that null will be returned if the parameter is invalid.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: This description is already consistent with the description used elsewhere in the `Authenticator`, so I'm going to leave it.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
anthonycr: The issue I see with a hybrid approach like this is that it cannot guarantee non nullity. So for example, in `VimeoApiClient.editAlbum`, currently we have 
```kotlin
val uri = album.uri.notEmpty() ?: return localVimeoCallAdapter.enqueueEmptyUri(callback)
```
which verifies that the URI is neither empty nor null. If we were to go with this hybrid approach or the current approach used in the `Authenticator`, then we would need an additional `album.uri!!` or `requireNotNull(album.uri)` when passing it to the service in order to satisfy the type system. Needing to use a lot of these assertions usually is an indication that there is an issue with the design. `VimeoApiClient` would need 16 of these assertions, out of a total of 68 error checks. Another thought, if we wanted to move away from throwing an exception if a user provides an `UNKNOWN` enum value (see `StringValue`) to a function that supplies enums, we will also need to support some sort of null assertion.

Do you have any thoughts on how to proceed based on that @howardrigberg?

anthonycr:
@howardrigberg I made addressed all of the suggestions, except the error handling one, which needs some further discussion. Please see the comment I made and let me know what you think we should do, or if we should address this at another time.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
howardrigberg: Good points. I think your concern about the non-null guarantee on a `String` could be addressed in the design of the `ApiParameter` class as follows:

```
data class ApiParameter(
    val fieldName: String,
    val fieldValue: String?,
    val errorCodeType: ErrorCodeType = ErrorCodeType.DEFAULT,
    var validatedFieldValue: String = ""
) 
```
It is a little odd at first glance, but when used in the context of the validation check from `validateParameters`, you'd be guaranteed to get what you want on the other end if you bail out whenever you get a non-null `VimeoRequest` back from `validateParameters`. The `validateParameters` method would set `isValidated` for each `ApiParameter` as it is checked.
```
fun <T> LocalVimeoCallAdapter.validateParameters(
        callback: VimeoCallback<T>,
        developerMessage: String?,
        vararg params: ApiParameter
    ): VimeoRequest? {
        val invalidParameters = mutableListOf<InvalidParameter>()
        for (param in params) {
            if (param.fieldValue.isNullOrBlank()) {
                invalidParameters.add(InvalidParameter(
                    field = param.fieldName,
                    errorCode = param.errorCodeType.value,
                    developerMessage = "${param.fieldName} is empty or null"
                ))
            } else {
               param.validatedFieldValue = param.fieldValue
           }
        }
        return if (invalidParameters.isNotEmpty()) {
            enqueueError(ApiError(
                invalidParameters = invalidParameters,
                developerMessage = developerMessage
            ), callback)
        } else {
            null
        }
    }
```

The revised way of interacting would be a little more cumbersome because you'd be keeping references around to the ApiParameters rather than creating them inline to pass into `validateParameters`:
```
val uriParam = ApiParameter("uri", httpUrl?.toString(), ErrorCodeType.INVALID_URI)
val authCodeParam = ApiParameter("authorization_code", authorizationCode)
val redirectUriParam = ApiParameter("redirect_uri", redirectUri)
localVimeoCallAdapter.validateParameters(callback, "SSO code grant error", uriParam, authCodeParam, redirectUriParam)?.let { return it }

iExpectANonNullStringParam(uriParam.validatedFieldValue)
```

anthonycr:
@howardrigberg this PR is now ready for your re-review.

api-core/src/main/java/com/vimeo/networking2/internal/LocalVimeoCallAdapter.kt
```diff
@@ -45,3 +48,39 @@ class LocalVimeoCallAdapter(private val callbackExecutor: Executor) {
...
 }
+
+/**
+ * Validate the parameters provided and enqueue an error if any of the parameters are invalid.
+ *
+ * @param callback The callback to enqueue if there are invalid parameters.
+ * @param developerMessage The developer message that describes the request that failed.
+ * @param parameters The parameters that will be validated. Upon validation, these parameters will be updated with the
+ * validated parameter value.
+ */
```
howardrigberg: It would be helpful to describe the `@return` in this case.

api-core/src/main/java/com/vimeo/networking2/config/VimeoApiConfiguration.kt
```diff
@@ -246,7 +246,7 @@ data class VimeoApiConfiguration(
             baseUrl = baseUrl,
             clientId = clientId,
             clientSecret = clientSecret,
-            scope = Scopes(scopes),
+            scope = Scopes(scopes.also { require(it.isNotEmpty()) { "You must specify at least one scope" } }),
```
howardrigberg, line 249: ❤️

api-core/src/main/java/com/vimeo/networking2/internal/params/ApiParameter.kt
```diff
@@ -0,0 +1,41 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.vimeo.networking2.internal.params
+
+/**
+ * A class representing an API parameter that can be validated.
```
howardrigberg: This class is designed to be used in association with LocalVimeoCallAdapter.validateParameters.

api-core/src/main/java/com/vimeo/networking2/internal/params/ApiParameter.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param parameterName The name of the field being validated.
+ * @param parameterValue The original pre-validation value being provided.
+ * @param developerMessage The optional message to provide if the parameter is invalid.
```
howardrigberg: "... is invalid. A default developer message will be added by LocalVimeoCallAdapter.validateParameters when the parameterValue is invalid if none is provided here.

api-core/src/main/java/com/vimeo/networking2/internal/params/ApiParameter.kt
```diff
@@ -0,0 +1,41 @@
...
+data class ApiParameter(
+    val parameterName: String,
+    val parameterValue: String?,
+    val developerMessage: String? = null
+) {
+    /**
+     * The validated version of the [parameterValue].
+     */
+    var validatedParameterValue: String = ""
+        internal set
```
howardrigberg, line 48: ooooh. nice move.

api-core/src/main/java/com/vimeo/networking2/internal/LocalVimeoCallAdapter.kt
```diff
@@ -45,3 +48,39 @@ class LocalVimeoCallAdapter(private val callbackExecutor: Executor) {
...
+fun <T> LocalVimeoCallAdapter.validateParameters(
```
howardrigberg: Optional: could write unit tests for this.

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
```diff
@@ -327,9 +407,25 @@ internal interface AuthService {
...
+        const val TOKEN = "token"
+        const val TOKEN_SECRET = "token_secret"
+        const val DOMAIN = "domain"
+        const val USER_CODE = "user_code"
+        const val DEVICE_CODE = "device_code"
+        const val ACCESS_TOKEN = "access_token"
+        const val CODE = "code"
+        const val CLIENT_ID = "client_id"
+        const val ID_TOKEN = "id_token"
+        const val USERNAME = "username"
```
howardrigberg, line 429: It is a bit confusing that this is called "username" but is always used in the code to pass an email. It would be clearer in the consumers if you renamed this as `EMAIL`. A line comment could be put here to explain the reason for the discrepancy. Unless "username" should actually be "email"?

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -50,204 +68,195 @@ internal class AuthenticatorImpl(
...
+            "OAuth1 token exchange error",
+            tokenParameter,
+            tokenSecretParameter
+        ) ?: authService.exchangeOAuth1Token(
             authorization = basicAuthHeader,
             grantType = GrantType.OAUTH_ONE,
-            token = token,
-            tokenSecret = tokenSecret,
+            token = tokenParameter.validatedParameterValue,
+            tokenSecret = tokenParameter.validatedParameterValue,
```
howardrigberg: `tokenSecretParameter.validatedParameterValue`

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -50,204 +68,195 @@ internal class AuthenticatorImpl(
...
+        )
+        val redirectUri = ApiParameter(REDIRECT_URI, redirectUri)
+
+        return localVimeoCallAdapter.validateParameters(
+            callback,
+            "SSO authentication error",
+            authorizationCode,
+            redirectUri
+        ) ?: authService.joinWithSsoCodeGrant(
+            basicAuthHeader,
```
howardrigberg: use named parameters

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -50,204 +68,195 @@ internal class AuthenticatorImpl(
...
+        )
+        val redirectUri = ApiParameter(REDIRECT_URI, redirectUri)
+
+        return localVimeoCallAdapter.validateParameters(
+            callback,
+            "Code grant authentication error",
             authorizationCode,
+            redirectUri
+        ) ?: authService.authenticateWithCodeGrant(
+            basicAuthHeader,
```
howardrigberg: use named parameters

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -262,7 +271,7 @@ internal class AuthenticatorImpl(
         val deviceCode = pinCodeInfo.deviceCode.orEmpty()
```
howardrigberg: No validation on these?

howardrigberg (changes requested):
Code looks much better now! Just a few more things to resolve...

auth/src/main/java/com/vimeo/networking2/internal/AuthService.kt
anthonycr: In practice, we only ever pass an email to these functions, and historically we have called the parameters "email," but since the API calls it "username" and accepts some weird edge cases other than emails, I'll just rename all the parameters annotated by this to "username."

api-core/src/main/java/com/vimeo/networking2/internal/LocalVimeoCallAdapter.kt
anthonycr: We have a ticket to write tests for the whole library, can defer to later.

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
anthonycr: Good catch.

anthonycr:
@howardrigberg good catches, thanks for the review. I think I got everything.

howardrigberg (approved):

PR #451, opened 2020-09-30
Add function to allow a folder to be fetched

anthonycr:
# Summary
I added a function to allow an individual folder to be fetched from the API.


howardrigberg (approved):
LGTM

PR #457, opened 2020-11-16
Update dependencies

anthonycr:
# Summary
A number of dependencies were out of date
- Kotlin: 1.3.72 -> 1.4.10
- AGP: 3.6.3 -> 4.1.1
- Detekt: 1.0.0-RC14 -> 1.14.2
- appcompat: 1.1.0 -> 1.2.0
- constraintlayout: 1.1.3 -> 2.0.4
- gradle: 5.6.4 -> 6.5
- Android target version: 29 -> 30
- Android build tools: ? -> 29.0.3

As a result of upgrading detekt, I fixed a few lint warnings and suppressed the rest. One thing we will have to fix in the future is the documentation on DTOs in the `models` module. Currently we are documenting them like this
```kotlin
data class Dto(
  /**
   * Documentation.
   */
  val property: String
)
```
However, this is technically incorrect as only the property is documented instead of both the constructor parameter AND property being documented. Apparently, the correct way to do it is to use parameter documentation which would then be automatically inherited by the property, as follows:
```kotlin
/**
 * @param property Documentation.
 */
data class Dto(
  val property: String
)
```
I saved this for another time and just added them to the baseline for now as it will require a lot of changes and may not be what we want to do in the end anyway.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Mon Nov 16 13:48:05 EST 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-all.zip
```
WhosNickDoglio: Unless there's something blocking this I'd like to jump to the latest with [6.7.1](https://github.com/gradle/gradle/releases/tag/v6.7.1)

example/build.gradle
```diff
@@ -26,13 +26,11 @@ android {
 }
 
 dependencies {
-    implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':auth')
     implementation project(':request')
     implementation project(':models')
 
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
WhosNickDoglio: Optional: As part of Kotlin 1.4 I'm pretty sure the standard lib is automatically implied if you apply any of the kotlin plugins, we could remove it as a explicit dependency from all the`build.gradle`

build.gradle
```diff
@@ -1,31 +1,31 @@
 buildscript {
-    ext.kotlin_version = '1.3.72'
+    ext.kotlin_version = '1.4.10'
```
WhosNickDoglio: Doesn't need to be part of this PR but since we're moving this project to 1.4 I think turning on [Explicit API Mode](https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors) would be a good idea 👍

WhosNickDoglio: Optional: [1.4.20 dropped recently](https://github.com/JetBrains/kotlin/releases/tag/v1.4.20) if we also want to make that upgrade in this PR

build.gradle
```diff
@@ -1,31 +1,31 @@
...
 }
 
 ext {
-    kotlin_version = '1.3.72'
     retrofitVersion = '2.9.0'
     okioVersion =  '2.7.0'
     moshiVersion = '1.9.3'
     stagVersion = '2.6.0'
     junitVersion = '4.13'
+    buildToolsVersion = '29.0.3'
```
WhosNickDoglio, line 25: Optional: The [Android plugin ](https://developer.android.com/studio/build#module-level) takes care of this by default so this can be omitted unless it's important to deal with the issues with Travis

build.gradle
```diff
@@ -1,31 +1,31 @@
...
 plugins {
-    id "io.gitlab.arturbosch.detekt" version "1.0.0-RC14"
+    id "io.gitlab.arturbosch.detekt" version "1.14.2"
 }
 
 ext {
-    kotlin_version = '1.3.72'
     retrofitVersion = '2.9.0'
     okioVersion =  '2.7.0'
     moshiVersion = '1.9.3'
```
WhosNickDoglio: Optional: Moshi can also be bumped to [1.11.0](https://github.com/square/moshi/releases/tag/moshi-parent-1.11.0)

WhosNickDoglio (approved):
Overall looks good

Left some optional suggestions and also opened up a PR to switch to Github actions (#460)  to avoid issues with Travis and android licenses 👍

build.gradle
anthonycr: Yeah perhaps we can do it in the future - it requires a lot of changes to be made to the code as we need to add visibility modifiers to everything.

example/build.gradle
anthonycr: Builds fine without it so I have removed it as you suggested.

PR #461, opened 2020-11-24
Add capability to clear the cache for a specific URI to the library

anthonycr:
# Summary
We want to be able to invalidate the cache on a granular level. This means we want to be able to invalidate the cache on a per URI bases. This PR adds that capability to the `RetrofitSetupModule`.

It adds the `clearRequestCacheForUri` function which takes the `VimeoApiConfiguration` instance used to initialize the library and the `uri` for which the library will be cleared. For example, providing `/user/12345` as the URI will clear the cache for `https://api.vimeo.com/user/12345`.

### DIFF
--- a/api-core/src/main/java/com/vimeo/networking2/config/RetrofitSetupModule.kt
+++ b/api-core/src/main/java/com/vimeo/networking2/config/RetrofitSetupModule.kt
@@ -83,6 +83,21 @@ object RetrofitSetupModule {
     }
 
     /**
+     * Clear the request cache entry associated with the [VimeoApiConfiguration] and the provided [uri].
+     */
+    @JvmStatic
+    fun clearRequestCacheForUri(vimeoApiConfiguration: VimeoApiConfiguration, uri: String) {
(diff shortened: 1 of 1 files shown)


api-core/src/main/java/com/vimeo/networking2/config/RetrofitSetupModule.kt
```diff
@@ -82,6 +83,32 @@ object RetrofitSetupModule {
         vimeoApiConfiguration.cache?.evictAll()
     }
 
+    /**
+     * Clear the request cache entry associated with the [VimeoApiConfiguration] and the provided [uri].
+     *
+     * For example, if the [VimeoApiConfiguration.baseUrl] is set to `https://api.vimeo.com`, and the [uri]
+     * provided is `/users/12345`, then the cache will be cleared for `https://api.vimeo.com/users/12345`.
```
howardrigberg: I realize that it isn't a pattern that is currently followed in this file, but shouldn't we have `@param` documentation here?

howardrigberg (approved):

PR #462, opened 2020-12-02
Fix bug where videos over ~2GB could not be uploaded

anthonycr:
# Summary
There is an issue with the `Gcs` DTO where the `endBytes` type was `Int` instead of `Long`. As a result, only videos up to `Int.MAX` size could be uploaded. The fix is to change the type of the property to `Long`. I also changed the start property to `Long` as well for consistency and in case an upload offset was used greater than `Int.MAX` bytes.


howardrigberg (approved):
Great find!

PR #464, opened 2021-01-21
Folders CRUD support

anthonycr:
# Summary

Supporting CRUD for folders. Adds the ability to create folders, edit folder settings, add videos to folders, removing videos from folders, and deleting folders. Additionally, I added the ability to get a list of folders, which required creating a `FolderList` DTO. The functions added are
- `createFolder(uri)`
- `createFolder(user)`
- `editFolder(uri)`
- `editFolder(folder)`
- `addToFolder(folder, video)`
- `addToFolder(folderUri, videoUri)`
- `removeFromFolder(folder, video)`
- `removeFromFolder(folderUri, videoUri)`
- `fetchFolderList(uri)`
- `deleteFolder(folder)` - no URI equivalent exists for this since `deleteContent(uri)` covers that use case.

I removed the following function which was duplicated by `deleteContent(uri)`
- `deleteAlbum(uri)`

I also added a new connection to `UserConnections` which returns all the folders of the user. I called it `folders` since it returns all the user's folders. Since we already had a `folders` connection which returned the projects in the root directory, I renamed that to `projectItemsRoot`. I think this is more clear than before, where the `folders` connection returned `ProjectItemList` instead of `FolderList`. Now `folders` returns a `FolderList` and the `projectItemsRoot` returns a `ProjectItemList`.

Additionally, I added the `parentFolder` property to a `Video` which allows us to see what folder (if any) that a video is currently in.

## Other
I fixed a bug where the consumer could try to clear empty URIs which results in a crash.

## Testing
1. Log in with `Authenticator.authenticateWithEmailLogin`
2. Create a folder by passing `Authenticator.currentAccount?.user` to `VimeoApiClient.createFolder(user, name, privacy)`
3. Fetch the current folders by passing `Authenticator.currentAccount?.user?.metadata?.connections?.folders?.uri` to `VimeoApiClient.fetchFolderList(uri)` to verify that the folder was created.
4. Add a video to the folder by passing the previously created folder and a video uploaded by the user to `VimeoApiClient.addToFolder(folder, video)`.
5. Remove the video from the folder by passing the previously created folder and a video uploaded by the user to `VimeoApiClient.removeFromFolder(folder, video)`.
6. Delete the folder by passing the folder previously created to `VimeoApiClient.deleteFolder(folder)`.


models/src/main/java/com/vimeo/networking2/FolderList.kt
```diff
@@ -0,0 +1,28 @@
+package com.vimeo.networking2
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import com.vimeo.networking2.common.Pageable
+
+/**
+ * List of [Folder]s to show in a user's feed.
```
howardrigberg: [Folders][Folder]

models/src/main/java/com/vimeo/networking2/UserConnections.kt
```diff
@@ -54,10 +54,10 @@ data class UserConnections(
     val feed: BasicConnection? = null,
 
     /**
-     * Information about this user's folders.
+     * Information about the root folder containing the user's projects (folders and videos).
```
howardrigberg: folder -> location?

Maybe this will be a little clearer?

models/src/main/java/com/vimeo/networking2/Video.kt
```diff
@@ -120,6 +120,12 @@ data class Video(
     @Json(name = "name")
     val name: String? = null,
 
+    /**
+     * Information about the folder that contains the video, or null if it is in the root folder.
```
howardrigberg: folder -> location?

Same comment as I made in `UserConnections`. Optional. Up to you. Just change if you think it is clearer.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -422,6 +410,147 @@ interface VimeoApiClient {
      */
     fun deleteConnectedApp(type: ConnectedAppType, callback: VimeoCallback<Unit>): VimeoRequest
 
+    /**
+     * Create a folder that will be used to organize videos.
+     *
+     * @param uri The URI of the user's projects connection.
```
howardrigberg: folders connection?

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -422,6 +410,147 @@ interface VimeoApiClient {
...
+        user: User,
+        name: String,
+        privacy: FolderViewPrivacyType,
+        callback: VimeoCallback<Folder>
+    ): VimeoRequest
+
+    /**
+     * Delete a folder.
+     *
+     * @param folder The folder being deleted, URI should not be null or empty.
```
howardrigberg: "URI should not be null or empty"... You mean `folder.uri`, right? It looks like you got this from the `deleteAlbum` method which arguably could be a little clearer too. In hindsight, I'm not sure if the warning is necessary because the actual implementation will just enqueue an empty api error in the event that the URI is null or empty, not crash or anything. The warning gives the impression that the caller should actively check the member uri before calling which is not necessary.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -422,6 +410,147 @@ interface VimeoApiClient {
...
+     * @param privacy The privacy of the folder.
+     * @param callback The callback which will be notified of the request completion.
+     *
+     * @return A [VimeoRequest] object to cancel API requests.
+     */
+    fun editFolder(
+        uri: String,
+        name: String,
+        privacy: FolderViewPrivacyType,
+        callback: VimeoCallback<Folder>
```
howardrigberg, line 494: The api also supports slack related parameters here. Maybe we should add them in now so we are ready to support this later? Web currently exposes it.

https://github.vimeows.com/Vimeo-API/design/blob/master/docs/3.4.2/apiblueprint/resources/Folders/Essentials.apib

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -422,6 +410,147 @@ interface VimeoApiClient {
...
+     * Create a folder that will be used to organize videos.
+     *
+     * @param user The user whose projects connection will be used for the request.
+     * @param name The name of the folder.
+     * @param privacy The privacy of the folder.
+     * @param callback The callback which will be notified of the request completion.
+     *
+     * @return A [VimeoRequest] object to cancel API requests.
+     */
+    fun createFolder(
```
howardrigberg, line 451: The api also supports slack related parameters and a parent web folder id here. Maybe we should add them in now so we are ready to support this later? 

https://github.vimeows.com/Vimeo-API/design/blob/master/docs/3.4.2/apiblueprint/resources/Folders/Essentials.apib

howardrigberg (changes requested):
Looks good. I've got some questions about missing parameters as well as some minor comments on documentation.

Not sure if this matters now or if it is deliberate, but it looks like there are properties of a folder that are not yet represented (e.g. link, slack_user_preferences, etc.). 

See:
https://github.vimeows.com/Vimeo-API/design/blob/master/docs/3.4.2/apiblueprint/representations/Project.apib

models/src/main/java/com/vimeo/networking2/Video.kt
anthonycr: How does "directory" sound instead of "location?" I used "root folder" in both of these docs because it's a common phrase used in file systems. I can understand that "root folder" is a little confusing as the root folder isn't actually a `Folder` in our case, but "root location" isn't really a familiar phrase used in file systems in my experience. I think "root directory" strikes a balance between familiarity and avoiding the confusion of "root folder" not being an actual `Folder`.

howardrigberg: Yeah, I like that. 👍

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: `URI should not be null or empty` is the standard warning we apply to functions in the client that accept a DTO instead of a URI. I agree it seems unnecessary, since we gracefully respond with a failure if the URI is null. I'll remove it from all the functions.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: I added the slack parameters, but decided against adding the parent folder ID since that's not an available feature yet. We can add it later when nested folders is functional and the API is fully fleshed out for it.

anthonycr:
@howardrigberg thanks for the feedback, needs another review.

models/src/main/java/com/vimeo/networking2/enums/SlackLanguagePreference.kt
```diff
@@ -0,0 +1,42 @@
+package com.vimeo.networking2.enums
+
+/**
+ * Language preference of the Slack channel being notified.
+ */
+enum class SlackLanguagePreference(override val value: String?) : StringValue {
```
howardrigberg: Add this to `EnumsTest`

models/src/main/java/com/vimeo/networking2/enums/SlackUserPreference.kt
```diff
@@ -0,0 +1,27 @@
+package com.vimeo.networking2.enums
+
+/**
+ * The preference for which changes a user wants the Slack channel to be notified.
+ */
+enum class SlackUserPreference(override val value: String?) : StringValue {
```
howardrigberg: Add this to `EnumsTest`

howardrigberg (changes requested):
Looks good. Just need to add the new enums to the tests.

models/src/main/java/com/vimeo/networking2/enums/SlackUserPreference.kt
anthonycr: Good call out. I looked over the API docs and realized our `Folder` implementation was missing the slack related properties, so I added them.

anthonycr:
@howardrigberg this PR is ready for another look.

howardrigberg (approved):
LGTM

PR #466, opened 2021-01-25
MPS-1617 Modifying createAlbum to support teams

howardrigberg:
#### Issue
In order to support album creation for teams, `createAlbum` needs to accept a uri parameter instead of setting it automatically from me/albums.

- https://github.com/vimeo//vimeo-networking-java/issues/MPS-1617

#### Summary
Added a uri parameter to `createAlbum`.

#### How to Test
N/A


request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -82,6 +83,7 @@ interface VimeoApiClient {
      * @return A [VimeoRequest] object to cancel API requests.
      */
     fun createAlbum(
+        uri: String,
```
anthonycr, line 89: Now that this function takes a URI, we should create a counterpart that accepts a `User` as well. This will make it easier for consumers to create an album, without needing to select the correct connection.

anthonycr (changes requested):

howardrigberg:
@anthonycr Good suggestion. Back to you.

anthonycr (approved):
Looks good!

PR #467, opened 2021-01-25
Avoid allowing empty URI in cache clear function

anthonycr:
# Summary
Trying to call the `clearCacheForUri` function with an empty `String` will result in a crash. This PR checks whether the URI is blank before trying to create the full path that will be cleared. While a blank URI is invalid and the previous behavior of crashing was technically correct, it's easy to accidentally pass an empty `String` here, so it's a quick safeguard to put in place to handle cases where the consumer probably didn't intentionally try to crash the code.


WhosNickDoglio (approved):
LGTM

PR #469, opened 2021-01-28
Add missing property to the Folder DTO

anthonycr:
# Summary
The `link` property, which provides the user shareable URL for a `Folder`, was missing from the `Folder` DTO. I believe it was added after we initially created it last year. This PR adds it. I also verified that were were not missing any other properties from `Folder`.


howardrigberg (approved):
Good catch!

PR #470, opened 2021-01-29
Add messaging to the empty URI error to make it more clear what's wrong

anthonycr:
# Summary
It's not very clear what's wrong when you provide an empty URI when making a request. While the error code is provided, it is easy to overlook. It is also easy to provide an empty URI without meaning to by using one of the `VimeoApiClient` functions that accepts a DTO instead of a URI. Adding a developer message to the error makes it more clear what is wrong.


WhosNickDoglio (approved):
LGTM 👍

PR #475, opened 2021-03-15
MPS-1889: Manage Teams support

WhosNickDoglio:
#### Issue

[MPS-1889](https://vimean.atlassian.net/browse/MPS-1889)

#### Summary
Added the endpoints needed for the manage team flow as well as a few new models

Also bumped Ben Manes Versions gradle plugin since Jitpack was having trouble finding it 

#### How to Test
There's a test branch I created in the core app that can be used for testing each new endpoint, see internal PR for bumping the network library.


request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -1551,6 +1536,163 @@ interface VimeoApiClient {
...
+     * @param cacheControl The optional cache behavior for the request, null indicates that the default cache behavior
+     * should be used.
+     * @param callback The callback which will be notified of the request completion.
+     *
+     * @return A [VimeoRequest] object to cancel API requests.
+     */
+    fun fetchTeamMembersList(
+        uri: String,
+        fieldFilter: String?,
+        queryParams: Map<String, @JvmSuppressWildcards String>,
```
anthonycr: Should be nullable since it's nullable. Same applies to other `queryParams` added in this PR.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -1551,6 +1536,163 @@ interface VimeoApiClient {
...
+     * @param uri the URI from which content will be sent to.
+     * @param body an [AddUserToTeam] object which contains all the details for the given request.
+     * @param queryParams Optional map used to refine the response from the API.
+     * @param callback The callback which will be notified of the request completion.
+     *
+     * @return A [VimeoRequest] object to cancel API requests.
+     */
+    fun addUserToTeam(
+        uri: String,
+        body: AddUserToTeam,
```
anthonycr: For simple one level bodies without any lists, we have been mostly utilizing `@Field` parameters instead of creating a body parameter that we transparently pass along. In my opinion the extra creation required here hurts the usability of the function, as it splits the documentation for making this request between this function and the `AddUserToTeam`. It also adds an extra level of nesting in the simple use case. In this specific case, it would allow us to make the other variant of `addUserToTeam` accept a `User` instead of a URI for improved usability. 

If there's an argument against using fields in favor of creating bodies, let's discuss.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -1551,6 +1536,163 @@ interface VimeoApiClient {
...
+     * to the given folder, if a user who currently has access to this folder is not present in this list they will have
+     * their folder permission revoked.
+     * @param queryParams Optional map used to refine the response from the API.
+     * @param callback The callback which will be notified of the request completion.
+     *
+     * @return A [VimeoRequest] object to cancel API requests.
+     */
+    fun grantUsersAccessToFolder(
+        uri: String,
+        usersIds: List<GrantFolderPermissionForUser>,
```
anthonycr: Do you have any concerns with making this `userIds: List<String>` and the other function `users: List<User>`? It would require additional logic inside the client, but I think it would improve usability a little. Another similar function we have here, `modifyVideoInAlbums(...)` (see its usage in codebase), has very verbose usage due to converting from the DTO used (`Album` in that case) to the actual parameter type.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -1551,6 +1536,163 @@ interface VimeoApiClient {
...
+        queryParams: Map<String, @JvmSuppressWildcards String>,
+        callback: VimeoCallback<Unit>
+    ): VimeoRequest
+
+    /**
+     * Grants permission for the given users to access the given folder.
+     *
+     * @param folder The [Folder] that [usersIds] will be granted access to.
+     * @param usersIds A list of [GrantFolderPermissionForUser] containing URIs for users that will have access
+     * to the given folder, if a user who currently has access to this folder is not present in this list they will have
```
anthonycr, line 1692: > if a user who currently has access to this folder is not present

Seems that this will have the unfortunate implication that we would need to recurse the entire list of users added to a folder before making such a request. Am I correct? I think this will not be an issue in the average use case as the maximum team member count is 10 for business accounts, a number which would not require us to make multiple requests, but for enterprise accounts that would be a custom limit. Did you have any thoughts on this yet?

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
```diff
@@ -561,6 +544,112 @@ internal class VimeoApiClientImpl(
...
+        callback: VimeoCallback<TeamMembership>
+    ): VimeoRequest {
+        val safeUri = team.owner?.metadata?.connections?.teamMembers?.uri.notEmpty()
+            ?: return localVimeoCallAdapter.enqueueEmptyUri(callback)
+        return vimeoService.addUserToTeam(authHeader, safeUri, body, queryParams).enqueue(callback)
+    }
+
+    override fun removeUserFromTeam(
+        uri: String,
+        queryParams: Map<String, @JvmSuppressWildcards String>,
```
anthonycr: Is this suppression necessary? I don't see suppression of any functions outside of these functions added in this PR. If it's not necessary I would suggest removing it for consistency with other functions in the library.

Same applies to suppressions added in the `MutableVimeoApiClientDelegate` and the `VimeoApiClient`.

models/src/main/java/com/vimeo/networking2/TeamMembershipList.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import com.vimeo.networking2.common.Pageable
+
+/**
+ * List of [TeamMemberships][TeamMembership] that can be paged.
+ */
+@JsonClass(generateAdapter = true)
+data class TeamMembershipList(
```
anthonycr, line 11: Should be added to `ModelsTest`.

anthonycr (changes requested):
I have a some requests and a question.

request/src/main/java/com/vimeo/networking2/internal/VimeoApiClientImpl.kt
WhosNickDoglio: I'm not sure why I added this 🤔

I could of sworn I saw it elsewhere but I guess I misread something. I'll remove them 👍

WhosNickDoglio: Ah I got `@JvmSuppressWildcards` from `VimeoService` since I worked backwards from there.

anthonycr: Ah that makes sense. Yeah suppression is necessary in the `VimeoService` layer in some places due to Retrofit not really knowing how to handle wildcards.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
WhosNickDoglio: Just to make sure I understand, you're suggesting the following 

```
    fun grantUsersAccessToFolder(
        uri: String,
        usersIds: List<String>,
        queryParams: Map<String, String>?,
        callback: VimeoCallback<Unit>
    ): VimeoRequest

    fun grantUsersAccessToFolder(
        folder: Folder,
        users: List<User>,
        queryParams: Map<String, String>?,
        callback: VimeoCallback<Unit>
    ): VimeoRequest
```
And internally we'd just map the URIs to the `GrantFolderPermissionForUser` model?

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
WhosNickDoglio: This is a good callout.  

My initial thoughts are since the Edit/Create Folder screen is going to load all the team members anyway I can just filter that for users with access to this folder _but_ like you said for custom limits this might not work.  We might have to load the entire list of users recursively like suggested, I am curious though what the limit per response is since we had concerns about having to make recursive calls for the Analytics APIs but I was able to over 300+ data points in responses so I ended up not needing that. I'll have to look how iOS is currently implementing but regardless this will be a use case I'll have to keep in mind when I start working on this screen.

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: Yes!

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
anthonycr: I know there is a limit of 100 items per page in some areas of the API. I'm not sure if that's a widespread limit that doesn't affect stats (since it's a different DB) or if it's an older limit that is no longer used.

WhosNickDoglio: Ah good point I forgot stats are separate

WhosNickDoglio:
@anthonycr I _think_ I got everything, back to you 👍

request/src/main/java/com/vimeo/networking2/VimeoApiClient.kt
```diff
@@ -1551,6 +1536,173 @@ interface VimeoApiClient {
...
+     * @param queryParams Optional map used to refine the response from the API.
+     * @param callback The callback which will be notified of the request completion.
+     *
+     * @return A [VimeoRequest] object to cancel API requests.
+     */
+    fun addUserToTeam(
+        team: Team,
+        email: String,
+        permissionLevel: TeamRoleType,
+        folderUri: String?,
```
anthonycr: Can we make this one `folder: Folder`?

anthonycr (approved):
Looks good, I have one minor request. Feel free to merge after addressing.

PR #476, opened 2021-03-22
Add missing team members connection attributes

anthonycr:
# Summary
We were missing a couple attributes on the `teamMembers` connection in `UserConnections`. Since this connection was previously a `BasicConnection`, I created a new `TeamMembersConnection`. 

Additionally, we have been violating detekt's rules for how we write documentation for our public properties. Previously, we only put documentation on the property itself. However, a recent update to detekt added the requirement that public properties that are also constructor parameters also need parameter docs. The requirements are that we need param docs for public, non-overridden, `val` and `var` properties. Param docs are also reflected by the property docs, so we really just need to move the documentation from the properties to the params. Since this is the new default, and it's not configurable, we should follow these rules rather than try to fight them. It simplifies our documentation by consuming fewer lines of code, and allows the parameters to have documentation in addition to the properties themselves. This improves copying and construction of these classes.

Going forward, if a class is changed, rather than running `detektBaseline` again to avoid changing the docs, we should just move the docs to params. As such, I changed
- `BasicConnection`
- `UserConnection`

So that they use param docs instead of just property docs. `TeamMembersConnection` also uses the correct documentation style.


elhicks (approved):
LGTM

PR #477, opened 2021-03-24
Make connection a data class

anthonycr:
# Summary
`TeamMembsConnection` should have been a `data` class, so I added it. Additionally, I discovered that there is a bug with the `resource_key` property and as a result we have to use the `uri` property as our identifier instead.


elhicks (approved):
LGTM

PR #478, opened 2021-03-26
MPS-1897: Grant Folder Access Fix

WhosNickDoglio:
#### Summary
While working on [MPS-1897](https://vimean.atlassian.net/browse/MPS-1897) I noticed some issues from my initial networking PR.

- The URI being pulled from the `Folder` for granting/revoking folder access was incorrect and needed to be pulled from the `teamMembers` connection.
- URIs being passed for granting/revoking folder access to team members needed to be the URI for the `TeamMembership` object and not the `User` object, `grantTeamMembersFolderAccess` has been updated to take the correct model. 
- I also moved some kdocs to using `@param`


anthonycr (approved):
LGTM

PR #481, opened 2021-04-06
Allow non ASCII characters as a user agent

anthonycr:
# Issue
Some Android devices have non-ASCII characters in their user agent. This caused our networking library to crash if the user agent was set on the library, as OkHttp by default disallows non-ASCII characters in header values.

## Summary
The solution was to add the user-agent header using the "unsafe" version of the header creation builder that allows non-ASCII characters.

## How to Test
Run the test that verifies that the unsafe header value is allowed.

### DIFF
--- a/api-core/src/main/java/com/vimeo/networking2/internal/interceptor/UserAgentHeaderInterceptor.kt
+++ b/api-core/src/main/java/com/vimeo/networking2/internal/interceptor/UserAgentHeaderInterceptor.kt
@@ -32,8 +32,9 @@ import okhttp3.Response
  * @param userAgent The user agent that should be sent with every request.
  */
 class UserAgentHeaderInterceptor(private val userAgent: String) : Interceptor {
-    override fun intercept(chain: Interceptor.Chain): Response =
-        chain.proceed(chain.request().newBuilder().header(HEADER_USER_AGENT, userAgent).build())
+    override fun intercept(chain: Interceptor.Chain): Response = chain.proceed(chain.request().run {
+        newBuilder().headers(headers().newBuilder().addUnsafeNonAscii(HEADER_USER_AGENT, userAgent).build()).build()
+    })
 
     companion object {
         private const val HEADER_USER_AGENT = "User-Agent"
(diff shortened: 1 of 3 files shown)


api-core/src/test/java/com/vimeo/networking2/internal/interceptor/UserAgentHeaderInterceptorTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import okhttp3.Connection
+import okhttp3.Interceptor
+import okhttp3.Protocol
+import okhttp3.Request
+import okhttp3.Response
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import java.util.concurrent.TimeUnit
+
+class UserAgentHeaderInterceptorTest {
```
WhosNickDoglio, line 13: yay tests!

api-core/src/main/java/com/vimeo/networking2/internal/interceptor/UserAgentHeaderInterceptor.kt
```diff
@@ -32,8 +32,9 @@ import okhttp3.Response
  * @param userAgent The user agent that should be sent with every request.
  */
 class UserAgentHeaderInterceptor(private val userAgent: String) : Interceptor {
-    override fun intercept(chain: Interceptor.Chain): Response =
-        chain.proceed(chain.request().newBuilder().header(HEADER_USER_AGENT, userAgent).build())
+    override fun intercept(chain: Interceptor.Chain): Response = chain.proceed(chain.request().run {
+        newBuilder().headers(headers().newBuilder().addUnsafeNonAscii(HEADER_USER_AGENT, userAgent).build()).build()
+    })
```
WhosNickDoglio: **Optional:** This feels more terse then concise to me, it too me a minute to understand what was happening with the `run` call here, I think it would be more readable broken up but not a blocker.

WhosNickDoglio (approved):
LGTM!

api-core/src/main/java/com/vimeo/networking2/internal/interceptor/UserAgentHeaderInterceptor.kt
anthonycr: I decided to use `let` with a named parameter and split out the header creation into a separate local property. I think it's a little more legible now.

PR #483, opened 2021-04-26
Support for subfolders

anthonycr:
# Summary
- Added the `metadata.interactions.add_subfolder` interaction to the `Folder`, creating the `AddSubfolderInteraction` and `FolderInteractions` classes.
- Added the `metadata.connections.ancestor_path` connection to the `Folder`, creating the `FolderAncestorConnection`.
- Added the `metadata.connections.parent_folder` connection to the `Folder`, creating the `DefaultConnection` for that.
- Updated the `addFolder` function to `VimeoApiClient` to accept a `parentFolder: Folder?` or a `parentFolderId: String?` to support adding subfolders.
- Added a `shouldDeleteClips: Boolean` parameter to `deleteFolder` to support conditionally deleting all the content in a folder when deleting it.

### DIFF
--- a/request/src/main/java/com/vimeo/networking2/VimeoService.kt
+++ b/request/src/main/java/com/vimeo/networking2/VimeoService.kt
@@ -191,6 +191,7 @@ internal interface VimeoService {
     fun createFolder(
         @Header(AUTHORIZATION) authorization: String,
         @Url uri: String,
+        @Field(PARENT_FOLDER_ID) parentFolderId: String?,
         @Field(PARAMETER_FOLDER_NAME) name: String,
         @Field(PARAMETER_FOLDER_PRIVACY) privacy: FolderViewPrivacyType,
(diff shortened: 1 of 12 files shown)


request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -211,6 +212,14 @@ internal interface VimeoService {
         @Field(SLACK_USER_PREF) slackUserPref: SlackUserPreferenceType?
     ): VimeoCall<Folder>
 
+    @FormUrlEncoded
+    @HTTP(method = "DELETE", hasBody = true)
```
anthonycr, line 216: Note: passing a body in a `DELETE` is not standard, and as such is not supported by default by Retrofit. In order to properly implement it we have to use this lower level API.

elhicks (approved):

PR #484, opened 2021-04-29
Making sure parcelable and serializable are both generated

anthonycr:
# Summary
The parcelable models were not being generated for some time, likely due to a change in the gradle version or android plugin. 

I was able to identify that the root cause was that the android plugin could not be found at the time of configuration. The solution was to use the more modern `withPlugin` expression that would run the lambda after the plugin was applied. This exposed another issue, where the `preBuild` task could not have a dependency added to it as we were initially doing it. The solution to this issue was to attach it in `afterEvaluate` and to find the real path of the task dependency after it was configured.

In order to prevent this issue from happening without us knowing it in the future, I added a couple tests to the serializable and parcelable test suites that verifies that the generated files are not empty.

Additionally, I made the example depend on the parcelable models since it is an android project.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -8,7 +8,7 @@ android {
 
     defaultConfig {
         applicationId "com.vimeo.example"
-        minSdkVersion 21
+        minSdkVersion 23
         targetSdkVersion 30
         versionCode 1
         versionName "1.0"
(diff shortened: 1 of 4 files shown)


WhosNickDoglio (approved):
Nice Fix 🙌

PR #485, opened 2021-04-30
Explicit path parameters

anthonycr:
# Summary
Right now all the functions of `VimeoApiClient` can make a request to any fully qualified URL. The `uri` parameter accepts a full URL rather than just the path. This creates the possibility of making requests to other domains with headers that should only be sent to the vimeo domain. For example, since your auth token is sent in the header, we wouldn't want to send that to a different domain.

Since this was not intentional design, and the reason that this current behavior exists is because it was the simplest implementation, I am removing that behavior here. Instead of the URI being annotated with `@Url`, it is now annotated with `@Path("path")` and the request definition will define "path" as the whole path, e.g. `@GET("{path}")`. This prevents the library from being able to hit a different domain than intended.

A side effect of this change is that path traversal is no longer supported, so passing in `/user/../videos` will no longer make a request to `/videos`. We feel that this behavior was not previously necessary as the URIs provided to us by the API do not utilize this traversal.

Only one instance of `@Url` remains in the library, within the  `AuthService`, `createSsoGrantRequest(...)`, and this is intentionally left, since we actually need to construct a request to an external URL for the consumer to direct users to for log in via SSO.

### DIFF
--- a/request/src/main/java/com/vimeo/networking2/VimeoService.kt
+++ b/request/src/main/java/com/vimeo/networking2/VimeoService.kt
@@ -84,7 +84,7 @@ internal interface VimeoService {
     @GET("{path}")
     fun getPublishJob(
         @Header(AUTHORIZATION) authorization: String,
-        @Path("path") uri: String,
+        @Path("{path}", encoded = true) uri: String,
         @Query(FIELD_FILTER) fieldFilter: String?,
         @Header(CACHE_CONTROL) cacheControl: CacheControl?
     ): VimeoCall<PublishJob>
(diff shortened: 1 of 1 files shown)


anthonycr:
cc @rpassis

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -81,33 +81,33 @@ internal interface VimeoService {
         @Path(TYPE) type: ConnectedAppType
     ): VimeoCall<Unit>
 
-    @GET
+    @GET("{path}")
     fun getPublishJob(
         @Header(AUTHORIZATION) authorization: String,
-        @Url uri: String,
+        @Path("path", encoded = true) uri: String,
```
howardrigberg, line 87: Optional: you could change the corresponding parameter names in this file from `uri` to `path` as well for clarity.

howardrigberg (approved):
LGTM. Just one optional suggestion.

request/src/main/java/com/vimeo/networking2/VimeoService.kt
anthonycr: Going to leave it as "URI" already didn't match "URL" and is named so to match the vimeo API's usage of the word "URI" rather than retrofit's.

PR #487, opened 2021-05-12
Allow prepopulated query parameters while disallowing path traversal and other hosts

anthonycr:
Recent changes to disallow path traversal and making requests to hosts other than the Vimeo API were made. Unfortunately these changes prevent prepopulated query params from being attached to a URI and a request being made to that URI. Since we rely heavily on making request to a URI that already has these params populated (for example the next page URI in a paging request, which has the field filters) we need to support this use case.

The solution I found was to revert to using `@Url` instead of `@Path` and manually prevent path traversal by using the existing validation code we had to validate that a non empty URI was used. Now the `notEmpty` function is `validate` and it avoids us making a request to a URI with a path traversal. In order to prevent erroneously making a request to a non Vimeo URL, I added the `HostValdidatorInterceptor`, which throws an `IOException` when the host does not match the host specified in the `VimeoApiConfiguration`.

### DIFF
--- a/request/src/main/java/com/vimeo/networking2/VimeoService.kt
+++ b/request/src/main/java/com/vimeo/networking2/VimeoService.kt
@@ -81,33 +81,33 @@ internal interface VimeoService {
         @Path(TYPE) type: ConnectedAppType
     ): VimeoCall<Unit>
 
-    @GET("{path}")
+    @GET
     fun getPublishJob(
         @Header(AUTHORIZATION) authorization: String,
-        @Path("path", encoded = true) uri: String,
+        @Url uri: String,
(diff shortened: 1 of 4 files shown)


elhicks (approved):
LGTM!

request/src/main/java/com/vimeo/networking2/VimeoService.kt
```diff
@@ -216,7 +216,7 @@ internal interface VimeoService {
     @HTTP(method = "DELETE", path = "{path}", hasBody = true)
```
howardrigberg, line 216: `@HTTP(method = "DELETE", hasBody = true)` ?

PR #491, opened 2021-06-03
Use parent_folder_uri instead of parent_folder_id

anthonycr:
# Summary
We don't want to have to manipulate the URI of a folder in order to create a subfolder in it, so we should use `parent_folder_uri` instead and let the API manipulate it for us.


elhicks (approved):
LGTM

PR #493, opened 2021-06-17
VCANDROID-1140 Add the new fields into editSession

abalcer-ls:
#### Issue
https://vimean.atlassian.net/browse/VCANDROID-1140

#### Summary
Add the following fields into editSession:
 - vsid
 - has_watermark
 - is_rated
- result_video_hash

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/EditSession.kt
+++ b/models/src/main/java/com/vimeo/networking2/EditSession.kt
@@ -18,5 +18,11 @@ data class EditSession(
      * Whether the current version of clip is of max resolution.
      */
     @Json(name = "is_max_resolution")
-    val isMaxResolution: Boolean? = null
+    val isMaxResolution: Boolean? = null,
+
+    /**
+     * Video session id
+     */
+    @Json(name = "vsid")
+    val vsid: String? = null
 )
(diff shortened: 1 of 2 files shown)


models/src/main/java/com/vimeo/networking2/EditSession.kt
```diff
@@ -18,5 +18,11 @@ data class EditSession(
      * Whether the current version of clip is of max resolution.
      */
     @Json(name = "is_max_resolution")
-    val isMaxResolution: Boolean? = null
+    val isMaxResolution: Boolean? = null,
+
+    /**
+     * Video session id
```
anthonycr: There's new detekt rules that require these property docs to be moved to parameter docs. The existing properties were added to the baseline and we've been moving them as we change each file. So this should be move to the class docs as `@param vsid Video session id`. Detekt doesn't require it, but if you have the time to move the docs for the other two properties in this class and then run `./gradlew detektBaseline` that would be appreciated.

See also `Folder.kt` for an example of where docs are required to be now.

anthonycr (changes requested):

RomanMinenok (approved):

anthonycr (approved):
Looks good!

anthonycr:
Latest alpha build including these changes is available, `2.0.0-alpha.93`: https://github.com/vimeo/vimeo-networking-java/releases/tag/2.0.0-alpha.93

PR #502, opened 2021-07-27
MPS-2372 User survey support

howardrigberg:
# Summary
Currently there is a fixed endpoint for getting survey responses for a user at /surveys available on this build: https://mps-xxx-usersurveyapiendpoints-api.ci.vimeows.com/ 

PR for networking API is here: https://github.vimeows.com/Vimeo/vimeo/pull/63516

We need to add the endpoint to the service/client and new models to the networking library.

## Description
The new endpoint `fetchSurveyQuestionList` returns a `UserSegmentSurveyList` representing a list of surveys that can be displayed to the authorized user. If a user has already answered a survey, it will no longer be returned from the API. A survey is represented by `UserSegmentSurvey`. It contains a list of `SurveyQuestion` that in turn contain a list of allowable responses represented by `SurveyResponseChoice`. Currently there is only one functional type of survey that will be offered: multiple choice answers where a user can only select one. More may be supported in the future. This is captured in a `SurveyQuestionType`.


anthonycr (approved):
Looks good, and nice catch on the missing rename in the delegate. If you need this released right away, consider creating a 2.0.1 release, since it's not breaking, and adding to the `CHANGELOG.md` and updating the version in `gradle.properties` and `ApiConstants.kt`. Otherwise we can pick this up whenever we get to the next release.

PR #510, opened 2021-08-10
VCANDROID-1320 Add edit_session.min_tier_for_movie field

ievgenii-vimeo:
# Summary
 Add edit_session.min_tier_for_movie field (https://vimean.atlassian.net/browse/VCANDROID-1320)

## Description
New field in EditSession model

## How to Test
Use the API with a video model

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/EditSession.kt
+++ b/models/src/main/java/com/vimeo/networking2/EditSession.kt
@@ -12,6 +12,7 @@ import com.squareup.moshi.JsonClass
  * @param resultVideoHash The result video hash.
  * @param hasWatermark Whether the video has watermark.
  * @param isRated Whether the video is rated.
+ * @param minTierForMovie Whether video requires paid plan. .
  */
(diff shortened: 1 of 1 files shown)


howardrigberg (approved):
LGTM

anthonycr (approved):

models/src/main/java/com/vimeo/networking2/EditSession.kt
```diff
@@ -12,6 +12,7 @@ import com.squareup.moshi.JsonClass
  * @param resultVideoHash The result video hash.
  * @param hasWatermark Whether the video has watermark.
  * @param isRated Whether the video is rated.
+ * @param minTierForMovie Whether video requires paid plan.
```
howardrigberg: I approved it, but I do have a question. The description sounds like this value should be a boolean instead of a string. Can you explain a little more about the potential values?

ievgenii-vimeo: I changed the description. Thanks

PR #512, opened 2021-08-24
Fix BillingPeriodType values

WhosNickDoglio:
# Summary
The `values` for the `BillingPeriodType` didn't match what we're seeing in the API response. This PR just fix that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 Change Log
 ==========
+
+Version 2.0.5 *(2021-08-24)*
+----------------------------
+- Fixed incorrect values for `BillingPeriodType`.
+
 Version 2.0.4 *(2021-08-10)*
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 Change Log
 ==========
+
+Version 2.0.5 *(2021-08-24)*
```
anthonycr, line 4: Version also needs to be bumped in the `ApiConstants.SDK_VERSION` constant.

anthonycr (changes requested):
Good catch, one quick thing needs changing before we can merge.

CHANGELOG.md
WhosNickDoglio: Good catch, fixed now 👍

anthonycr (approved):

PR #513, opened 2021-08-24
[TECH] Migrate sample to ViewBinding

WhosNickDoglio:
# Summary
The example was still using Kotlin synthetics which have been depreciated in favor of ViewBinding, this PR just covers migrating the sample to use ViewBinding. 

## How to Test
Run the example app after adding the necessary information and confirm it works as expected.


anthonycr (approved):
LGTM

PR #516, opened 2021-09-16
[TECH] Disable incremental compiling for model generator plugin 

WhosNickDoglio:
## Description
When the model generator plugin was originally written it was intended to support incremental compilation of the new models to speed up build time when testing changes to models.  Unfortunately I didn't quite set up incremental support properly and it has a habit of breaking builds sometimes.  This PR removes the broken incremental support. 

## How to Test
Run `gradle clean build` modify a model, run `gradle build` and confirm you don't see a "Cannot query incremental changes` error.


WhosNickDoglio:
Closing in favor of #517

PR #517, opened 2021-09-16
Potential incremental model creation task fix

elhicks:
# Summary
Sometimes model generation breaks, this is an attempt to fix it.

## Description
Create a stable handle to the model files, which might potentially work around a concurrency/query problem which might result in the error seen.  Being concurrency related might also explain how sporadic the error is.

## How to Test
Modify some model classes and run models-parcelable:generateModels tasks a bunch of times.  I just added a random fake member, run the task, removed the fake class member, and run the task again.

### DIFF
--- a/model-generator/plugin/src/main/java/com/vimeo/modelgenerator/tasks/GenerateModelsTask.kt
+++ b/model-generator/plugin/src/main/java/com/vimeo/modelgenerator/tasks/GenerateModelsTask.kt
@@ -8,13 +8,14 @@ import com.vimeo.modelgenerator.visitor.SerializableClassVisitor
 import com.vimeo.modelgenerator.visitor.SerializableInterfaceVisitor
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.ConfigurableFileTree
-import org.gradle.api.plugins.JavaPluginConvention
+import org.gradle.api.plugins.JavaPluginExtension
 import org.gradle.api.tasks.Input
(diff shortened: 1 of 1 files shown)


model-generator/plugin/src/main/java/com/vimeo/modelgenerator/tasks/GenerateModelsTask.kt
```diff
@@ -48,12 +48,15 @@ open class GenerateModelsTask : DefaultTask() {
     private val output =
         project.convention.getPlugin(JavaPluginConvention::class.java).sourceSets.maybeCreate("main").java.srcDirs.first().path
```
WhosNickDoglio:
```suggestion
        project.extensions.getByType(JavaPluginExtension::class.java).sourceSets.maybeCreate("main").java.srcDirs.first().path
```

**Optional:** This was in my PR (#516), the `convention` getter and `JavaPluginConvention` have both been deprecated so this is the recommended replacement.

WhosNickDoglio (approved):
Great find

I tried a bunch of stuff and was only able to break it by deleting files.  We could get around this by filtering out removed files by updating `files` to look like this: 


```kotlin
        val files: Iterable<File> =
            if (inputChanges.isIncremental) inputChanges.getFileChanges(models).asIterable()
                .filter { it.changeType != ChangeType.REMOVED }
                .map { it.file } else models.files
```

Overall very happy to see this fixed 🎉

anthonycr (approved):

PR #524, opened 2021-09-30
[TECH] Version Catalogs

WhosNickDoglio:
## Description

This PR focusing on adopting the new [Gradle Version Catalog API](https://docs.gradle.org/current/userguide/platforms.html).

Dependency management was moved from the root `build.gradle` into a `libs.versions.toml` file that can be shared with the model generator plugin. 

## How to Test
Confirm the project still builds and works as expected.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 detekt = "1.18.1"
 moshi = "1.12.0"
-kotlin = "1.5.21"
+kotlin = "1.5.31"
 retrofit = "2.9.0"
 
 [plugins]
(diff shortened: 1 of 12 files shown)


gradle/libs.versions.toml
```diff
@@ -0,0 +1,44 @@
...
+[plugins]
+detekt = { id = "io.gitlab.arturbosch.detekt", version.ref = "detekt" }
+ben-manes = { id = "com.github.ben-manes.versions", version = "0.39.0" }
+kotlin = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
+
+[libraries]
+
+#Plugins
+android-gradle = { module = "com.android.tools.build:gradle", version = "7.0.2" }
+kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version = "1.5.30" }
```
anthonycr: I would think that we should always match the plugin and the kotlin version. Is there a reason I'm forgetting about not to?

anthonycr (approved):
Left one suggestion, though not strictly necessary. Looks good overall.

gradle/libs.versions.toml
WhosNickDoglio: Oops 🙃

Yeah you're correct I'll fix that now

PR #527, opened 2021-11-02
MPS-2644:  Add missing upload error codes

elhicks:
## Summary

Some of the upload error codes recently changed on the backend.  This is an update to match those changes on the networking library side.

This is related to a bug recently filed on the core app where the codes changed in the last few months.

## How to Test
Verify everything builds as expected and there are no regressions.


anthonycr (approved):
Looks good

PR #532, opened 2021-12-13
[TECH] Migrate from Moshi KAPT to Moshi KSP

WhosNickDoglio:
# Summary
[KSP ](https://github.com/google/ksp) is a new API for creating compiler plugins that Google worked with JetBrains to create as a replacement for KAPT. As of the [most recent Moshi release](https://github.com/square/moshi/blob/master/CHANGELOG.md) Moshi now has an official KSP artifact.  This PR just migrates from using the KAPT artifact to the KSP artifact. 

## Description
Added the ksp plugin to our `libs` file and replaced KAPT with KSP for all the model modules. 

## How to Test
Run `gradle build` and verify the Moshi adapters are generated expected in the `build/generated/ksp/main/kotlin` directory for each models module.


build.gradle
```diff
@@ -13,11 +13,12 @@ buildscript {
 plugins {
     alias(libs.plugins.detekt)
     alias(libs.plugins.ben.manes)
+    alias(libs.plugins.ksp) apply false
```
anthonycr, line 16: Why is apply false here? Is the point to just add the plugin to the namespace so it can be applied at the module level?

anthonycr (approved):

build.gradle
WhosNickDoglio: If I didn't apply false this plugin would be applied to our root project which isn't what it's meant for _but_ since I applied it here with a version number I can just apply the ID at the module level and maintain the version in our libs file.  If I just applied it at each individual module level I would have to do `apply plugin: 'com.google.devtools.ksp' version "1.5.31-1.0.1"` for each module and the would be annoying to maintain.

PR #534, opened 2022-02-10
VCANDROID-1583: Keep access token on server with flag in logOut

vbevans94:
# Summary
Don't delete access token on server if corresponding flag is passed to logOut.

## Description
New parameter added. Added tests to cover the flows.

## How to Test
./gradlew test

### DIFF
--- a/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
+++ b/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
@@ -303,10 +303,14 @@ internal class AuthenticatorImpl(
         ).enqueueWithAccountStore(callback)
     }
 
-    override fun logOut(callback: VimeoCallback<Unit>): VimeoRequest {
+    override fun logOut(callback: VimeoCallback<Unit>, keepOnServer: Boolean): VimeoRequest {
         val accessToken = currentAccount?.accessToken
         accountStore.removeAccount()
         accessToken ?: return NoOpVimeoRequest
+
+        if (keepOnServer) {
+            return NoOpVimeoRequest
+        }
 
         return authService.logOut(authorization = "Bearer $accessToken").enqueue(callback)
     }
--- a/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
+++ b/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
@@ -303,16 +303,16 @@ internal class AuthenticatorImpl(
         ).enqueueWithAccountStore(callback)
     }
 
-    override fun logOut(callback: VimeoCallback<Unit>, keepOnServer: Boolean): VimeoRequest {
+    override fun logOut(callback: VimeoCallback<Unit>): VimeoRequest {
         val accessToken = currentAccount?.accessToken
         accountStore.removeAccount()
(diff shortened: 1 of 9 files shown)


RomanMinenok (approved):

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -303,11 +303,15 @@ internal class AuthenticatorImpl(
         ).enqueueWithAccountStore(callback)
     }
 
-    override fun logOut(callback: VimeoCallback<Unit>): VimeoRequest {
+    override fun logOut(callback: VimeoCallback<Unit>, keepOnServer: Boolean): VimeoRequest {
```
anthonycr: The `callback` won't be invoked if the flag is set to true, which breaks the contract of the function. I realize that there's already a no-op condition here if the access token is already null, but that's a case we shouldn't emulate.

There's two options here then that I see, neither of which I have a strong preference for since I'm not going to be the initial consumer. I'll leave it up to you to choose or if you have another suggestion I'm open.

1. Create a separate function called `logOutLocally()` that has no parameters and no return type and that just calls
    ```kotlin
    accountStore.removeAccount()
    ```
    and doesn't invoke any callback or return a `VimeoRequest`. Docs would be something like `"logs out the library locally without logging out of the token on the server."`

2. Keep this function signature as is, and manually invoke `callback.onSuccess`. This requires changing the `LocalCallVimeoAdapter` by adding an `enqueueSuccess` function
    ```kotlin
    /**
     * Enqueue the [VimeoCallback] with the provided [data] response.
     */
    fun <T> enqueueSuccess(data: T, callback: VimeoCallback<T>): VimeoRequest {
        callbackExecutor.execute {
            callback.onSuccess(
                VimeoResponse.Success(
                    data,
                    ResponseOrigin.CACHE,
                    ApiConstants.NONE
                )
            )
        }

        return NoOpVimeoRequest
    }
    ```
    Then the callback can be invoked as follows (we could also use this opportunity to correct the behavior of the `accessToken ?: return NoOpVimeoRequest` check)
    ```kotlin
    // Instead of
    return NoOpVimeoRequest
    // Do this
    return run {
        localVimeoCallAdapter.enqueueSuccess(Unit)
        NoOpVimeoRequest
    }
    ```

auth/src/test/java/com/vimeo/networking2/internal/AuthenticatorImplTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.vimeo.networking2.vimeoCallback
+import org.junit.Test
+import org.mockito.Mockito.mock
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.verifyNoInteractions
+
+class AuthenticatorImplTest {
```
anthonycr, line 17: These tests are great!

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
vbevans94: Hey @anthonycr! Thanks for review. Took first option since it was my initial idea.

anthonycr (approved):
Looks great!

PR #538, opened 2022-02-16
VCANDROID-1628 Add header to getMagistoTeamToken call

alex-shteinle:
# Summary
https://vimean.atlassian.net/browse/VCANDROID-1628
Add auth header to getMagistoTeamToken api call. 

## Description
Add auth header to getMagistoTeamToken api call. 

## How to Test
Try to obtain team token with the valid team id.

### DIFF
--- a/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
+++ b/auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
@@ -317,7 +317,12 @@ internal class AuthenticatorImpl(
     }
 
     override fun getMagistoTeamToken(teamId: String, callback: VimeoCallback<TeamToken>): VimeoRequest {
-        return authService.getMagistoTeamToken(teamId).enqueue(callback)
+        return authService
+            .getMagistoTeamToken(
+                authorization = authenticationMethod.basicAuthHeader,
+                teamId = teamId
+            )
+            .enqueue(callback)
     }
 
     private fun VimeoCall<VimeoAccount>.enqueueWithAccountStore(callback: VimeoCallback<VimeoAccount>): VimeoRequest =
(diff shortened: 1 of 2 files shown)


RomanMinenok (approved):

auth/src/main/java/com/vimeo/networking2/internal/AuthenticatorImpl.kt
```diff
@@ -317,7 +317,12 @@ internal class AuthenticatorImpl(
     }
 
     override fun getMagistoTeamToken(teamId: String, callback: VimeoCallback<TeamToken>): VimeoRequest {
-        return authService.getMagistoTeamToken(teamId).enqueue(callback)
+        return authService
+            .getMagistoTeamToken(
+                authorization = authenticationMethod.basicAuthHeader,
```
anthonycr: I'm not a user of this endpoint, so I am not 100% sure, but after reading the code for the endpoint it seems that the user needs to be already authenticated in order to get the team token. The `basicAuthHeader` is a logged out auth header that will not point to any specific user and will result in a logged out failure. We need to pass the correct auth header to this function.

We should do this by using 
```kotlin
authorization = currentAccount?.accessToken?.let { "Bearer $it" } ?: authenticationMethod.basicAuthHeader
```
instead of just `authenticationMethod.basicAuthHeader`. This way it uses the real logged in account, and if the account isn't logged in, it falls back to the basic auth header which will trigger the appropriate failure.

Let me know if I'm misunderstanding the endpoint. We just need a little more verification on this before merging it. If you're having trouble testing, you can always use `gradle publishToMavenLocal` to deploy a local version of the library to your machine's maven repo.

alex-shteinle: Thanks a lot! Your point was right. We need to use accessToken. I've check with local version of the library. It works as expected.

anthonycr (approved):
Looks good!

PR #543, opened 2022-03-08
MPS-2939 Adding an id to Live

howardrigberg:
We needed to expose a Live identifier for logging purposes in the application


anthonycr (approved):
LGTM!

PR #546, opened 2022-03-24
Remove certificate pinning

anthonycr:
# Summary
The PR completely removes the certificate pinning ability from the library.

## Description
We are no longer using certificate pinning as it creates problems for the Vimeo security team around managing certificates. It does not allow us to automatically swap to a new certificate on the fly and requires careful coordination with the apps using certificate pinning to avoid breaking those apps when a certificate needs to be renewed.

While we could leave in certificate pinning and default the value to false, we don't want to encourage anyone to use certificate pinning so that API certificates can be renewed without updating this library. As a result, we are completely removing the ability to pin certificates, as it is no longer recommended.

This is a breaking change and will be versioned behind a 3.0 version. Previous 2.x and 1.x versions of this library will stop working with certificate pinning enabled when the current intermediate certificate expires on **3 March 2023**. For any consumers that are still on those versions of the networking library, they will need to either disable certificate pinning explicitly, or upgrade to a 3.x version.

## How to Test
Verify that tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,6 @@ In the below sections, we cover examples of ways to customize the `VimeoApiClien
 
 We recommend the following for all apps when configuring the SDK with the `VimeoApiConfiguration.Builder`:
 - Provide a user agent to identify your app and the device using `withUserAgent(String)`. By default the library creates its own user agent, but supplementing it with more info provides the Vimeo API with a better picture of what devices are accessing it. Not required, and we provide a default if none is provided.
-- Decide if you need certificate pinning. If you don't care about certificate pinning, or if it's not working on the Android version you're working with, we suggest to disable it using `withCertPinning(Boolean)`. It is enabled by default.
 - Provide a cache directory with `withCacheDirectory(File)`. By default the SDK does not set the cache directory and **responses will not be cached**, so you'll have to choose one to get the best performance.
 - Provide an `AccountStore` using `withAccountStore(AccountStore)`. Unless you are accessing the API with a fixed access token, then you should probably provide a store that is disk backed. The default used by the SDK is an in-memory store that **will not remember the logged in account** after the app is restarted.
 
--- a/api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
+++ b/api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
@@ -33,8 +33,6 @@ object ApiConstants {
 
     const val SSL_URL_PATTERN = "*.vimeo.com"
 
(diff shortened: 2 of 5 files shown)


api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
```diff
@@ -33,8 +33,6 @@ object ApiConstants {
 
     const val SSL_URL_PATTERN = "*.vimeo.com"
 
-    const val SSL_PUBLIC_KEY = "sha256/5kJvNEMw0KjrCAu7eXY5HZdvyCS13BbA0VJG1RSP91w="
-
     const val SDK_VERSION = "2.4.0"
```
shanestaples, line 36: Just to make sure, I'm guessing we wait to change the version as part of the release process?

README.md
```diff
@@ -131,7 +131,6 @@ In the below sections, we cover examples of ways to customize the `VimeoApiClien
 
 We recommend the following for all apps when configuring the SDK with the `VimeoApiConfiguration.Builder`:
 - Provide a user agent to identify your app and the device using `withUserAgent(String)`. By default the library creates its own user agent, but supplementing it with more info provides the Vimeo API with a better picture of what devices are accessing it. Not required, and we provide a default if none is provided.
-- Decide if you need certificate pinning. If you don't care about certificate pinning, or if it's not working on the Android version you're working with, we suggest to disable it using `withCertPinning(Boolean)`. It is enabled by default.
```
shanestaples, line 134: Did you want to also add the warning about upgrading to v3 and deprecation in March 2023, or will you add that as part of another PR?

shanestaples (approved):
LGTM! I had a few questions / comments, otherwise +1 to merge.

shanestaples:
> LGTM! I had a few questions / comments, otherwise +1 to merge.

Doh, ignore both of those comments, I see now that they are both in a separate PR which has been assigned to me as well. Good to merge!

api-core/src/main/java/com/vimeo/networking2/ApiConstants.kt
anthonycr: yes changing as part of the release process. wanted to split the PRs into two since it's a major release.

anthonycr:
@shanestaples don't worry, good questions, answering them anyway for anyone reading this PR in the future that hasn't been involved in our discussions about this.

README.md
anthonycr: Updating the readme with a deprecation warning when I prep the 3.0.0 release.

PR #555, opened 2022-04-27
Adding SsoConnections DTO and associated endpoint

anthonycr:
# Summary
Some SSO connected accounts are not bound to a specific domain, but are email specific. As a result, these accounts can't log in with the current SSO domain method. This PR introduces a way for these accounts to log in with SSO even if they aren't bound by domain.

## Description
- Added `SsoConnection`
- Added `SsoConnectionInteractions`
- Added `checkSsoConnection` to `Authenticator`
- Added `createSsoAuthorizationUri` that accepts an `SsoConnection` to `Authenticator`.
- Deprecated all the old DTOs and functions associated with SSO in favor of the above.

The SSO flow follows nearly identically to the old one, so I am not going to completely describe it here. The difference is that instead of calling `fetchSsoDomain` by passing a `domain` as a query parameter, we call `checkSsoConnection` and set the `email` parameter. This returns to us an `SsoConnection` instead of an `SsoDomain`. We then call `createSsoAuthorizationUri` by passing the connection instead of the domain and then the flow proceeds the same as it did previously.

## How to Test
- Run the tests
- Verify that login with SSO works by following the steps partially outlined above.

### DIFF
--- a/auth/src/main/java/com/vimeo/networking2/Authenticator.kt
+++ b/auth/src/main/java/com/vimeo/networking2/Authenticator.kt
@@ -265,6 +265,23 @@ interface Authenticator {
     ): VimeoRequest
 
     /**
+     * Create a URI which can be used to log in the user via a browser and will redirect to a URI that can be exchanged
+     * using [authenticateWithSso] for a logged in account. Obtain the [SsoConnection] required by this function first
+     * using [checkSsoConnection].
+     *
+     * @param ssoConnection The connection info obtained from the API that specifies the connection URI which will be
+     * used as the base for this URI. NOTE: The connection URI (obtained from [SsoConnectionInteractions.connect]) must
+     * not be null, if a domain is provided with a null connection, then this function will return null.
+     * @param responseCode An arbitrary response code that can be used to verify the origin of the redirect URI. The
+     * API will return this value to later as a security measure in a query string parameter named `state` in the
+     * callback URI.
+     *
+     * @return The URI which can be opened in a browser, or null if the parameters provided were invalid.
+     */
+    @Internal
+    fun createSsoAuthorizationUri(ssoConnection: SsoConnection, responseCode: String): String?
+
+    /**
      * Find a supported SSO domain that matches the [domain] parameter. If this request returns a valid [SsoDomain],
      * then SSO authentication can be initiated, starting with the creation of the SSO authorization URI via
      * [createSsoAuthorizationUri].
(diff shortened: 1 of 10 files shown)


auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -258,6 +291,7 @@ interface Authenticator {
      *
      * @return A [VimeoRequest] object to cancel API requests.
      */
+    @Deprecated("Deprecated in favor of checkSsoConnection")
```
howardrigberg, line 294: Optional: you could provide a clearer explanation if you make use of `replaceWith` in the `Deprecated` annotation.

```
@Deprecated(
        message = "This function has been deprecated in favor of using an email-based approach.",
        replaceWith = ReplaceWith(
            expression = "checkSsoConnection"
        )
    )
```

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
```diff
@@ -278,6 +312,7 @@ interface Authenticator {
      *
      * @return The URI which can be opened in a browser, or null if the parameters provided were invalid.
      */
+    @Deprecated("Deprecated in favor of the function by the same name that accepts an SsoConnection")
```
howardrigberg, line 315: Optional: you could provide a clearer explanation if you make use of replaceWith in the Deprecated annotation.

```
@Deprecated(
        message = "Deprecated in favor of the function by the same name that accepts an SsoConnection.",
        replaceWith = ReplaceWith(
            expression = "createSsoAuthorizationUri(ssoConnection: SsoConnection, responseCode: String)"
        )
    )
```

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
howardrigberg: Follow up question: Why are the `fetchSsoDomain` and `createSsoAuthorizationUri` deprecated instead of deleted?

howardrigberg (approved):
Looks good. Just a couple of optional suggestions with deprecation documentation and a related question.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: Because the approach still technically works, and it is a better practice to deprecate functionality before removing it in order to give any consumers that were using it some leeway.

auth/src/main/java/com/vimeo/networking2/Authenticator.kt
anthonycr: I was hoping to use something like this, but I was having trouble getting these `ReplaceWith` to work right and what's here is also giving me the same problem. They aren't producing compileable code for me, so I'm going to merge as is.

PR #561, opened 2022-05-04
[TECH] Add convenience extension for TeamPermission data class for getting the current PermissionPolicy

elhicks:
# Summary
Unfortunately we only have the uri provided for the current permission policy.  Applicable permission policies however do most likely contain the name of the policy the uri belongs to. It's possible you could have an assigned policy which is no longer available to select again. In that case you wouldn't produce a match using this method and result in a null PermissionPolicy. For the sake of how this is used, I believe this is an acceptable quirk.  

## How to Test

Make sure the app compiles right, and analyze the code for any logical faults.

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/TeamPermission.kt
+++ b/models/src/main/java/com/vimeo/networking2/TeamPermission.kt
@@ -32,3 +32,16 @@ data class TeamPermission(
 ) : Entity {
     override val identifier: String? = teamEntity?.identifier
 }
+
+/**
+ * Get the current [PermissionPolicy] if possible; This search is predicated upon
+ * [TeamPermission.applicablePermissionPolicies] containing an item with the same uri as
+ * [TeamPermission.currentPermissions], [TeamPermissionCurrentPermissions.permissionPolicyUri].
+ */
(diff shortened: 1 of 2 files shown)


elhicks:
I noticed the metadata on TeamGroup wasn't an optional with a default null so I corrected that as well.

models/src/main/java/com/vimeo/networking2/TeamPermission.kt
```diff
@@ -32,3 +32,15 @@ data class TeamPermission(
 ) : Entity {
     override val identifier: String? = teamEntity?.identifier
 }
+
+/**
+ * Get the current [PermissionPolicy] if possible; This search is predicated upon
+ * [TeamPermission.applicablePermissionPolicies] containing an item with the same uri as
+ * [TeamPermission.currentPermissions], [TeamPermissionCurrentPermissions.permissionPolicyUri].
+ */
+val TeamPermission.currentPermissionPolicy: PermissionPolicy? get() =
```
howardrigberg, line 41: Probably should move this to application level. See relevant comments on Slack.

howardrigberg (changes requested):

PR #562, opened 2022-05-05
Make metadata nullable with default value of null for TeamGroup

elhicks:
# Summary
Make metadata nullable with default value of null for TeamGroup

## Description
Make metadata nullable with default value of null for TeamGroup

## How to Test
Library builds successfully.


anthonycr (approved):
LGTM

PR #565, opened 2022-06-08
Update PermissionPolicy with missing fields

elhicks:
# Summary
Added display name and display description properties to PermissionPolicy.  Implemented Entity since we have a unique uri.

Relevant backend change:
https://github.vimeows.com/Vimeo/vimeo/pull/74829

## How to Test
Verify library builds correctly.

### DIFF
--- a/models/src/main/java/com/vimeo/networking2/PermissionPolicy.kt
+++ b/models/src/main/java/com/vimeo/networking2/PermissionPolicy.kt
@@ -2,15 +2,18 @@ package com.vimeo.networking2
 
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
+import com.vimeo.networking2.common.Entity
 import java.util.Date
 
 /**
  * A permission policy which represents a collection or permutation of certain [PermissionActions].
  *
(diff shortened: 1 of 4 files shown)


models/src/main/java/com/vimeo/networking2/PermissionPolicy.kt
```diff
@@ -30,4 +39,7 @@ data class PermissionPolicy(
 
     @Json(name = "permission_actions")
     val permissionActions: PermissionActions? = null
-)
+) :Entity {
+    override val identifier: String?
+        get() = uri
```
anthonycr: Can be simplified to 
```
override val identifier: String? = uri
```

anthonycr: Can be simplified to 
```
override val identifier: String? = uri
```

anthonycr (commented):
Left a minor comment and you have a detekt error that needs to be fixed

anthonycr (approved):
Feel free to merge after fixing the build and the identifier property

anthonycr (approved):

PR #570, opened 2022-07-13
MPS-3318 Add live event

howardrigberg:
# Summary
Adding support for live events to the networking library.

## Description
- Added `LiveEvent` data class and its related supporting data classes.
- Added `fetchLiveEvent` to `VimeoAPIClient` as a way to get a `LiveEvent` from the API.


elhicks (approved):
LGTM

PR #578, opened 2022-09-23
VIMAND-38 Add a live events endpoints

abalcer-ls:
# Summary
Added
- Create a live event endpoint
- Stop a live event endpoint
- Get list of the live events

Removed the isAutoCCEnabled field because it comes as an integer from the API, not a boolean.
`"auto_cc_enabled": 0,`


models/src/main/java/com/vimeo/networking2/LiveEvent.kt
```diff
@@ -87,9 +87,6 @@ data class LiveEvent(
     @Json(name = "head_clip")
     val firstVideoInPlaylist: Video? = null,
 
-    @Json(name = "auto_cc_enabled")
-    val isAutoCCEnabled: Boolean? = null,
```
RomanMinenok, line 91: Please add into PR description why we had to remove this just in case

RomanMinenok (approved):

PR #590, opened 2023-01-27
CXAN-71 Add editLiveEvent API and convinience methods

EStepiuk:
# Summary
https://vimean.atlassian.net/browse/CXAN-71

* Added edit live event APIs
* Added live event Schedule models
* Added convenience sealed interface VideoContainer
* Added related convenience extensions
* Fixed related model-generator issues


models-core/src/main/java/com/vimeo/networking2/adapters/Time.kt
```diff
@@ -0,0 +1,7 @@
+package com.vimeo.networking2.adapters
+
+import com.squareup.moshi.JsonQualifier
+
+@Retention(AnnotationRetention.RUNTIME)
+@JsonQualifier
+annotation class Time
```
RomanMinenok: models-annotations

EStepiuk: done

RomanMinenok (approved):

PR #615, opened 2023-06-09
Add editSession and trim fields into the Video object

abalcer-ls:
# Summary
Add editSession and trim fields to the Video model.
Fix parsing the webinar field.


maksym-ambroskin:
Is it possible to apply this serialisation strategy without `@SafeList` annotation?

